PR #192, opened 2022-05-16
Feature/wait for results

bartvanb:
See https://github.com/IKNL/vantage6-client/pull/41


vantage6-client/vantage6/client/__init__.py
```diff
@@ -628,6 +639,109 @@ def authenticate(self, username: str, password: str) -> None:
...
+            result = self.result.get(result_id)
+            start = dateutil.parser.isoparse(result['assigned_at'])
+        else:
+            start = None
+            result = None
+
+        try:
+            from rich.progress import Progress, ProgressColumn, SpinnerColumn, TimeElapsedColumn
+            from rich.text import Text
+            from rich.table import Column
```
frankcorneliusmartin, line 672: Can we move these imports to the top? Or is there a good reason not to do this?

vantage6-client/vantage6/client/__init__.py
```diff
@@ -628,6 +639,109 @@ def authenticate(self, username: str, password: str) -> None:
...
+            animation =  itertools.cycle(['|', '/', '-', '\\'])
+            t = time.time()
+
+            while not self.task.get(task_id)['complete']:
+                frame = next(animation)
+                sys.stdout.write(f'\r{frame} Waiting for task {task_id} ({int(time.time()-t)}s)')
+                sys.stdout.flush()
+                time.sleep(sleep)
+
+            sys.stdout.write('\rDone!                  ')
```
frankcorneliusmartin, line 734: The nbsp here are a bit of a hack. Would it be possible to do this neater and flush the line or something similar?

frankcorneliusmartin (changes requested):
Imports should be sorted'

PR #203, opened 2022-06-07
Added default roles 

### LINKED ISSUE #202: Set up default roles
The permission system in Vantage6 consists of 40-50 rules that can be grouped together in roles. These roles are usually then assigned to users. In the past, this system has been designed to allow complete flexibility for setting up roles exactly as one may wish. 
The experience from doing research projects with Vantage6 taught us that there are usually a few roles that recur in most projects. Therefore, we want to make these roles available by default, which makes it easier for new users to pick existing roles rather than create their own.

The idea is to make the following roles available by default:
- **Organization admin**: manages an organization and its users, roles, and nodes.
- **Viewer**: A user that can view their own organization, including the results of computations
- **Researcher**: Same as Viewer with the extension that a Researcher can also create new tasks (and delete them)
- **Collaboration admin**: Someone who manages a collaboration. This person can e.g. create new organizations and add them to their collaboration, but does not have full root permission

bartvanb:
Fix #202

### DIFF
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -378,14 +378,14 @@ class ServerApp:
             module = importlib.import_module('vantage6.server.resource.' + res)
             module.setup(self.api, self.ctx.config['api_path'], services)
 
-    def _add_default_roles(self):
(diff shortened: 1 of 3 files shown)


frankcorneliusmartin (approved):
I also get now why you ran into the circular import issue

PR #207, opened 2022-06-16
Added check if role exists instead of returning empty json in GET role/<id>

### LINKED ISSUE #206: GET `role/{id}` doesn't give error if id doesn't exist
The following is not consistent in the API:

GET organization/999 -> `"msg": "Organization id=999 not found!"` (status 404)
GET user/999 -> `"msg": "user id=999 is not found"`  (status 404)
(same for other resources)

however, for `role` we find:

GET role/999 -> `{}` (empty dictionary, status 200)

We need to adjust the role endpoint to also yield an error instead of an empty dictionary

bartvanb:
Fix #206. This was really an easy fix

### DIFF
--- a/vantage6-server/vantage6/server/resource/role.py
+++ b/vantage6-server/vantage6/server/resource/role.py
@@ -296,6 +296,10 @@ class Role(RoleBase):
     def get(self, id):
         role = db.Role.get(id)
 
+        if not role:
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
This is also an easy aprove. Lets do more of these..

PR #212, opened 2022-06-22
Using UUID4 instead of UUID1 to generate API keys for nodes

### LINKED ISSUE #211: API keys created within milliseconds after each other are rather similar
While working on the UI, I noticed that the API keys for nodes that were created at virtually the same moment are rather similar:

image

It would be better if they were completely unrelated. Currently we are using uuid1, I suggest using uuid4. See [here](https://stackoverflow.com/questions/1785503/when-should-i-use-uuid-uuid1-vs-uuid-uuid4-in-python) for a comparison.

bartvanb:
They are dissmilar now:
image

Fix #211

### DIFF
--- a/vantage6-server/vantage6/server/resource/node.py
+++ b/vantage6-server/vantage6/server/resource/node.py
@@ -247,7 +247,7 @@ class Nodes(NodeBase):
                         HTTPStatus.BAD_REQUEST
 
         # Ok we're good to go!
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):

frankcorneliusmartin:
I quote "We should not worry about collision", so lets not :P

Issue #59, opened 2021-12-01
Improve login requirements


bartvanb:
Attackers can currently do unlimited attempts to retrieve user passwords. There are a few good practices we may add to improve this.

I think the following would be nice:
- [x] Do not show message 'username does not exist' as it can inform attacker which usernames actually have accounts
- [x] Maximum number of login attempts: 5, then block 15min (except when reset via email)*
- [x] Minimum password length: 8
- [x] Maximum password length: 128 (https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/)
- [x] Include special character (0-9a-zA-Z special characters)
- [x] Make sure that when creating new server, root user is properly created

*these are default values that may be adapted in the server config file

bartvanb:
Note: not all of this is backwards compatible so maybe in 4.0?

bartvanb:
Login attempts: 3, block time 15min

bartvanb:
The password requirements are not checked yet when creating a new user

PR #220, opened 2022-06-29
change password endpoint

### LINKED ISSUE #210: Implement endpoint to allow users to change their password using the current password
We currently offer two ways in which users can reset their password:
- `client.util.reset_my_password()` which sends an email to the user with a renewal token followed by `client.util.set_my_password()`  to use said token to set a new password
- simply do a PATCH `/user/<id>` with a new password -> this should be avoided because it is not obvious it can be done and because it in theory allows user A to reset the password of user B if they have sufficient permissions.

So two action points:
- [ ] Adapt PATCH `/user/<id>` so that password changes are no longer allowed there
- [ ] Create a new endpoint to change passwords using the old password

bartvanb:
Fix #210 

- New endpoint to change password using current password
- Changing password no longer possible via PATCH user
- Also made both changes for the Python client


vantage6-server/vantage6/server/resource/recover.py
```diff
@@ -141,6 +149,73 @@ def post(self):
...
+                    description: current password
+                  new_password:
+                    type: string
+                    description: new password
+
+        responses:
+          200:
+            description: Ok
+          400:
+            description: Current or new password is missing from JSON body
```
frankcorneliusmartin, line 180: Or password match

frankcorneliusmartin (changes requested):
Same changes that are mentioned in https://github.com/vantage6/vantage6/issues/59

vantage6-server/vantage6/server/resource/recover.py
bartvanb: Done

frankcorneliusmartin (approved):

PR #221, opened 2022-06-29
Exit immediately if node tries to authenticate with wrong API key

### LINKED ISSUE #169: Nodes shouldn't re-authenticate if they use a wrong API key
See: https://github.com/IKNL/vantage6-node/pull/63#discussion_r792398258

bartvanb:
Instead of trying 10 times

Fix #169

### DIFF
--- a/vantage6/vantage6/cli/globals.py
+++ b/vantage6/vantage6/cli/globals.py
@@ -24,6 +24,8 @@ DEFAULT_NODE_ENVIRONMENT = "application"
 #
 PACAKAGE_FOLDER = Path(__file__).parent.parent.parent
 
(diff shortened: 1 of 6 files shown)


vantage6/vantage6/cli/globals.py
```diff
@@ -24,6 +24,8 @@
 #
 PACAKAGE_FOLDER = Path(__file__).parent.parent.parent
 
+# FIXME BvB 22-06-28 think this is also defined in the node globals, and this
+# one appears not to be used
```
frankcorneliusmartin, line 28: Before removing it here: check that this is not used by the algorithm wrapper thingies. (it shouldn't but that the only place that would make kinda sense if it where.

frankcorneliusmartin (approved):

PR #226, opened 2022-07-05
Set all nodes to offline before initiating the socket connection 

### LINKED ISSUE #40: Nodes appear to be offline
If the server reboots, the nodes status field is not set properly when using multiple workers. It seems likely that the `start()` method is triggered after some of the clients (=nodes) reconnected. 

Reported by @anjavangestel

bartvanb:
instead of afterward, to prevent mixups in the node status stored in the database

Fix #40 

**I have not tested this code yet with multiple workers** If we diagnosed the issue right, it would make sense that this fixes it but we should test after deploying it as well!

### DIFF
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -134,7 +134,7 @@ class ServerApp:
 
     @staticmethod
     def _set_nodes_offline():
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin:
Closed because this does not resolve the issue. See issue for more details

PR #229, opened 2022-07-07
Feature/rabbitmq from vserver start

### LINKED ISSUE #198: Support RabbitMQ from `vserver start`
Currently, RabbitMQ functionality cannot be combined properly with `vserver start`. That is, a new dockerized RabbitMQ service is started for every `vserver start` that is executed, thus not allowing multiple servers to use the same rabbitMQ queue. This should be changed

- [x] If RabbitMQ uri is not a local address, don't create any docker containers but link the server to the RabbitMQ service
- [x] If it is a local address, check if the RabbitMQ already exists. Only create if it doesn't
- [x] On server shutdown, think about whether to keep of kill RabbitMQ. If killing, check if other servers are running.

bartvanb:
Fix #198

### DIFF
--- a/vantage6-common/vantage6/common/docker/addons.py
+++ b/vantage6-common/vantage6/common/docker/addons.py
@@ -10,6 +10,7 @@ import pathlib
 from dateutil.parser import parse
 from docker.client import DockerClient
 from docker.models.containers import Container
+from typing import Dict
 
 from vantage6.common import logger_name
 from vantage6.common import ClickLogger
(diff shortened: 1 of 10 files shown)


vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -310,10 +311,59 @@ def remove_container(container: Container, kill=False) -> None:
...
+    container: Container
+        The container in which we are interested
+
+    Returns
+    -------
+    int
+        Number of networks in which the container resides in which there are
+        also other containers
+    """
+    client = docker.from_env()
```
frankcorneliusmartin, line 351: At line 21 there is already a docker client defined at module level. So it would be better to either remove that or this?

vantage6-common/vantage6/common/docker/network_manager.py
```diff
@@ -53,20 +55,31 @@ def create_network(
...
-            self.network_name,
-            driver="bridge",
-            internal=is_internal,
-            scope="local",
+        existing_networks = self.docker.networks.list(
+            names=[self.network_name]
         )
+        if existing_networks:
+            if len(existing_networks) > 1:
+                self.log.error(
```
frankcorneliusmartin, line 63: Why do we no longer delete them automatically? Is that because it sometime fails when there are services attached to these networks?

frankcorneliusmartin (approved):
Just a note that the command `vserver start` is going to be a develop/test command and that we prefer starting them using a Docker compose file. So starting a rabbitMQ container is usefull for non production environments only. (That makes sense because `vserver start` is not able to spawn multiple workers)

vantage6-common/vantage6/common/docker/addons.py
bartvanb: Removing local client, the global one is used in several other functions

vantage6-common/vantage6/common/docker/network_manager.py
bartvanb: In principle it is now possible to start two servers with `vserver start` and connect them to the same RabbitMQ container. If you shutdown a server, but there are still other servers running that connect to the RabbitMQ container, the RabbitMQ container is not shutdown, and also the network is not deleted. 
An alternative I thought about was to remove the RabbitMQ service from the server network when the server is shutdown, and then remove the network. But if I remember correctly, it was quite complex to add the RabbitMQ container back into the network when the server would be restarted in such cases.

PR #231, opened 2022-07-11
Improved password and login requirements

### LINKED ISSUE #59: Improve login requirements

bartvanb:
Better requirements for password, maximum number of login attempts and improved user feedback

Fix #59 

NOTE: this makes a breaking change because database fields are added! We should therefore consider when to release this.

NOTE 2: this branch was initiated from another non-main branch that has a PR here: #220. Be sure to merge that PR before this one!


vantage6-client/vantage6/client/__init__.py
```diff
@@ -206,6 +206,7 @@ def request(self, endpoint: str, json: dict = None, method: str = 'get',
                 self.log.debug(response.content)
 
             if first_try:
+                # FIXME BvB 2022-08-01 only try refresh if 401 unauthorized?!
```
frankcorneliusmartin, line 209: I agree that the token only needs to be refreshed when a 401 is returned. But for robustness it would be good to also retry when other error codes occur. If I am not mistaken, this happens also at other places in the code.

vantage6-client/vantage6/client/__init__.py
```diff
@@ -651,6 +652,34 @@ def get_server_health(self) -> dict:
             """
             return self.parent.request('health')
 
+        def change_my_password(self, current_password: str,
+                               new_password: str) -> dict:
+            """Start reset password procedure
+
+            Either a username of email needs to be provided.
```
frankcorneliusmartin, line 659: This does not make sense to me? It seems not the case in this method? And you are already logged in when using this method?

vantage6-server/tests_server/test_resources.py
```diff
@@ -114,18 +114,16 @@ def login(self, type_='root'):
 
         return headers
 
-    def create_user(self, organization=None, rules=[]):
+    def create_user(self, organization=None, rules=[], password="password"):
```
frankcorneliusmartin, line 117: Why supply a default password? If you want a default option, then I suggest to accept a `None` value and generate a solid password for them?

frankcorneliusmartin: Oh sorry nevermind me.. This is a testfile 👍

vantage6-server/vantage6/server/globals.py
```diff
@@ -37,3 +37,8 @@
 # Whenever the refresh tokens should expire. Note that setting this to true
 # would mean that nodes will disconnect after some time
 REFRESH_TOKENS_EXPIRE = False
+
+# Globals for how many authentication attempts a user can have and what the
+# cooldown period is when user is temporarily blocked
+MAX_FAILED_LOGIN_ATTEMPTS = 5
+INACTIVATION_PERIOD_MINUTES = 15
```
frankcorneliusmartin, line 44: I think it would even be nicer if these are overwritable in the config file.

vantage6-server/vantage6/server/model/user.py
```diff
@@ -49,7 +57,40 @@ def __repr__(self):
     def _validate_password(self, key, password):
         return self.hash(password)
 
-    def set_password(self, pw):
+    def set_password(self, pw) -> Union[str, None]:
+        """
+        Set the password of the current user. This function doesn't save the
+        new password to the database
```
frankcorneliusmartin, line 63: Any particular reason we do not do that?

frankcorneliusmartin: I get that if you do `user.password='something'` the user object is not saved. But I would expect `.set_password(...)` to also save it to the database?

vantage6-server/vantage6/server/resource/recover.py
```diff
@@ -43,6 +43,14 @@ def setup(api, api_base, services):
...
     )
 
+    api.add_resource(
+        ChangePassword,
+        path+'/change',
+        endpoint='change_password',
+        methods=('POST',),
+        resource_class_kwargs=services
+    )
+
```
frankcorneliusmartin, line 53: I am a little unconfortable with this being in the recover section. Would the `user` not be a better spot? Although I see that might be confusing with the PATCH on `/user/<id>`... A suggestion:

* PATCH `/user/<id>/password`
* PATCH  `/account/password`

Let me know what you think

vantage6-server/vantage6/server/resource/recover.py
```diff
@@ -141,6 +157,75 @@ def post(self):
...
+                    description: current password
+                  new_password:
+                    type: string
+                    description: new password
+
+        responses:
+          200:
+            description: Ok
+          400:
+            description: Current or new password is missing from JSON body
```
frankcorneliusmartin, line 187: Or curent password matches new password

frankcorneliusmartin (changes requested):

vantage6-client/vantage6/client/__init__.py
bartvanb: Fair enough, good reason

vantage6-client/vantage6/client/__init__.py
bartvanb: Ctrl-C + Ctrl-V, and then forgot to change this...

vantage6-server/vantage6/server/globals.py
bartvanb: Yes that would be nicer indeed. Let me check if I can do that quickly, otherwise let's discuss if we do this now or later

vantage6-server/vantage6/server/resource/recover.py
bartvanb: Also to me it didn't seem right for this to have the path `/recover/change`. On the other hand I do like this being in recover.py where we do all the other password/API key resets.
Maybe we can keep it in the same file but simply modify the endpoint name to make it PATCH `/account/password`

vantage6-server/vantage6/server/resource/recover.py
bartvanb: And also: if new pw doesnt meet the requirements

vantage6-server/vantage6/server/model/user.py
bartvanb: Yes, I will adapt this. I guess I didn't do it because we do all of these `save()` actions in the resource folder. But this will then be the first `self.save()` we implement :-)

bartvanb:
I just incorporated most review comments, but there are two things left to do/discuss:
- where to put the change password endpoint code and what to name the endpoint
- make the number of password attempts and block time settable from the config file & wizard

vantage6-server/vantage6/server/resource/recover.py
bartvanb: This is now also already updated on the branch where the change password endpoint is implemented

vantage6-server/vantage6/server/resource/recover.py
bartvanb: Let's change this into `/password/change`

bartvanb:
You can now add the following in the config file to define how often a user can mistype their password and how long their account will be blocked when they do.
```    
password_policy:
    max_failed_attempts: 5
    inactivation_minutes: 15
```

vantage6-server/vantage6/server/globals.py
bartvanb: Added

frankcorneliusmartin (approved):

PR #237, opened 2022-07-28
Require that API keys are defined in vserver import files. 

### LINKED ISSUE #55: vserver import(ed) nodes unknow api-keys
We cannot see the API_KEY assigned to the nodes which are imported. 

Suggestions: 
* Maybe print them to the terminal when the import is finished (not to the logs)
* Make them a key in the import yaml file

bartvanb:
This prevents they are unknown after the import. Also added `--mount-src` option to `vserver import` while I was at it.

Fix #55 

Note: when releasing this, we should also update the documentation [here](https://docs.vantage6.ai/usage/running-the-server/importing-entities) with the new content of example_fixture.yaml (see code changes)

### DIFF
--- a/vantage6-server/vantage6/server/controller/fixture.py
+++ b/vantage6-server/vantage6/server/controller/fixture.py
@@ -7,6 +7,26 @@ module_name = __name__.split('.')[-1]
 log = logging.getLogger(module_name)
 
 
+# TODO maybe move this function to a more general location (utils?)
+def _is_valid_uuid(uuid_to_test):
+    """
+    Check if uuid_to_test is a valid UUID.
+
+     Parameters
+    ----------
+    uuid_to_test : str
+
+     Returns
+    -------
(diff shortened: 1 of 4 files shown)


vantage6-server/vantage6/server/controller/fixture.py
```diff
@@ -44,13 +64,22 @@ def load(fixtures, drop_all=False):
...
+                          "fields 'name' and 'api_key'. This is not the case "
+                          f"for participant {participant} in collaboration "
+                          f"{collaboration.name}")
+                exit(1)
+
+            org_name = participant.get("name")
+            node_api_key = participant.get("api_key")
+
+            # check if api key is valid uuid
+            if not _is_valid_uuid(node_api_key):
```
frankcorneliusmartin, line 80: I am not completely sure if we need to force this to be a UUID.

frankcorneliusmartin (approved):

PR #290, opened 2022-08-22
Fix 285: Backwards compatibility issues due to pyJWT upgrade

frankcorneliusmartin:


bartvanb (changes requested):
Fully agree with the changes, but I think there should be an additional change in client/tools/dispatch_rpc.py as we just discussed offline

bartvanb (approved):

PR #291, opened 2022-08-22
Froze discord dependency

frankcorneliusmartin:


.github/workflows/release.yml
```diff
@@ -138,7 +138,7 @@ jobs:
 
       - name: Install dependancies
```
bartvanb, line 139: There are no a's in dependencies :-)

bartvanb (approved):

.github/workflows/release.yml
bartvanb: Already resolved in other PR

bartvanb (approved):

PR #308, opened 2022-09-14
Thread safe flask context

frankcorneliusmartin:


vantage6-node/vantage6/node/proxy_server.py
```diff
@@ -249,21 +212,64 @@ def proxy_task():
...
+        log.exception('Could not extract headers from request...')
+
+    log.debug(f"{len(organizations)} organizations")
 
     # For every organization we need to encrypt the input field. This is done
     # in parallel as the client (algorithm) is waiting for a timely response.
     # For every organizationn the public key is retrieved an the input is
     # encrypted specifically for them.
+    @copy_current_request_context
+    def encrypt_input(organization: dict) -> dict:
```
dsmits, line 227: To me it looks like `encrypt_input` is now inside the `decrype_result` function. Or is that github formatting?

dsmits (changes requested):
Just this one comment, I will check it locally on my ide after my meeting

dsmits (approved):

vantage6-node/vantage6/node/proxy_server.py
frankcorneliusmartin: That is indeed the case. We only want to copy the context if we are actually in the call, and not when the module is loaded

PR #311, opened 2022-09-20
Create SECURITY.md

frankcorneliusmartin:
Hi @bartvanb can you check if you agree with this?

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,14 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 3.x.x   | :white_check_mark: |
+| 2.3.x   | :white_check_mark: |
+| < 2.3   | :x:                |
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,14 @@
...
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 3.x.x   | :white_check_mark: |
+| 2.3.x   | :white_check_mark: |
+| < 2.3   | :x:                |
+
+## Reporting a Vulnerability
+
+Please report (suspected) security vulnerabilities to vantage6@iknl.nl. You will receive a response from us within 72 hours.
```
bartvanb, line 13: 'We always strive to respond within 72 hours' -> if someone sends it at 4am on Friday morning, I'm not gonna see before monday if I'm not working on Friday. Plus we might miss the email sometimes.

bartvanb (commented):
One small comment. Also a question: how will people see this file in Github (other than clicking on it in the repo?)

SECURITY.md
bartvanb: -> as discussed offline, having a term is recommended, so I'm ok with it

bartvanb (approved):

PR #320, opened 2022-09-22
Improve checks for non-existing resources in unittests

### LINKED ISSUE #265: [Bug] Unittest with <id:int> values of -1 should be replaced
**Describe the bug**
Flask does not support negative int values for the <:int> pattern always resulting in a 404. We always tested for a 404 so the unittests passed but not because the resource could not be found but the route itself. 

**To Reproduce**
Run the unittests with logging enabled

**Expected behavior**
We should test the resource not founf 404 rather than the route not found.

bartvanb:
… as Flask doesn't support negative ints
Fix #265


frankcorneliusmartin (approved):
Simple enough.

PR #323, opened 2022-09-26
Added robots.txt file to block web crawlers

### LINKED ISSUE #322: [Change Request] Add `robots.txt` to fend off web crawlers
**Description**

Search engine web crawlers try to find which URLs can be queried on a website so they can index them in search results. With a `robots.txt`, we can list the URLs we have and the web crawlers can take that information. This prevents a lot of 404 errors in server logs when crawlers query non-existing routes.

See [here for a guide](https://developers.google.com/search/docs/crawling-indexing/robots/intro)

bartvanb:
Fix #322

### DIFF
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -14,7 +14,9 @@ import traceback
 from http import HTTPStatus
 from werkzeug.exceptions import HTTPException
 from flasgger import Swagger
-from flask import Flask, make_response, current_app
+from flask import (
+    Flask, make_response, current_app, request, send_from_directory
+)
 from flask_cors import CORS
 from flask_jwt_extended import JWTManager
 from flask_marshmallow import Marshmallow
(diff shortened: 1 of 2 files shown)


vantage6-server/vantage6/server/__init__.py
```diff
@@ -243,6 +246,11 @@ def error2_remove_db_session(error):
             return {'msg': f'An unexpected error occurred on the server!'}, \
                 HTTPStatus.INTERNAL_SERVER_ERROR
 
+        @self.app.route('/robots.txt')
+        def static_from_root():
+            return send_from_directory(self.app.static_folder,
+                                       request.path[1:])
```
frankcorneliusmartin, line 252: Why isnt this just `robots.txt`?

bartvanb: Would be possible, just kept it a bit general so that if we want to serve other static files in the future (e.g. a file on which routes we have), we need only add them in line 249 and keep line 252 the same.

frankcorneliusmartin (approved):

PR #326, opened 2022-10-04
Track which users initiate which tasks

### LINKED ISSUE #43: Get all tasks for a given user
The `task` endpoint is currently able to return all tasks that a given organization initiated. The feature request here is to extend this to getting the tasks for a given user.

This feature requires a new field to be added in the database.

- [x] Add column to database and fill it when task is created
- [x] Add filter option to the `task` endpoint
- [x] Implement in client
- [x] Implement in UI

bartvanb:
Fix #43 (except for the UI part which is a separate PR in the UI repo)


vantage6-client/vantage6/client/__init__.py
```diff
@@ -1509,8 +1509,9 @@ def get(self, id_: int, include_results: bool = False) -> dict:
             return self.parent.request(f'task/{id_}', params=params)
 
         @post_filtering()
-        def list(self, initiator: int = None, collaboration: int = None,
-                 image: str = None, parent: int = None, run: int = None,
+        def list(self, init_org_id: int = None, init_user_id: int = None,
```
frankcorneliusmartin, line 1512: I do find this a bit incosistent. We never specified that it needs to be an `_id`. So either we need to change that all enpoints or I would not start using it here.

frankcorneliusmartin: Same goes for the user

vantage6-node/vantage6/node/server_io.py
```diff
@@ -134,14 +134,14 @@ def set_task_start_time(self, id: int):
             "started_at": datetime.datetime.now().isoformat()
         })
 
-    def patch_results(self, id: int, initiator_id: int, result: dict):
+    def patch_results(self, id: int, init_org_id: int, result: dict):
```
frankcorneliusmartin, line 137: Internally this is not an issue

frankcorneliusmartin (commented):
Could you also add the init user and org in the __repr__ function of the `model/task`?
Other than that it looks good to me!

frankcorneliusmartin (changes requested):
Remove logging from the `only_for` decorator, as mentioned

vantage6-client/vantage6/client/__init__.py
bartvanb: Made this consistent

vantage6-node/vantage6/node/server_io.py
bartvanb: Left this the same - as it is not an issue

frankcorneliusmartin (approved):
lgtm

PR #329, opened 2022-10-07
Logging endpoint on each request

### LINKED ISSUE #295: [Change Request] Add logging for each endpoint request at the server
**Description**
Not every endpoint writes log messages if it is called in an appropriate manner (it only writes warnings/errors if it is called wrongly). Sometimes, it is informative to check the server logs to see which endpoints have been called. Therefore, the idea of this change request is to add a log message on a generic place for all endpoints saying that that particular endpoint has been called.
Note that there should *not* be a message for non-existing endpoints.

bartvanb:
and which user/node/container is doing it (only for authenticated requests)

Fix #295

### DIFF
--- a/vantage6-server/vantage6/server/resource/__init__.py
+++ b/vantage6-server/vantage6/server/resource/__init__.py
@@ -104,18 +104,23 @@ def only_for(types=['user', 'node', 'container']):
                 g.user = user
                 assert g.user.type == g.type
                 log.debug(
-                    f"Received request from user {user.username} ({user.id})")
+                    f"Received request {request.method} {request.url_rule} "
+                    f"from user {user.username} ({user.id})")
 
             elif g.type == 'node':
                 node = get_and_update_authenticatable_info(identity)
                 g.node = node
                 assert g.node.type == g.type
                 log.debug(
-                    f"Received request from node {node.name} ({node.id})")
+                    f"Received request {request.method} {request.url_rule} "
+                    f"from node {node.name} ({node.id})")
 
             elif g.type == 'container':
                 g.container = identity
-                log.debug("Received request from container")
+                log.debug(
+                    f"Received request {request.method} {request.url_rule} "
+                    f"from container from node id {identity['node_id']} and "
+                    f"task id {identity['task_id']}")
 
             else:
                 raise Exception(f"Unknown entity: {g.type}")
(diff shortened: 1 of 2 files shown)


vantage6-server/vantage6/server/resource/__init__.py
```diff
@@ -103,14 +103,19 @@ def decorator(*args, **kwargs):
...
             elif g.type == 'node':
                 node = get_and_update_authenticatable_info(identity)
                 g.node = node
                 assert g.node.type == g.type
+                log.debug(
+                    f"Received request from node {node.name} ({node.id})")
 
             elif g.type == 'container':
                 g.container = identity
+                log.debug("Received request from container")
```
frankcorneliusmartin, line 118: Can we also log the `node_id` and `task_id` from the container?

frankcorneliusmartin (changes requested):
other than that it looks good

vantage6-server/vantage6/server/resource/__init__.py
bartvanb: Added this.
Also, I added the request method and route to the logging in `resource/__init__.py` as that looked clearer in the logs. However, as we also log all request in the `@before_request` decorator, we now add two log messages for each request where JWT token is checked. Let's discuss what we want with that.

frankcorneliusmartin (approved):
Is the assumption safe that all the items always exists?

vantage6-server/vantage6/server/resource/__init__.py
frankcorneliusmartin: Why not remove the `f"Received request {request.method} {request.url_rule} "` from the `only_for` decorator?

bartvanb: We can remove the logging from `only_for` (and keep it only in `before_request`), but then we only know which requests are done, and not by whom

bartvanb:
> Is the assumption safe that all the items always exists?

I *think* so. `method` and `url_rule` seem to be reliable properties of a Flask request object (briefly checked the code). The other fields are in our own Sqlalchemy models so should also exist

vantage6-server/vantage6/server/resource/__init__.py
frankcorneliusmartin: I ment only remove the ` {request.method} {request.url_rule}` part from the logging. As it will be a duplicate with the line from `before_request`.

bartvanb: Ah clear. Just adapted it

PR #330, opened 2022-10-10
Feature/remove hardcoded email addresses

### LINKED ISSUE #319: [Feature Request] Hard coded email adresses
**Problem description**
The server has several hard coded emails. 
* https://github.com/vantage6/vantage6/blob/main/vantage6-server/vantage6/server/templates/mail/reset_password_token.html
* https://github.com/vantage6/vantage6/blob/9d4d043c6069538bbd74543666b4dd873dc283fc/vantage6-server/vantage6/server/resource/recover.py#L190
* `/token/user` 

**Desired solution**
Make this a setting in the configuration file.

bartvanb:
Fix #319 

To repeat question in issue: @frankcorneliusmartin I don't see why you added /token/user in the description of this issue, can you clarify?

### DIFF
--- a/vantage6-server/vantage6/server/resource/recover.py
+++ b/vantage6-server/vantage6/server/resource/recover.py
@@ -189,6 +189,7 @@ class RecoverPassword(ServicesResources):
         email_info = self.config.get("smtp", {})
         email_sender = email_info.get("username",
                                       DEFAULT_SUPPORT_EMAIL_ADDRESS)
+        support_email = self.config.get("support_email", email_sender)
 
         self.mail.send_email(
             "password reset",
@@ -198,7 +199,7 @@ class RecoverPassword(ServicesResources):
                                       token=reset_token),
             html_body=render_template("mail/reset_password_token.html",
                                       token=reset_token,
-                                      support_email=email_sender)
+                                      support_email=support_email)
         )
 
         return ret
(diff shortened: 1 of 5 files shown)


frankcorneliusmartin:
> Fix #319
> 
> To repeat question in issue: @frankcorneliusmartin I don't see why you added /token/user in the description of this issue, can you clarify?

See issue. tl;dr: I dont know

vantage6-server/vantage6/server/resource/recover.py
```diff
@@ -185,14 +186,19 @@ def post(self):
             {"id": str(user.id)}, expires_delta=expires
         )
 
+        email_info = self.config.get("smtp", {})
+        email_sender = email_info.get("username",
+                                      DEFAULT_SUPPORT_EMAIL_ADDRESS)
```
frankcorneliusmartin, line 191: I wonder if the mail sender is also the support mail address?

frankcorneliusmartin (commented):
Looks good however I am not completely sure if the smtp email is also the email you want to use for support in all cases?

vantage6-server/vantage6/server/resource/recover.py
bartvanb: So you mean that in line 201, we should replace `email_sender` by some settable support email right? We could make an option in the config to set a support email address.

frankcorneliusmartin (commented):
See comments

frankcorneliusmartin (approved):
lgtm

PR #365, opened 2022-10-20
On server startup, add columns in the database if they were missing

### LINKED ISSUE #364: [Feature Request] Add columns to the SQL database on startup
**Problem description**
Whenever we want to add a column to the database, we lose backwards compatibility because people that have already deployed a server cannot update that server to the new version without modifying the database.
Therefore, any addition of database columns should currently always result in a major version update, which is undesirable for small features.

**Desired solution**
On server startup, check if columns are missing from the database and add them if necessary.

bartvanb:
Fix #364

### DIFF
--- a/vantage6-server/vantage6/server/model/base.py
+++ b/vantage6-server/vantage6/server/model/base.py
@@ -1,14 +1,16 @@
 import logging
 import os
 import inspect as class_inspect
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Looks really nice bart

PR #366, opened 2022-10-21
Add algorithm input to node logs if it is serialized as pickle

### LINKED ISSUE #171: Display task information in the node log
It would be nice to have a little bit of “user-friendly” information in the log of the node? In particular, the id of the task being run could be supplemented by a small description of what it does (e.g., “retrieve column headers”, “Poisson model for <dataset)”) just so that the users can have at least a remote idea of the status of the analysis.

bartvanb:
Fix #171


frankcorneliusmartin (approved):
lgtm, but to removed again soon -i think-

PR #377, opened 2022-11-02
Bugfix/pipeline versioning is broken

frankcorneliusmartin:


.github/workflows/release.yml
```diff
@@ -20,30 +20,67 @@ jobs:
...
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+          echo "major=${MAJOR}" >> $GITHUB_OUTPUT
+          echo "minor=${PARTS[1]}" >> $GITHUB_OUTPUT
           PARTS2=( $(grep -Eo '[[:digit:]]+|[^[:digit:]]+' <<< ${PARTS[2]} ) )
-          echo "::set-output name=patch::${PARTS2[0]}"
-          echo "::set-output name=stage::${PARTS2[1]}"
-          echo "::set-output name=build::${PARTS2[2]}"
+          echo "patch=${PARTS2[0]}" >> $GITHUB_OUTPUT
+          echo "stage=${PARTS2[1]}" >> $GITHUB_OUTPUT
+          echo "build=${PARTS2[2]}" >> $GITHUB_OUTPUT
```
bartvanb, line 43: Out of curiosity: why these changes?

.github/workflows/release.yml
```diff
@@ -20,30 +20,67 @@ jobs:
...
+          if [[ $stage == 'rc' ]]; then SPEC="--spec candidate"; fi
+          FLAGS="--version ${major}.${minor}.${patch} $BUILD --post 0 $SPEC"
+          make set-version FLAGS="$FLAGS"
+
+      - name: commit
+        env:
+          version: ${{ steps.split.outputs.version }}
+        run: |
+          git config --local user.name github-actions
+          git config --local user.email github-actions@github.com
```
bartvanb, line 79: Why have you chosen to implement it this way? I think there are also alternatives - is this to keep it simple?
And do we now need to give this `github-actions` user permissions to push on certain branches?

.github/workflows/release.yml
```diff
@@ -20,30 +20,67 @@ jobs:
...
           if [[ $MAJOR == '1' ]]; then MAJOR_NAME="trolltunga"; fi
-          echo "::set-output name=major_name::${MAJOR_NAME}"
+          echo "major_name=${MAJOR_NAME}" >> $GITHUB_OUTPUT
+
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: main
+          token: ${{ secrets.COMMIT_PAT }}
```
bartvanb, line 55: Just curious what we are doing here?

bartvanb (approved):

.github/workflows/release.yml
bartvanb: -> other version was deprecated

.github/workflows/release.yml
bartvanb: -> Frank will document this in the devops docs. It is a personal token required to be able to do the git commit

.github/workflows/release.yml
bartvanb: -> no need to give this user permissions, it uses the secret token from comment below

PR #375, opened 2022-11-01
bugfix: non existent images wont crash the speaker worker

frankcorneliusmartin:

PR #382, opened 2022-11-02
Bugfix/non json response in local proxy

frankcorneliusmartin:
Reponse failed when the server responded with a list. This happend due to the implementation of the pagination. The request is now completely forwarded as it is. 

Used: https://stackoverflow.com/questions/19568950/return-a-requests-response-object-from-flask

Want to merge this to main, as I want to release a patch asap. I banched of another fix #375  , so lets release that fix too. (Merge that one first)


bartvanb (approved):
It's only the last line that's new in this PR right? The rest is all in #375

PR #389, opened 2022-11-15
Feature/latest algorithm base tag

frankcorneliusmartin:


bartvanb (commented):
I'm a bit confused here. This change doesn't seem to address the issue it's linked to: it only creates another image with the `latest` tag. Has the issue been addressed elsewhere? If so, it would be nice to refer to the relevant PR in the issue

frankcorneliusmartin:
> I'm a bit confused here. This change doesn't seem to address the issue it's linked to: it only creates another image with the `latest` tag. Has the issue been addressed elsewhere? If so, it would be nice to refer to the relevant PR in the issue

Yes indeed, this has been resolved with the pipeline overhaul. There the base is already tagged with the vantage6-client version that is inside the container. See 
https://github.com/vantage6/vantage6/blob/1046f4f6983141b1f6eda3a297c4637279591515/Makefile#L64-L70

and

https://github.com/vantage6/vantage6/blob/1046f4f6983141b1f6eda3a297c4637279591515/.github/workflows/release.yml#L278-L279

where the `$version` is obtained from the _tag_ on the commit.

PR #404, opened 2022-11-23
Remove iknl-flasgger from requirements as it is not used but leads to…

### LINKED ISSUE #394: [Bug] API docs tables are not properly formatted
**Describe the bug**
The tables in our API docs are no longer formatted properly. See e.g. https://petronas.vantage6.ai/apidocs/#/Collaboration/get_collaboration. We should figure out why that is.

Version: 3.4.2

bartvanb:
… inproper formatting of the tables in the API docs

Fix #394

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,7 +32,6 @@ gevent==21.8.0
 gevent-websocket==0.10.1
 greenlet==1.1.1
 idna==2.10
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Ah good find

PR #406, opened 2022-11-28
Change/update infra base image

frankcorneliusmartin:
* Base image is major versioned
* I made a rebuild for version 3 (located in harbor2)
* Base image is only build with x.0.0 versions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -204,7 +204,9 @@ jobs:
     needs: [version, build-n-release] # we need to packages to be released
     env:
       version: ${{ needs.version.outputs.version }}
+      major: ${{ needs.version.outputs.major }}
       minor: ${{ needs.version.outputs.minor }}
+      patch: ${{ needs.version.outputs.patch }}
       stage: ${{ needs.version.outputs.stage }}
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -237,12 +239,12 @@ jobs:
 
       - # we only release a new base image if it is a minor and major release
         name: Build Docker base image
-        if: ${{ env.minor == 0 && env.stage == '' }}
-        run: make base-image TAG=$version
+        if: ${{ env.minor == 0 && env.patch == 0 && env.stage == '' }}
```
bartvanb, line 242: The comment in line 240 doesn't match this line: it looks like we only release base image for major release right?

Makefile
```diff
@@ -55,8 +60,10 @@ install-dev:
 
 base-image:
 	@echo "Building ${REGISTRY}/infrastructure/infrastructure-base:${TAG}"
+	@echo "Building ${REGISTRY}/infrastructure/infrastructure-base:lastest"
```
bartvanb, line 63: 'la*s*test' -> 'latest'

docker/infrastructure-base.Dockerfile
```diff
@@ -8,22 +8,32 @@ LABEL maintainer="Frank Martin <f.martin@iknl.nl>"
...
 
+# install requirements. We cannot rely on setup.py because of the way
+# python resolves package versions. To control all dependencies we install
+# them from the requirements.txt
+COPY requirements.txt /vantage6/requirements.txt
+RUN pip install -r /vantage6/requirements.txt \
+    --extra-index-url https://www.piwheels.org/simple
+
+# Removed individual requirements files. Not sure if we want to keep
+# this at project level or in the packages themselfs.
```
bartvanb, line 19: What do you mean here? I assume we'll keep the individual requirements files because they are used if someone does e.g. `pip install vantage6-client`. I think they're now only not used here when Docker images are built, right?

bartvanb (commented):
Just a few comments and a question

docker/infrastructure-base.Dockerfile
frankcorneliusmartin: I refered here to the `requirements.txt` which are not used by pip install

bartvanb (approved):

PR #424, opened 2022-12-07
Implementation of multi db wrapper

### LINKED ISSUE #398: [Feature Request] Multi database wrapper
**Problem description**
All current database wrappers read one database file and pass that on to the algorithm. Some algorithms may want to use several databases and doing so would require a new wrapper that looks something like this:

```
class MultiDBWrapper(WrapperBase):
    @staticmethod
    def load_data(database_uri, input_data):
        data = {}
        for uri in database_uri:
            data[uri] = pandas.read_csv(database_uri)
        return data
```

Requested by @anjavangestel

bartvanb:
Fix #398.

This wrapper passes the following as 'data' to an algorithm:

```{'my_label': '/mnt/data/my_label.csv', 'default': '/mnt/data/default.csv}```

The algorithm is then responsible itself to read these datafiles (e.g. `pd.read_csv(data['my_label'])`. I made that choice to allow passing databases of different types to the algorithm.

### DIFF
--- a/vantage6-client/vantage6/tools/docker_wrapper.py
+++ b/vantage6-client/vantage6/tools/docker_wrapper.py
@@ -44,42 +44,43 @@ class WrapperBase(ABC):
 
     def wrap_algorithm(self, module):
         """
-            Wrap an algorithm module to provide input and output handling for the
-            vantage6 infrastructure.
+        Wrap an algorithm module to provide input and output handling for the
+        vantage6 infrastructure.
 
-            Data is received in the form of files, whose location should be specified
-            in the following environment variables:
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: vantage6-client/vantage6/tools/docker_wrapper.py [hunk 1]
```diff
@@ -96,9 +103,10 @@ def wrap_algorithm(self, module):
         with open(token_file) as fp:
             token = fp.read().strip()
 
+        # TODO in v4+, we should work with multiple databases instead of this
+        # default one
```
frankcorneliusmartin, line 107: I would keep it for legacy purposes?

frankcorneliusmartin (approved):

Issue #419, opened 2022-12-05
[Feature Request] Redesign encryption module


bartvanb:
**Problem description**
There are a few things with the current setup of encryption that are not logical. E.g. there is a command `vnode create-keypair` that creates public and private key and updates the public one to the database. The users of that organization then have to distribute the private key amongst themselves, which is strange. Probably there should be keypair per user, and potentially also a separate one for nodes.

This rework will also help to lose dependency of the CLI on the client (see #292)

**Desired solution**
Encryption module is improved

frankcorneliusmartin:
Another weird thing is that if you use `v6 node create-private-key` it does not check whether the node is connected to an encrypted collaboration or not. This is confusing as it is not clear that the setting also needs to be enabled at collaboration level.

Issue #292, opened 2022-08-23
[Change Request] The `vantage6` package should not be dependant on the `vantage6-client` package.


frankcorneliusmartin:
**Description**
The idea behind the `vantag6` package is to be extremely light weight. Now that it is dependant on the `vantage6-client` this is no longer the case. I also feel that it is only needed for uploading the public key to the server. 

Related #243

bartvanb:
Depends on #419

PR #425, opened 2022-12-07
Migrating client functions to common if used outside the client

bartvanb:
This is a partial solution for #292. We have decided not to fully solve this at present, but instead wait for implementation of #419.


frankcorneliusmartin (approved):
lgtm

PR #433, opened 2022-12-09
Added check so users cannot change usernames of other users

### LINKED ISSUE #396: [Change Request] Users should not be able to change usernames of other users
**Description**
It doesn't make sense for users to change usernames of other users so we should make that impossible. 

Optionally, we could completely fix usernames so that they cannot be changed after they are created. But that might make more sense when users register themselves (as proposed in #372)

bartvanb:
Fix #396


frankcorneliusmartin:
See my comment in the issue

bartvanb:
See my response in the issue ;-)

frankcorneliusmartin (approved):

PR #450, opened 2022-12-21
Implemented ping/pong via socket to establish which nodes are online/offline more reliably

### LINKED ISSUE #40: Nodes appear to be offline

bartvanb:
Fix #40

Note that the node status will not be correct when starting up a server, and will only be correct when the full polling cycle has completed (takes about a minute). 
Two choices may be made: assume 1) all nodes are offline or 2) assume the same status as the last time the server was online.
For now I chose option 2.

### DIFF
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -41,7 +41,8 @@ from vantage6.server.globals import (
     RESOURCES,
     SUPER_USER_INFO,
     REFRESH_TOKENS_EXPIRE,
-    DEFAULT_SUPPORT_EMAIL_ADDRESS
+    DEFAULT_SUPPORT_EMAIL_ADDRESS,
+    MAX_RESPONSE_TIME_PING
 )
 from vantage6.server.resource.common.swagger_templates import swagger_template
 from vantage6.server._version import __version__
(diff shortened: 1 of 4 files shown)


vantage6-server/vantage6/server/__init__.py
```diff
@@ -453,16 +463,56 @@ def start(self):
...
-            node.status = 'offline'
-            node.save()
-        # session.commit()
+    def __socket_pingpong_worker(self) -> None:
+        """
+        Send ping messages periodically to nodes over the socketIO connection
+        and set node status online/offline depending on whether they respond
+        or not.
+        """
+        PING_SLEEP = 60
```
frankcorneliusmartin, line 474: make this a setting in globals, just to be transparent

frankcorneliusmartin (changes requested):
lgtm 2 minor comments

vantage6-server/vantage6/server/__init__.py
bartvanb: Done

PR #451, opened 2022-12-22
Resolve circular import db session

### LINKED ISSUE #53: Resolve circular import
We have a circular import in the server code:
- vantage6.server.db imports vantage6.server.model.Base
- in the init file of vantage6.server.model, Base is imported from vantage6.server.model.base
- In vantage6.server.model.base, vantage6.server.db is imported

Currently, this is not causing issues because in https://github.com/IKNL/vantage6-server/blob/DEV3/vantage6/server/__init__.py, server.db is imported *before* DatabaseSessionManager and Database from vantage6.server.model.base. If this import order is changed, we get import errors.

We should refactor this code such that it does not have a circularity in the imports. At present, we have not determined a plan which code should move, so investigate that first.

bartvanb:
Fix #53 

Resolved this by putting the `session` variable in `db.py` into its own separate file, `session.py`.
Then it was much easier than expected...

Tested that nodes can still connect, API still works, and vserver shell still works

### DIFF
--- a/vantage6-server/tests_server/test_models.py
+++ b/vantage6-server/tests_server/test_models.py
@@ -4,13 +4,12 @@ import yaml
 import datetime
 
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm.exc import NoResultFound
 
 from vantage6.server.controller.fixture import load
 from vantage6.server.model.base import Database, DatabaseSessionManager
 from vantage6.server.globals import PACAKAGE_FOLDER, APPNAME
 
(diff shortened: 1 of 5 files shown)


frankcorneliusmartin (commented):
lgtm.

Did you test if the iPython shell still works? And is it still `db.` in the shell?

frankcorneliusmartin (approved):

bartvanb:
> Did you test if the iPython shell still works? And is it still db. in the shell?

Yes! See here:
image

PR #455, opened 2022-12-22
Bugfix/permission error config folder non existent

### LINKED ISSUE #1: Permission error when parent folder does not exist
If .config folder does not exist it crashes!
image

https://github.com/IKNL/vantage6-common/blob/97710e3abe3cd15598dd8ac4a999aa3c0960555e/vantage6/common/__init__.py#L92

bartvanb:
Fix #1. Quite an old issue apparently

This is how it looks like now (I just added `lala` to my config dir to check for non-existent dir):
image

People will get the same error if they don't have write permissions but a different warning above it

### DIFF
--- a/docs/user-documentation/use.rst
+++ b/docs/user-documentation/use.rst
@@ -287,7 +287,7 @@ set up and if they have access to the private key.
 
 .. code:: python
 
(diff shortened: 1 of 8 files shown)


frankcorneliusmartin (approved):
lgtm.

Shouldn't we create parent folders, or do we already do that?

PR #456, opened 2022-12-23
Assign rules to root automatically, extra check on role name in POST/PATCH role

### LINKED ISSUE #442: [Change Request] Add rules for collaboration scope in v4.0
**Description**
It is our plan to add a full collaboration scope in v4.x (see #245). It would be nice if we can add the required rules already in v4.0 so that root users will automatically possess these rules. 
If we create them later (in 4.x), they will be added automatically but server admins would have to assign these rules manually to themselves by changing the database.

bartvanb:
Fix #442 (in a different/better way than was originally intended)

- When a new rule is created, it is assigned to the Root user automatically

As this is done based on matching by name Root, also the following was changed:
- Users cannot create roles with any of the default role names
- Users cannot change names of existing roles into any of the default role names
- The names of default roles cannot be changed

Finally, this lead to the introduction of an Enum to keep track of the default role names, and use that wherever it should be used.

### DIFF
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -33,10 +33,8 @@ from vantage6.common import logger_name
 from vantage6.server.permission import RuleNeed, PermissionManager
 from vantage6.server.globals import (
     APPNAME,
-    CONTAINER_ROLE,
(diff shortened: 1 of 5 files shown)


frankcorneliusmartin (approved):
lgtm

PR #459, opened 2023-01-05
Updated docstrings and PEP8 fixes for part of node

bartvanb:


frankcorneliusmartin (approved):
lgtm.

However I also looked into writing (and typing) the click functions. Maybe remove them from the CLI or add them here as well?

PR #462, opened 2023-01-06
Added recommendations on how to secure VPN server to documentation

bartvanb:


docs/install/server.rst
```diff
@@ -119,6 +119,40 @@ After the installation is done, you need to configure the server to:
...
     protocol and port in order to communicate with each other.
 
+.. warning::
+    The EduVPN server should usually be available to the public internet to
+    allow all nodes to find it. Therefore, it should be properly secured, for
+    example by closing all public ports (except http/https).
+
+    Additionally, you may want to explicitly allow *only* VPN traffic between
+    nodes, and not between a node and the VPN server. You can achieve that by
+    executing the following ``iptables`` commands.
```
dsmits, line 129: If you execute the iptables commands they will be erased once you restart the machine.

I think you need to instruct the users to edit their iptables files.
```suggestion
    updating the firewall rules on your machine. 
    
    On Debian machines, these rules can be found in `/etc/iptables/rules.v4` and `/etc/iptables/rules.v6`, on CentOS, Red Hat Enterprise Linux and Fedora they can be found in `/etc/sysconfig/iptables` and `/etc/sysconfig/ip6tables`.  You will have to do the following:
```

dsmits (changes requested):
I think it needs to be clear that the iptables rules should be saved in a file

frankcorneliusmartin (approved):

PR #465, opened 2023-01-10
Added option to run algorithms without data.

mellesies:
Required minimal modification to wrapper. Is backwards compatible.


frankcorneliusmartin (approved):
lgtm, although I think we should remove the `data` argument from the RPC method then aswell (as it should not be used anyways)

PR #480, opened 2023-01-12
Improved the error message if node cannot authenticate to server beca…

### LINKED ISSUE #460: [Change Request] Better error message when server is not reachable
**Description**
When you put a wrong server URL into the node configuration file, the node will obvisously not be able to authenticate and gives the following message (over and over):

image

It would be nice to improve this message:
- [x]  Clear message that server cannot be found
- [x] If they are using localhost, write specific message that they should use `172.0.0.1` or `host.docker.internal` instead

bartvanb:
…use server cannot be reached

Fix #460

### DIFF
--- a/vantage6-node/vantage6/node/__init__.py
+++ b/vantage6-node/vantage6/node/__init__.py
@@ -460,7 +460,8 @@ class Node(object):
                 self.__print_connection_error_logs()
                 time.sleep(SLEEP_BTWN_NODE_LOGIN_TRIES)
             except Exception as e:
-                msg = 'Authentication failed. Retrying in 10 seconds!'
+                msg = ('Authentication failed. Retrying in '
+                       f'{SLEEP_BTWN_NODE_LOGIN_TRIES} seconds!')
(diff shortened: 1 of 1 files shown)


vantage6-node/vantage6/node/__init__.py
```diff
@@ -439,6 +456,9 @@ def authenticate(self) -> None:
                 self.log.warning(msg)
                 self.log.debug(e)
                 break
+            except requests.exceptions.ConnectionError:
+                self.__print_connection_error_logs()
+                time.sleep(SLEEP_BTWN_NODE_LOGIN_TRIES)
             except Exception as e:
                 msg = 'Authentication failed. Retrying in 10 seconds!'
```
frankcorneliusmartin, line 463: Could we edit the `10` here to reflect the setting? `SLEEP_BTWN_NODE_LOGIN_TRIES`

frankcorneliusmartin (changes requested):

frankcorneliusmartin (approved):
Thanks

PR #481, opened 2023-01-13
Upgraded dockerpy to 5.0.3, which allows us to automatically pull onl…

### LINKED ISSUE #473: [Bug] `docker.pull(.)` pulls all tags of an image
**Describe the bug**
`docker.pull` pulls all tags if not tag is specified, we need only need the `latest` in this case. (This is expected behevious from docker-py < 4.4.0: https://docker-py.readthedocs.io/en/4.2.0/images.html#docker.models.images.ImageCollection.pull).

**To Reproduce**
Send a task to a node with an image without tag.

**Expected behavior**
Only pull the `latest` tagged version of the image

**Possible solutions**
- docker-py fixed this in 4.4.x, see This has been fixed a newer release of docker-py, see https://docker-py.readthedocs.io/en/stable/change-log.html#id24. 
- We can add the `default` tag as a gaurd before pulling new images

bartvanb:
…y the 'latest' tag when doing docker.images.pull(some_image), instead of all tags

Upgrade to higher version of dockerpy (6.0.0) led to errors

Fix #473

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ contextlib2==0.5.5
 cryptography==3.3.2
 decorator==5.0.9
 dnspython==2.2.1
-docker==4.2.0
+docker==5.0.3
 eventlet==0.33.0
 flasgger==0.9.5
 Flask==2.0.2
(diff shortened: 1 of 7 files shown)


frankcorneliusmartin (approved):
lgtm

Issue #467, opened 2023-01-11
[Change Request] Support release branches in Github actions


frankcorneliusmartin:
**Description**
* Currently the version tag is applied in the code and commited in the main branch. This blocks us to release from different branches 
* Maybe there are other issues also

bartvanb:
We should probably create the release branch in the Github Action pipeline

frankcorneliusmartin:
Idea:

* pre-releases should be build from the dev3 branch
* only releases on the main branch

Pipeline should be able to accomodate the pre-releases from the dev3 branch (especially the version commit)

frankcorneliusmartin:
Pipeline failed for 3.9, see: https://github.com/vantage6/vantage6/actions/runs/4916657216

Issue #474, opened 2023-01-12
[Change Request] Replace `set-output` method from the Github actions


frankcorneliusmartin:
**Description**
Github actions complains about some of the commands we use:

```
[github-release](https://github.com/vantage6/vantage6/actions/runs/3900643513/jobs/6661546997)
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: marvinpinto/action-automatic-releases@latest. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

bartvanb:
Fix for 1 of the 4 warnings in the linked PR.

The rest seems to be related to the action from 'marvinpinto' we're using. There is an open PR there that would probably solve the deprecation warnings: https://github.com/marvinpinto/action-automatic-releases/pull/2/files

Issue #472, opened 2023-01-12
[Change Request] Improve the check-version gaurd of Github action


frankcorneliusmartin:
**Description**
The version `version/3.6.0/rc1` got accepted. This led to a broken pypi release. This test should be stricter.

PR #488, opened 2023-01-13
Feature/release pipeline updates

### LINKED ISSUE #485: [Change Request] Change order release pipeline; PyPi release later
**Description**
We've had the issue a few times that we had to skip a release, because the release pipeline crashed *after* the step where we had released the new version to PyPi. This step is the only irreversible step in the pipeline, so if we can do it later that would prevent such version skips.

Check if this works by moving the `build-n-release` block to a location after the docker builds but before the `notify` step.

bartvanb:
Fix #485, #472

Partial fix for #474


bartvanb:
And also fix #467

PR #489, opened 2023-01-16
Change/create node for org admin

bartvanb:
Org admins were not allowed to create nodes for their own organization, and neither were collaboration admins allowed to do so. This led to admins needed to set up a collaboration, which is not what you want


frankcorneliusmartin (approved):
lgtm

PR #490, opened 2023-01-16
Change/dont alert joins in all nodes room

bartvanb:


frankcorneliusmartin (changes requested):
See if you can make this more clean code and remove the `on_join_room` method which is no longer used.

frankcorneliusmartin (approved):
lgtm

PR #500, opened 2023-01-19
Improve dependencies between steps in release pipeline so that we are…

### LINKED ISSUE #485: [Change Request] Change order release pipeline; PyPi release later

bartvanb:
… sure everything works before we release to PyPi and notify discord

Fix #485

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -268,7 +268,9 @@ jobs:
 
     name: Build and release the project
     runs-on: ubuntu-latest
-    needs: version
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
lgtm

PR #501, opened 2023-01-19
Added new endpoint /vpn/algorithm/addresses to retrieve the IP addres…

### LINKED ISSUE #499: [Feature Request] Endpoint for getting VPN addresses of other algorithm containers in same run
**Problem description**
In the community meeting, the need became apparent for an endpoint that gives an algorithm the addresses+ports of the containers it can connect to via VPN.
We had first thought to include this in an improved implementation of the algorithm client (#493) but an additional endpoint is better as that will make the functionality more readily available to algorithms written in other programming languages than Python.

**Additional context**
Thanks to Florian van Daalen for the question at the community meeting!

bartvanb:
…ses + ports of other algorithm containers in the same task

Fix #499

### DIFF
--- a/vantage6-server/vantage6/server/resource/__init__.py
+++ b/vantage6-server/vantage6/server/resource/__init__.py
@@ -91,6 +91,8 @@ def only_for(types=['user', 'node', 'container']):
             # log.debug(f"Endpoint accessed as {g.type}")
 
             if g.type not in types:
+                # FIXME BvB 23-10-19: user gets a 500 error, would be better to
+                # get an error message with 403(?) code
                 msg = f"{g.type}s are not allowed to access {request.url} " \
                       f"({request.method})"
                 log.warning(msg)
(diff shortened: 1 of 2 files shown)


vantage6-server/vantage6/server/resource/__init__.py
```diff
@@ -91,6 +91,8 @@ def decorator(*args, **kwargs):
             # log.debug(f"Endpoint accessed as {g.type}")
 
             if g.type not in types:
+                # FIXME BvB 23-10-19: user gets a 500 error, would be better to
+                # get an error message with 403(?) code
```
frankcorneliusmartin, line 95: No I actually think this is a bad request? 

As it is not an authentication failure, but input that is not allowed making it a bad request in my opinion.

frankcorneliusmartin: The message bellow is then also incorrect. it should  be something like: 
`unrecognised type: {g.type}`

frankcorneliusmartin (changes requested):
lgtm.

However shouldnt we return an error code when no VPN is configured?

vantage6-server/vantage6/server/resource/__init__.py
bartvanb: I agree with the former, so I have updated the comment to say code 400. 
However, the second comment is incorrect: the `types` variable does not always contain all types. E.g. if `types=['user']` and a node attempts to use the endpoint, the message is correct (just verified this to be sure)

frankcorneliusmartin (approved):
lgtm

PR #505, opened 2023-01-24
Catch exceptions on checking if a database if a file.

### LINKED ISSUE #502: [Bug] Remote, non-file databases lead to error in mounting databases on windows
**Describe the bug**
A database configuration like this:

image

Leads to the following error on node startup:

image

**To Reproduce**
Start the node with a configuration of databases as detailed above. Note that this bug is also present when using e.g. `sql://` instead of `https://`

**Expected behavior**
Databases should be loaded as non-file databases

bartvanb:
If we get an exception, it is safe to assume the database is not a file

Fix #502


frankcorneliusmartin (approved):
lgtm

PR #510, opened 2023-01-27
Feature/updated algorithm client

### LINKED ISSUE #493: [Feature Request] Update the ContainerClient
**Problem description**
The `ContainerClient`, which is used by algorithms to communicate with the central server, is outdated. 
- [ ] It should have a similar interface as the UserClient, so e.g. `client.task.create()` instead of `client.create_task`
- [ ] There should be good support for obtaining the (VPN) addresses of other algorithm containers

bartvanb:
Fix #493


vantage6-client/vantage6/client/algorithm_client.py
```diff
@@ -0,0 +1,296 @@
...
+            description : str, optional
+                Description of the subtask
+            """
+            self.log.debug(f"Creating new subtask for {organization_ids}")
+
+            description = (
+                description or
+                f"task from container on node_id={self.host_node_id}"
+            )
+            name = subtask or "subtask"
```
frankcorneliusmartin, line 168: Why not make this as a default in the argument? `subtask: str = "subtask"`

vantage6-client/vantage6/client/algorithm_client.py
```diff
@@ -0,0 +1,296 @@
...
+                List of dictionaries containing the IP address and port number,
+                and other information to identify the containers.
+            """
+            results = self.request("vpn/algorithm/addresses", params={
+                "include_children": include_children,
+                "include_parent": include_parent,
+                "label": label
+            })
+
+            return results['addresses']
```
frankcorneliusmartin, line 231: Since you are not passing the output of the request directly, this could raise an exception. 

Check that the key is present or the status code. But I dont think we want to raise an exception here.

frankcorneliusmartin (changes requested):
Nice solution with the `use_new_client`.
Maybe something that we can in the decorator as well ^.^
`@AlgorithmClient(version=1)`

Ofc. I did not check the VPN changes

docs/algorithms/classic_tutorial.rst
```diff
@@ -309,7 +309,7 @@ We can distinguish two types of methods that a user can trigger:
     information.
 
 .. warning::
-    The ``client`` the master method receives is a ``ContainerClient`` which is
```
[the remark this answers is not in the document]
bartvanb: Yeah true, at least if we can find a way in v4. I guess I was a bit lazy in updating the docs - I'll update this

vantage6-client/vantage6/client/algorithm_client.py
```diff
@@ -0,0 +1,296 @@
...
+            ----------
+            input_ : bytes
+                Input to the task. Should be b64 encoded.
+            organization_ids : List[int], optional
+                List of organization IDs that should execute the task
+            name: str, optional
+                Name of the subtask
+            description : str, optional
+                Description of the subtask
+            """
```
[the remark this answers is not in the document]
bartvanb: Added

vantage6-client/vantage6/client/algorithm_client.py
bartvanb: good point

vantage6-client/vantage6/client/algorithm_client.py
bartvanb: Added check on the presence of the key, otherwise sending a message back to the algorithm.

frankcorneliusmartin (approved):
lgtm

PR #514, opened 2023-01-31
Change/termcolor dep removed

bartvanb:
Remove the `termcolor` dependency - it is only used to color the output of unit tests (which is not even working on Windows) and is not the most widely used package.

Note that this branch was created from the Python 3.10 branch


frankcorneliusmartin (approved):
Good find

PR #516, opened 2023-02-01
Change/specific operations send receive socket

### LINKED ISSUE #402: [Change Request] remove backwards compatibility socketIO in v4
**Description**
There have been recent updates to the sockets as discussed in #167. Some things have been left intact for v3 (petronas) backwards compatibility, but they should be removed in v4.
This issue is a todo list of what these things are so we don't forget them.

- [x] Change rules for socket to have operation 'send' and 'receive' instead of 'view'/'create'. Or consider creating operations such as 'kill' for kill task. (this is not possible in v3 because all operations must be either view/create/delete/edit.
- [x] Remove the property 'complete' in returning task json. This option is superseded by the 'status' field that gives more detail.

bartvanb:
Fix #402 

- Created new operations 'send' and 'receive' for socket rules
- Changed socket rules to reflect new operations
- Changes in endpoints to reflect changes in rules
- Removed the 'complete' property from db.Task which is superseded by 'status', but kept a function in its stead as that is used elsewhere
- Remove unused function

### DIFF
--- a/vantage6-server/vantage6/server/model/task.py
+++ b/vantage6-server/vantage6/server/model/task.py
@@ -37,9 +37,6 @@ class Task(Base):
     initiator = relationship("Organization", back_populates="created_tasks")
     init_user = relationship("User", back_populates="created_tasks")
 
-    def complete(self):
-        return all([r.complete for r in self.results])
-
     @hybrid_property
     def status(self) -> str:
         """
@@ -52,11 +49,7 @@ class Task(Base):
         """
         # TODO what if there are no result ids? -> currently returns unknown
         result_statuses = [r.status for r in self.results]
-        if all([status is None for status in result_statuses]):
-            # TODO remove in v4 (this is for backwards compatibility because
-            # task statuses where not present in <3.6)
-            return 'unknown'
-        elif any([has_task_failed(status) for status in result_statuses]):
+        if any([has_task_failed(status) for status in result_statuses]):
             return TaskStatus.FAILED.value
         elif TaskStatus.ACTIVE in result_statuses:
             return TaskStatus.ACTIVE.value
--- a/vantage6-server/vantage6/server/model/task.py
+++ b/vantage6-server/vantage6/server/model/task.py
@@ -37,9 +37,6 @@ class Task(Base):
     initiator = relationship("Organization", back_populates="created_tasks")
     init_user = relationship("User", back_populates="created_tasks")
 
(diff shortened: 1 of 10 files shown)


vantage6-server/vantage6/server/model/task.py
```diff
@@ -37,9 +37,6 @@ class Task(Base):
     initiator = relationship("Organization", back_populates="created_tasks")
     init_user = relationship("User", back_populates="created_tasks")
 
-    # TODO remove this property in v4. It is superseded by status but now left
-    # here for backwards compatibility with other v3 versions
-    @hybrid_property
```
frankcorneliusmartin, line 42: Shouldn't the entire function `complete` be deleted?

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -488,7 +488,7 @@ def __verify_container_permissions(container, image, collaboration_id):
             return False
 
         # check master task is not completed yet
-        if db.Task.get(container["task_id"]).complete:
+        if db.Task.get(container["task_id"]).complete():
```
frankcorneliusmartin, line 491: Why not use the `task.status` for this?

vantage6-server/vantage6/server/resource/token.py
```diff
@@ -312,7 +312,7 @@ def post(self):
                 HTTPStatus.UNAUTHORIZED
 
         # validate that the task not has been finished yet
-        if db_task.complete:
+        if db_task.complete():
```
frankcorneliusmartin, line 315: Same for this one

frankcorneliusmartin (changes requested):

vantage6-server/vantage6/server/model/task.py
bartvanb: I have done a new commit that should address this comment and the comments below.
This property has now been fully removed, and in other places now making use of the `has_task_finished()` function from vantage6-common. Also, I removed the same property in the `result` class.

vantage6-server/vantage6/server/resource/task.py
bartvanb: See above

vantage6-server/vantage6/server/resource/token.py
bartvanb: See above

frankcorneliusmartin (approved):

PR #517, opened 2023-02-01
Change/rename initiator

### LINKED ISSUE #416: [Change Request] `task.initiator_id` should be unambiguous
**Description**
We recently added `task.init_user_id` for the user that initiates the task. It is now a bit unclear what the `initiator_id` stands for (it stands for the organization that initiated the task). We should probably rename this to `init_org_id`, but that will cause backwards compatibility issues, so we should do so in v4.0

bartvanb:
Fix #416 

Rename `task.initiator_id` to `task.init_org_id` (and similar others)

### DIFF
--- a/vantage6-client/vantage6/client/__init__.py
+++ b/vantage6-client/vantage6/client/__init__.py
@@ -1675,7 +1675,7 @@ class UserClient(ClientBase):
             return self.parent.request(f'task/{id_}', params=params)
 
         @post_filtering()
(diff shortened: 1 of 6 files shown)


frankcorneliusmartin (changes requested):
lgtm, however this should not be merged into `main` i think?

bartvanb:
You're right, it is supposed to go into release/v4. I'll merge that now since you otherwise agreed with the code

PR #521, opened 2023-02-03
Feature/share some node config details

### LINKED ISSUE #12: request allowed images
As a new feature it would be nice if the user could request which images are allowed in a collaboration.
(somehow get this from the node configuration files)

bartvanb:
Nodes can share information from their node configuration with the central server. The following information is shared via a new socket event:

- which algorithms are allowed
- what the node's setting on encryption is

Also, an option is included with which the node owner can prevent that this information is shared.

Fix #12


bartvanb:
Superseded by #523

PR #523, opened 2023-02-07
Feature/share node config details v3

### LINKED ISSUE #12: request allowed images

bartvanb:
Same as #521 but now from a v3 branch so we can release it in v3.8

Fix #12


frankcorneliusmartin (approved):
lgtm, i treid really hard to find something to bitch about ;-)

PR #532, opened 2023-02-14
336 bug separate containers on same node cannot talk to eachother over n2n

### LINKED ISSUE #336: [Bug] Separate containers on same node cannot talk to eachother over n2n
**Describe the bug**
It is impossible to reach a separate algorithm container that is running on the same node using n2n communication.

When triggering a task on the same node it is possible to get the ip and port number back. However, they containers will not be able to connect.

**To Reproduce**
A simple way to check this is to use the n2n diagnostics algorithm.

This should run correctly:
```python
# Get organizations
active_nodes = client.node.list(is_online=True)
active_nodes = active_nodes['data']
org_ids = [n['organization']['id'] for n in active_nodes]

# Pick one node to run as master algorithm
master_node = org_ids[0]

# Pick the nodes that it should connect to
# In this case the organization of the master node is excluded
other_nodes = org_ids[1:]

n2nclient = N2NDiagnosticsClient(client)
task = n2nclient.echo(master_node, collaboration_id, other_nodes)

# Wait for results, etc.
```
The following will give an error:
```python
# Get organizations
active_nodes = client.node.list(is_online=True)
active_nodes = active_nodes['data']
org_ids = [n['organization']['id'] for n in active_nodes]

# Pick one node to run as master algorithm
master_node = org_ids[0]

# Pick the nodes that it should connect to
# In this case the organization of the master node is *included*
other_nodes = org_ids

n2nclient = N2NDiagnosticsClient(client)
task = n2nclient.echo(master_node, collaboration_id, other_nodes)

# Wait for results, etc.
```

**Expected behavior**
Connecting to an algorithm container on the same node should behave the same as connecting to a separate node.

**Details**
- Version: `node: '3.3.7', cli: '3.3.6'`

**Additional context**
Discussed with @frankcorneliusmartin

dsmits:
Added firewall rules for talking to algorithms that run on the same node.

You need to add special rules for when you want machines in a private network to talk to some other machine on the private network but with an external IP address. Apparently this is called [hairpinning](https://blog.lordvan.com/blog/learned-something-new-today-hairpin-nat-with-iptables-on-linux/)

 Closes #336

### DIFF
--- a/vantage6-node/vantage6/node/docker/vpn_manager.py
+++ b/vantage6-node/vantage6/node/docker/vpn_manager.py
@@ -323,14 +323,25 @@ class VPNManager(DockerBaseManager):
         for port in ports:
             port['port'] = vpn_client_port_options.pop()
 
+        vpn_ip = self._get_vpn_ip()
+
         # Set up forwarding VPN traffic to algorithm container
         command = 'sh -c "'
         for port in ports:
+            # Rule for directing external vpn traffic to algorithms
             command += (
                 'iptables -t nat -A PREROUTING -i tun0 -p tcp '
                 f'--dport {port["port"]} -j DNAT '
                 f'--to {algo_ip}:{port["algo_port"]};'
             )
+
+            # Rule for directing internal vpn traffic to algorithms
+            command += (
+                f'iptables -t nat -A PREROUTING -d {vpn_ip}/32 -p tcp '
+                f'--dport {port["port"]} -j DNAT '
+                f'--to {algo_ip}:{port["algo_port"]};'
+            )
+
             # remove the algorithm ports from the dictionaries as these are no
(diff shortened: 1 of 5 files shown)


frankcorneliusmartin (approved):
lgtm

vantage6-node/vantage6/node/docker/vpn_manager.py
```diff
@@ -460,6 +471,11 @@ def _get_isolated_interface(self):
                 isolated_interface = ip_interface
         return isolated_interface
 
+    def _get_vpn_ip(self):
```
bartvanb, line 474: How is this different from the `get_vpn_ip()` function that is already there?

dsmits (commented):
Looks good

PR #548, opened 2023-02-21
After algorithm run has completed, delete the port entries in the dat…

bartvanb:
…abase (as these ports are no longer open)


frankcorneliusmartin (approved):
lgtm

PR #554, opened 2023-02-23
Improved ping/pong between server and nodes

bartvanb:
- No longer starting ping/pong in a separate thread but in a socketio background task to prevent complications
- Whether nodes responded to ping is now based on last_seen state and not on a list. The latter is not reliable with multiple server instances


frankcorneliusmartin (approved):
much nicer

PR #556, opened 2023-02-27
Option to set allowed users and orgs on a node

### LINKED ISSUE #335: [Feature Request] Allow per user/organization i.o. images
**Problem description**
As a node I can allow certain images to be executed on my data.
Anyone in a collaboration then can execute these. 
Is there also a way I can allow a certain user or organization to use (whichever image on) my data?

So instead of saying 'I trust this image', I want to be able to say 'I trust this person', or 'I trust this organization'.

bartvanb:
Fix #335


vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -219,13 +221,54 @@ def is_docker_image_allowed(self, docker_image_name: str) -> bool:
...
         """
+        # in case of subtasks, don't check anymore, as parent has already
+        # been checked
+        if task_info['parent'] is not None:
+            return True
 
+        # check if algorithm matches any of the regex cases
+        allowed_algorithms = self._policies.get('allowed_algorithms')
+        if allowed_algorithms:
+            if isinstance(allowed_algorithms, str):
```
frankcorneliusmartin, line 240: why would we support this? I think it should _always_ be a list

vantage6-node/vantage6/node/server_io.py
```diff
@@ -226,3 +226,68 @@ def refresh_vpn_keypair(self, ovpn_file: str) -> bool:
...
+        bool
+            Whether or not the user is allowed to send a task to this node
+        """
+        # check if task-initating user id is in allowed users
+        for user in allowed_users:
+            try:
+                assert init_user_id == int(user)
+                return True
+            except Exception:
+                pass
```
frankcorneliusmartin, line 259: why not `user_allowed = init_user_id in allowe_users`?

frankcorneliusmartin: Is this some co-pilot suggestion?

frankcorneliusmartin (changes requested):

vantage6-node/vantage6/node/server_io.py
bartvanb: Oh wait... list comprehension wouldn't work if somebody gives a list like `['frank', '33']` for allowed users because it would fail on the first one. And in python `'1' == 1` resolves to `False` so we need the cast to int

vantage6-node/vantage6/node/server_io.py
```diff
@@ -226,3 +226,68 @@ def refresh_vpn_keypair(self, ovpn_file: str) -> bool:
...
+
+        # check if task-initiating org id is in allowed orgs
+        for org in allowed_orgs:
```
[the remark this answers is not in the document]
bartvanb: See above

vantage6-node/vantage6/node/docker/docker_manager.py
frankcorneliusmartin: Meh. Only card I have is that it adds ugly code.

vantage6-node/vantage6/node/server_io.py
bartvanb: Now changed this to cast to string instead of to int, so we don't need the try/except

PR #573, opened 2023-03-01
Updated server docstrings

bartvanb:
With this, all server functions should now have docstrings, with the exception of the Schema code, which will soon be refactored.

Also, I removed all the Union|Dict|List etc since we're now on 3.10

### DIFF
--- a/vantage6-server/vantage6/server/resource/pagination.py
+++ b/vantage6-server/vantage6/server/resource/pagination.py
@@ -1,18 +1,53 @@
 import math
 import logging
-
-from urllib.parse import urlencode
+import flask
 import sqlalchemy
 
+from urllib.parse import urlencode
+
 from vantage6.common import logger_name
+from vantage6.server import db
 
 module_name = logger_name(__name__)
 log = logging.getLogger(module_name)
 
 
 class Page:
+    """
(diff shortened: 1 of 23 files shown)


bartvanb:
remaining codacy issue seems to be a bug in codacy?

vantage6-server/vantage6/server/resource/pagination.py
```diff
@@ -96,7 +194,23 @@ def from_query(cls, query: sqlalchemy.orm.query, request):
         return cls(items, page_id, per_page, total, request)
 
     @classmethod
-    def from_list(cls, items: list, request):
+    def from_list(cls, items: list[db.Base],
+                  request: flask.Request) -> 'Pagination':
```
frankcorneliusmartin, line 198: Wat is deze? Om een circulaire import te voorkomen of e.d.?

frankcorneliusmartin (approved):

vantage6-server/vantage6/server/resource/pagination.py
bartvanb: This is a way to specify type hints, but better would be `from __future__ import annotations` which always removing the quotes. We agreed to leave it for now, but do this in the other way in the future

PR #575, opened 2023-03-01
Updated docstrings on the CLI

bartvanb:


vantage6/vantage6/cli/_version.py
```diff
@@ -7,7 +7,7 @@
     __build__ = json.load(fp)
 
 # Module version
-version_info = ((( 3, 8, 0, 'candidate', __build__, 0)))
+version_info = (((3, 8, 0, 'candidate', __build__, 0)))
```
frankcorneliusmartin, line 10: Be carefull here. I think this breaks the release pipeline (the set-version method needs this space). Can you check that by manually triggering the `make set-version` function?

frankcorneliusmartin (approved):

PR #586, opened 2023-03-08
Update manner in which version is updated

bartvanb:
- no space between `(` and major version
- no superfluous brackets


frankcorneliusmartin (approved):
Much nicer

PR #594, opened 2023-03-13
Changed output of verify_user_rules and changed its name to...

### LINKED ISSUE #512: [Change Request] Rename `verify_user_rules`
**Description**

Read this docstring and be amazed: 
```python
        """
        Check if an user, node or container has all the `rules`.

        Parameters
        ----------
        rules: List[Rule]
            List of rules that user is checked to have

        Returns
        -------
        Union[dict, bool]
            False if user has all rules, else a dict with a message
        """
```

False when the user has all the correct permissions...

bartvanb:
check_user_rules which is a bit simpler

Fix #512

It is a bit of a different change than I imagined but I think this works

### DIFF
--- a/vantage6-server/vantage6/server/permission.py
+++ b/vantage6-server/vantage6/server/permission.py
@@ -295,9 +295,9 @@ class PermissionManager:
         return result
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


frankcorneliusmartin (approved):
Much better, i can breath again

PR #598, opened 2023-03-16
rename result to run - more changes

bartvanb:
This should be everything to rename result to run

We only should have a discussion regarding the endpoint `/task/<id>/run`. This now gives all run details including the result. We may consider creating a separate `/task/<id>/result` just for the result


frankcorneliusmartin (approved):
One minor change

PR #599, opened 2023-03-17
Feature/filter task if has parent or not

### LINKED ISSUE #583: [Feature Request] Option to list only parent tasks
**Problem description**
I'd like a filter option on the get `/task` endpoint that allows me to get only tasks with no parents. This would e.g. allow filtering subtasks out of the UI more easily.

**Desired solution**
- [x] Have option to only get main tasks from GET `/task`
- [x] Also have this option in the Python client
- [x] Also use this option in the UI (people can still see subtasks by going to the parent task and then click on the subtask)
- [x] Consider an option to only get subtasks?

bartvanb:
Fix #583


frankcorneliusmartin (approved):
lgtm

PR #600, opened 2023-03-20
In 'vnode files', take the reformatted structure of the databases in …

bartvanb:
…the node configuration into account

I got the error `AttributeError: 'list' object has no attribute 'items'` when doing `vnode files` with v3.8.0, which is because we've changed the way the databases are configured in the node (including 'type' now)

Fixing that here


frankcorneliusmartin (approved):
lgtm

PR #602, opened 2023-03-20
Fix for checking whether to pull a newer node/server image

### LINKED ISSUE #507: [Bug] Pulling image is now based on `push_time` which is not updated if image is identical
**Describe the bug**
In `pull_if_newer()`, we pull the image if the remote timestamp is newer than the local one. Recently, we changed this (https://github.com/vantage6/vantage6/pull/482), and based the timestamp on the push time. 

It can occur that the local and remote image are identical, but that the remote push timestamp is newer. Then, the 'newer' remote image is pulled over and over again without ever updating the timestamp. This should be achieved somehow so that no pulls are attempted.

Note that when the images are identical, the pull time is short, so this issue is not very pressing for that reason.

bartvanb:
Since the dates were not identical between local and remote, extracting the sha256 digest and determining if they are the same based on that

Fix #507

### DIFF
--- a/vantage6-common/vantage6/common/docker/addons.py
+++ b/vantage6-common/vantage6/common/docker/addons.py
@@ -116,7 +116,7 @@ def registry_basic_auth_header(
 def inspect_remote_image_timestamp(
     docker_client: DockerClient, image: str,
     log: logging.Logger | ClickLogger = ClickLogger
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Elegant fix

PR #604, opened 2023-03-22
Settable config file location for the server

frankcorneliusmartin:
* default behaviour is preserved


bartvanb (approved):
Looks good. Are we merging this directly into main? If so, have you tested it?

PR #610, opened 2023-03-27
Feature/get algorithm addresses children or parent

bartvanb:
New addition to `/vpn/algorithm/addresses`. 

Previously, the algorithm developer would be presented with a list of addresses from which they had to determine which tasks were parents/children/same level. 
Now, there are new utility functions to help algo developers do this with a simple function call.


frankcorneliusmartin (approved):
Lgtm, let merge this to `3.9` and not directly to `main`

PR #613, opened 2023-03-29
Clarified what RPC stands for in Python client user docs

### LINKED ISSUE #608: [DOC] Add RPC description
**Description**
I am trying to setup a Python algorithm and am reading the following page:
https://docs.vantage6.ai/en/main/user/pyclient.html

This refers to RPC methods, but it does not describe what they are of how they need to be created. Also, it is not mentioned in earlier sections.

Can the documentation be updated with:
* what RPC stands for and why it is needed
* how to create an RPC method
* how the vantage engine knows that there is an RPC method and how it fetches it for task execution

bartvanb:
Fix #608


frankcorneliusmartin (approved):
lgtm

PR #616, opened 2023-03-30
Protected socket events algorithm_status_change and node_info_update …

### LINKED ISSUE #615: [Bug] Users can submit socket events that are not intended for them
**Describe the bug**
There are two socket events (`on_node_info_update` and  `on_algorithm_status_change`) that are intended for nodes but which a user connected to a socket event could also access. They can't really use these to do anything harmful, but it should not be possible nevertheless.

bartvanb:
…to be called by users

Fix #615


frankcorneliusmartin (approved):
lgtm

PR #618, opened 2023-03-30
Added reference to renamed docker_wrapper module so algorithms are ba…

bartvanb:
…ckwards compatible

### DIFF
--- a/vantage6-client/vantage6/tools/docker_wrapper.py
+++ b/vantage6-client/vantage6/tools/docker_wrapper.py
@@ -1,0 +1,2 @@
+# FIXME this is here for backwards compatibility, remove in v4+
+from vantage6.tools.wrapper import *


vantage6-client/vantage6/tools/docker_wrapper.py
```diff
@@ -0,0 +1,2 @@
+# FIXME this is here for backwards compatibility, remove in v4+
+from vantage6.tools.wrapper import *
```
frankcorneliusmartin, line 2: I would prefer to implicitly import the names

frankcorneliusmartin (changes requested):

vantage6-client/vantage6/tools/docker_wrapper.py
frankcorneliusmartin: This is also why codacy complains

frankcorneliusmartin (approved):
lgtm

PR #619, opened 2023-03-31
Change/proxy server logs

bartvanb:


docs/node/node_config.yaml
```diff
@@ -133,9 +133,6 @@ application:
     # levels: CRITICAL, ERROR, WARNING, INFO, DEBUG, NOTSET
     level:        DEBUG
 
-    # Filename of the log-file, used by RotatingFileHandler
-    file:         my_node.log
```
frankcorneliusmartin, line 137: Wouldn't a folder be appropiate here? This was possible before

frankcorneliusmartin (approved):
* Mention that the log folder is no longer settable in the release notes.

Issue #454, opened 2022-12-22
[Change Request] Remove backwards compatibility in v4


bartvanb:
**Description**
This issue intends to keep track of small backwards compatibility things that we have to fix/remove for the next major release (version 4)

- [x] `verbose` option for setting log level in the client
- [x] Update documentation for setting log level in the client

Issue #245, opened 2022-08-03
[Feature Request] Extend API security layers with Collaboration scope


frankcorneliusmartin:
**Problem description**
Currently it is not possible to edit settings on a Collaboration level. In practice this is usually needed and currently these kind of collaboration admins receive too much permission on the server.

**Desired solution**
Currently the main scopes we are using are `organization` and `global`. The idea is to add a scope `collaboration` for all endpoints. The main advantage is that non-superadmin users can manage collaborations without seeing everything on the server

bartvanb:
Resources todo:

- [ ] User: CRUD
- [ ] Organization: U      (delete doesn't exist, create makes no sense - same rule as global)
- [ ] Role: CRUD
- [ ] Node: CRUD
- [ ] Task: CRD                (no update, doesn't exist)
- [ ] Run: R
- [ ] Port: R

frankcorneliusmartin:
Tested with the UI:

✅ User: CRUD
✅ Organization: U (delete doesn't exist, create makes no sense - same rule as global)
🟥 You did create them for read
✅  Role: CRUD
✅  Node: CRUD
✅  Task: D (update doesn't exist, create and read makes no sense - same as organization scope)
 Task: modify organization delete, and create rule for deleting own tasks
 Task: change organization scope create+view to collaboration, as that is the actual level
 Run: R -> change org scope to col 
 Result: R -> change org scope to col
 Collaboration: add option to only edit/delete collaborations you are part of yourself
 ❓ Port endpoint will be deleted in v4 so we don't do that one
 Extend the permission checks on the requests generating lists of the resources above
 Check if this needs to be done for double endpoints - which do we keep in v4?
✅ Test / check if unit tests are OK
✅  Add new permissions to unit tests
✅ Add rules to default role -> checked all permissions of the default roles
✅ Adapt UI

bartvanb:
Check comments above
- [ ] Check read rule organizations
- [ ] Add collaboration scope for `/vpn/algorithm/addresses`

bartvanb:
- as indicated above, I created a 'read organizations from collaboration' rule while I indicated to only create 'update orgs from collab'. However, the rule should be created so all is well.
- We don't need to add a collaboration scope for `/vpn/algorithm/addresses` as the endpoint is only accessible for containers and the endpoint only returns addresses from the container's own task (so no rules are used).

PR #622, opened 2023-04-05
Change/remove backwards compatibility

bartvanb:
Addresses a number of points from #454


vantage6-client/vantage6/client/__init__.py
```diff
@@ -612,24 +612,20 @@ def __init__(self, parent) -> None:
 class UserClient(ClientBase):
     """User interface to the vantage6-server"""
 
-    def __init__(self, *args, verbose=False, log_level='debug',
-                 **kwargs) -> None:
+    def __init__(self, *args, log_level='debug', **kwargs) -> None:
```
frankcorneliusmartin, line 615: Should the default level be `info`?

vantage6-server/vantage6/server/resource/collaboration.py
frankcorneliusmartin, line 1: Could you elaborate on why these are compatibility endpoints?

frankcorneliusmartin (commented):
Ok

vantage6-server/vantage6/server/resource/collaboration.py
bartvanb: Not sure I understand the question, but e.g. endpoint `collaboration/<id>/nodes` is superseded by `/node?collaboration_id=<id>`

vantage6-client/vantage6/client/__init__.py
bartvanb: We can make it `info` but I think we'd need to go over the log statements and check if some debugs should not be info level. This was I guess the easy choice, and also the same as we make by default for the node/server logs

bartvanb:
We need to think about what to do with the permission checks of the (herein removed) double endpoints. Should we add the collaboration scope now in order to deal with that in a structured manner? Or keep them after all?

vantage6-client/vantage6/client/__init__.py
frankcorneliusmartin: Ok lets leave it like this for now. I created an issue to keep track of this: https://github.com/vantage6/vantage6/issues/692

frankcorneliusmartin:
> We need to think about what to do with the permission checks of the (herein removed) double endpoints. Should we add the collaboration scope now in order to deal with that in a structured manner? Or keep them after all?

I dont think that belongs within this PR. Lets work on that separately in https://github.com/vantage6/vantage6/issues/245

bartvanb:
The collaboration scope, with the permission checks discussed above, is now under review in #711. So we can also proceed with merging this (after conflicts are resolved ofc)

bartvanb:
Closing this, superseded by #741

Issue #597, opened 2023-03-16
[Feature Request] Enable GPU access by algorithm container


agjharms:
**Problem description**
In order to train a deep learning network more quickly, GPUs are often used. When using Vantage6 for this purpose, the algorithm needs to be able to access the GPUs. In the current official version of Vantage6 (3.8.1) this is not natively supported.

**Desired solution**
The desired outcome is that the Docker container containing the algorithm can access the GPUs on the system running the Node. It should be a configurable setting, set by default to 'no access', so algorithms are only granted access when the user gives permission to do so.

**Additional context**
Currently I have implemented a workaround for Vantage6 version 3 which adds a DeviceRequest for the GPUs to DockerTaskManager._run_algorithm() when an algorithm container is started. This is similar to what was done for Vantage6 version 2 in the Argos project ([link](https://github.com/AnanyaCN/argos-node/blob/master/vantage6/node/docker_manager.py)). This could be made configurable on the Node level, or perhaps on the Task level if the Vantage6 system allows for this.

PR #623, opened 2023-04-06
Feature/gpu support

agjharms:
# Feature/GPU Support

This pull request addresses issue #597 by adding a configuration option for the Node to allow the algorithm container access to the GPUs. This is done by creating and passing on a list of DeviceRequest objects from the docker-py package. Currently the only options are to block GPU access or allow access to all the GPUs on a machine, but there is an opportunity to make extend the configurability.

To interpret the Node configuration in DockerManager and then get the DeviceRequest to the algorithm container, I had to modify the arguments of DockerTaskManager. An alternative without this alteration would be to pass along the configuration through the algorithm environment variables, but seen as these configuration variables do not play a role _inside_ the container, I didn't think this was the best place to put this. 

It was not really clear to me which branch I should create the pull request for, dev3 seems to be somewhat behind main and I didn't want to make the choice for a release for you. If the PR should go somewhere else, please let me know.

### DIFF
--- a/vantage6-node/vantage6/node/docker/docker_manager.py
+++ b/vantage6-node/vantage6/node/docker/docker_manager.py
@@ -137,6 +137,7 @@ class DockerManager(DockerBaseManager):
         # name of the container that is running the node
         self.node_container_name = ctx.docker_container_name
 
+
         # login to the registries
         docker_registries = ctx.config.get("docker_registries", [])
         self.login_to_registries(docker_registries)
@@ -147,8 +148,13 @@ class DockerManager(DockerBaseManager):
         # keep track of linked docker services
         self.linked_services: List[str] = []
 
+        # set algorithm device requests
+        self.algorithm_device_requests = []
+        if config.get('algorithm_device_requests', False):
+            self._set_algorithm_device_requests(config['algorithm_device_requests'])
+
     def _set_database(self, databases: Union[Dict, List]) -> None:
-        """"
+        """
         Set database location and whether or not it is a file
 
         Parameters
@@ -199,6 +205,12 @@ class DockerManager(DockerBaseManager):
             self.databases[label] = {'uri': uri, 'is_file': db_is_file,
                                      'type': db_type}
         self.log.debug(f"Databases: {self.databases}")
+
+    def _set_algorithm_device_access(self, device_requests_config: Dict):
+        device_requests = []
+        if device_requests_config.get('gpu', False):
+            device_requests.append(docker.types.DeviceRequest(count=-1, capabilities=[['gpu']]))
+        self.algorithm_device_requests = device_requests
 
     def create_volume(self, volume_name: str) -> None:
         """
@@ -414,7 +426,8 @@ class DockerManager(DockerBaseManager):
             isolated_network_mgr=self.isolated_network_mgr,
             databases=self.databases,
             docker_volume_name=self.data_volume_name,
-            alpine_image=self.alpine_image
+            alpine_image=self.alpine_image,
+            device_requests=self.algorithm_device_requests
         )
         database = database if (database and len(database)) else 'default'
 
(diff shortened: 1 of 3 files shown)


vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -200,6 +206,21 @@ def _set_database(self, databases: Union[Dict, List]) -> None:
                                      'type': db_type}
         self.log.debug(f"Databases: {self.databases}")
 
+    def _set_algorithm_device_requests(self, device_requests_config: Dict):
```
frankcorneliusmartin, line 209: * In python 3.10+ thypes can be simply `dict`.
* set return type to `None`

vantage6-node/vantage6/node/docker/task_manager.py
```diff
@@ -41,7 +41,8 @@ def __init__(self, image: str, vpn_manager: VPNManager, node_name: str,
                  result_id: int, task_info: Dict, tasks_dir: Path,
                  isolated_network_mgr: NetworkManager,
                  databases: dict, docker_volume_name: str,
-                 alpine_image: Union[str, None] = None):
+                 alpine_image: Union[str, None] = None,
+                 device_requests: Union[List, None] = None):
```
frankcorneliusmartin, line 45: Also something cool in python 3.10+, you can write `Union[str, None]` as `str | None` much more readable.

vantage6-node/vantage6/node/docker/task_manager.py
```diff
@@ -427,7 +438,7 @@ def _setup_environment_vars(self, algorithm_env: Dict = {},
                              "exist. Available databases are: "
                              f"{self.databases.keys()}. This is likely to "
                              "result in an algorithm crash.")
-            self.debug(f"User specified database: {database}")
+            self.log.debug(f"User specified database: {database}")
```
frankcorneliusmartin, line 441: Thx

frankcorneliusmartin (changes requested):

frankcorneliusmartin:
* I changed the merge branch to `release/3.9`
* Added the issue to the project board 
* @agjharms Could you resolve the merge conflict
* @agjharms Could you fix the minor comments (/ question)

Other than that I think this looks good. Development for `3.9` is complete from our end so I expect that we release it before the end of the month.

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -200,6 +206,21 @@ def _set_database(self, databases: Union[Dict, List]) -> None:
...
+
+        Parameters
+        ----------
+        device_requests_config: Dict
+           A dictionary containing configuration options for device access. Supported keys:
+           - 'gpu': A boolean value indicating whether GPU access is required.
+        """
+        device_requests = []
+        if device_requests_config.get('gpu', False):
+            device_requests.append(docker.types.DeviceRequest(count=-1, capabilities=[['gpu']]))
```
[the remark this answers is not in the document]
agjharms: Yes, with '-1' access is given to all GPUs. Currently I think that Vantage6 nodes are often run in a separate (virtual) machine, so it's not a big deal to allot all of the resources to the node. I can imagine a situation in the future where a user has multiple Vantage6 nodes running on a machine with multiple GPUs, and wishes to designate specific GPUs to each node. In that case you could use the `device_ids` argument for DeviceRequest to pick out specific GPUs. I didn't have the opportunity to look into making this configurable in a nice way, so I reduced it to a boolean option for now with the option to extend it if the need arises.

agjharms:
Thanks for the feedback. I have addressed the points that you mentioned. I tried to rebase the `release/3.9` branch on the feature branch I had in my fork to accommodate the change in target branches, but that seems to not have been the right approach. I will fix that.

vantage6-node/vantage6/node/docker/task_manager.py
```diff
@@ -101,6 +105,12 @@ def __init__(self, image: str, vpn_manager: VPNManager, node_name: str,
         # keep track of the task status
         self.status: TaskStatus = TaskStatus.INITIALIZING
 
+        # set device requests
+        self.device_requests = []
+        if device_requests is not None:
```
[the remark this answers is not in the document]
agjharms: You're right that `if device_requests` is a bit cleaner. Thanks for the heads-up on the `is` operator.

frankcorneliusmartin (approved):

PR #642, opened 2023-05-08
Feature/control individual log levels

frankcorneliusmartin:


docs/node/node_config.yaml
```diff
@@ -165,6 +165,30 @@ application:
...
+        level: warning
+
+  # some additional debug flags
+  debug:
+
+    # sets the socketio client into debug mode
+    socketio: true
+
+    # sets the Flask app used for the LOCAL proxy service into debug mode
+    proxy_server: true
```
bartvanb, line 189: Shouldn't we have these by default at false?

docs/server/yaml/server_config.yaml
```diff
@@ -67,6 +67,31 @@ environments:
...
+
+    # some additional debug flags
+    debug:
+
+      # sets the socketio server into debug mode
+      socketio: true
+
+      # sets the Flask app in debug mode
+      flask: true
+
```
bartvanb, line 94: same as above

docs/node/node_config.yaml
bartvanb: And would it make sense to add 1 indent? And should we list all the options for packages? (I think it's just 'flask' missing right?)

bartvanb (changes requested):
Good improvement, just a few minor comments

docs/node/node_config.yaml
frankcorneliusmartin: Made them by default False. Identation does not make sense to me as the logging session refers to the configuration of the  python logger.

PR #643, opened 2023-05-08
Removed configuration environments as these were not used much

### LINKED ISSUE #260: [Feature Request] Get rid of environments in config files
**Problem description**
Vantage6 offers the option to define multiple environments (dev/test/prod/application) for a configuration. This was originally intended to support people running the same node with different configurations.
However, in practice this feature is hardly used and it complicates the CLI code. It may therefore be a good idea to remove it.
If people then want to use multiple environments, they can always define multiple configuration files for the same node.

**Desired solution**
Remove the environments from the CLI code for node and server. Possibly refactor the code after cleanup.

bartvanb:
Fix #260


bartvanb:
codacy may be ignored here

docs/node/configure.rst
```diff
@@ -63,9 +66,8 @@ You can download this file here: :download:`node_config.yaml`
     - ``prod``: Production environment. The version of the proposed solution
       where the final analyses are executed.
 
-    You can also specify the key ``application`` if you do not want to specify
-    one of the environments. This is also done in the example configuration
-    shown above.
+    You may also specify the key ``application`` if you do not want to specify
+    one of the environments.
```
frankcorneliusmartin, line 70: I dont think 
> if you do not want to specify one of the environments
is correct.  

However does this mean you still can use `application` key at the top level, but it's optional?

frankcorneliusmartin (changes requested):
( ´･･)ﾉ(._.`) Bye environments

docs/node/configure.rst
bartvanb: This is meant also to be deprecated, as is stated 10 lines above. But for clarity I now clarified that this is deprecated as well

docs/server/configure.rst
```diff
@@ -73,10 +76,12 @@ You can download this file here: :download:`server_config.yaml <yaml/server_conf
     - ``prod`` Production environment. The version of the proposed solution
       where the final analyses are executed.
 
-    You can also specify the key ``application`` if you do not want to specify
+    You may also specify the key ``application`` if you do not want to specify
```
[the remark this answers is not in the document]
bartvanb: see above

PR #645, opened 2023-05-09
No longer wrongly prepend '/mnt/' to non-file-based databases when ru…

bartvanb:
…nning in docker


frankcorneliusmartin (approved):
lgtm, although this should go to main

PR #673, opened 2023-05-17
Fix being able to connect to algorithms on same node on startup of VPN

### LINKED ISSUE #671: [Bug] VPN communication seems not to work in all cases
**Describe the bug**
When I run the `test_on_v6.py` script from https://github.com/vantage6/v6-node-to-node-diagnostics, in a collaboration with 2 online nodes, I expect 2 echo's to work, but instead I only see one:

(Logs from script)
```
info > Polling results for port numbers...
info > [{'ip': '10.76.0.162', 'port': 49159, 'label': None}, {'ip': '10.76.0.160', 'port': 49155, 'label': None}]
info > Echoing to 2 algorithms...
info > Sending message to 10.76.0.162:49159
info > Checking echo on 10.76.0.162:49159
info > Timeout! Skipping to next address.
info > Sending message to 10.76.0.160:49155
info > Checking echo on 10.76.0.160:49155
info > Succeeded echos: [True]
info > Writing output to /mnt/data/task-000000165/output
```

**Expected behavior**
I don't expect a timeout, and 2 succeeded echoes

**Additional context**
Might be the case that 2 containers on the same node don't work - maybe related to https://github.com/vantage6/vantage6/issues/336

bartvanb:
In the previous situation, the isolated subnet was not always immediately available when the eth1_interface was, so now it is included in the while loop

This explains the behaviour we saw where sometimes it worked and sometimes it didn't

Fix #671

### DIFF
--- a/services/vpn-client/entry.sh
+++ b/services/vpn-client/entry.sh
@@ -14,16 +14,16 @@ iptables -A FORWARD -i eth1 -o eth1 -j ACCEPT
 iptables -t nat -A POSTROUTING -o tun0 -j MASQUERADE
 
 # Check isolated network interface. Wait until it is up.
-eth1_interface=""
-
-while [ "$eth1_interface" = "" ];do
+isolated_subnet=""
+while [ "$isolated_subnet" = "" ];do
   eth1_interface=$(ip addr show eth1)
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Good find! 🎄

Issue #683, opened 2023-05-22
[Change Request] Update ClientMockProtocol


harmbuisman:
**Description**
The ClientMockProtocol is a nice feature that allows to locally test part of the algorithm. It would be helpful if it would get some maintenance and some additional features:

- remove the print statement from the mockclient.get_results() method. It makes more sense to print the output of the call rather than auto-assume a user wants a subresult (which could be bytes) printed
- unify the apis with the algorithmclient, such that it also has e.g. .task.create
- allow to provide the data itself directly into the datasets argument of the ClientMockProtocol constructor. This way we can use mock data initialized in Python for e.g. unit tests rather than depending on files on a harddisk (which is now also limited to csv)
- provide a way to extract the mock node id from within the mock method execution such that we can also simulate writing to the docker container temp folder for each unique node. e.g. an option would be to set an attribute org_id/node_id on the ClientMockProtocol that is updated before the call to method(). --> ideally this attribute is also added to the algorithmclient itself (now the only option to get it there is to use the jwt token, right?)

bartvanb:
Preliminary work on branch `feature/mock-client-update`

- [ ] Made interface similar to other clients
- [ ] Removed print statement detailed above
- [ ] Added new functionality to mock eg VPN features, but also node IDs
- [ ] Make MockClient compatible with multiple data sources
- [ ] Adapt MockClient to v4 - or put this in the task list for v4

harmbuisman:
To confirm, with the new changes I can change the following

Now:
MockClient
``` 
child_task = mockclient.create_new_task(
                    organization_ids=org_ids,
                    input_=input_,
                )
```

AlgorithmClient
```
 child_task = algorithm_client.task.create(
                organization_ids=child_nodes,
                name=str(transformer),
                description=str(transformer),
                input_=self.get_task_input_(server_round),
            )

```

to something like this?

```
def create_task(client, organization_ids, input_, name="", description="")
           return client.task.create(
                organization_ids=organization_ids,
                name=name,
                description=description,
                input_=input_,
            )

# MockClient
child_task = create_task(mockclient, ...)

# AlgorithmClient
child_task = create_task(algorithm_client, ...)

```

bartvanb:
Yes it will be similar to that indeed. Only the `create_task()` function will look a bit different as the `client.task.create()` args will be a bit different for the MockClient and AlgorithmClient (e.g. MockClient doesn't accept a `name` or `description` arg)

[...]

PR #702, opened 2023-06-01
Feature/mock client update

bartvanb:
A first setup of the new ``MockAlgorithmClient``.

Partial implementation of #683 - we still need to expand on it with e.g. VPN services in v4+

### DIFF
--- a/vantage6-client/vantage6/tools/mock_client.py
+++ b/vantage6-client/vantage6/tools/mock_client.py
@@ -146,3 +146,176 @@ class ClientMockProtocol:
                 "domain": f"mock-{i}.org",
             })
         return organizations
+
+
+# TODO in v4+, rename to ClientMockProtocol?
+class MockAlgorithmClient:
+    """
+    The MockAlgorithmClient is mimics the behaviour of the AlgorithmClient. It
+    can be used to mock the behaviour of the AlgorithmClient and its
+    communication with the server.
+
+    Parameters
+    ----------
+    datasets : list[str]
+        A list of paths to the datasets that are used in the algorithm.
+    module : str
+        The name of the module that contains the algorithm.
+    """
+    # TODO not only read CSVs but also data types
+    def __init__(self, datasets: list[str], module: str) -> None:
+        self.n = len(datasets)
+        self.datasets = []
+        for dataset in datasets:
+            self.datasets.append(
+                pandas.read_csv(dataset)
+            )
+
+        self.lib = import_module(module)
+        self.tasks = []
+
+        self.task = self.Task(self)
+
+    class SubClient:
+        """
+        Create sub groups of commands using this SubClient
+
+        Parameters
+        ----------
+        parent : MockAlgorithmClient
+            The parent client
+        """
+        def __init__(self, parent) -> None:
+            self.parent: MockAlgorithmClient = parent
+
+    class Task(SubClient):
+        """
+        Task subclient for the MockAlgorithmClient
+        """
+        def create(self, input_: dict, organization_ids: list[int]) -> int:
+            """
+            Create a new task with the MockProtocol and return the task id.
+
+            Parameters
+            ----------
+            input_ : dict
+                The input data that is passed to the algorithm. This should at
+                least  contain the key 'method' which is the name of the method
+                that should be called. Another often used key is 'master' which
+                indicates that this container is a master container. Other keys
+                depend on the algorithm.
+            organization_ids : list[int]
+                A list of organization ids that should run the algorithm.
+
+            Returns
+            -------
+            int
+                The id of the task.
+            """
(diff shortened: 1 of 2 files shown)


vantage6-client/vantage6/tools/mock_client.py
```diff
@@ -146,3 +149,433 @@ def get_organizations_in_my_collaboration(self) -> list[dict]:
...
+    can be used to mock the behaviour of the AlgorithmClient and its
+    communication with the server.
+
+    Parameters
+    ----------
+    datasets : list[dict]
+        A list of dictionaries that contain the datasets that are used in the
+        mocked algorithm. The dictionaries should contain the following:
+        {
+            "database": str,
```
frankcorneliusmartin, line 167: Can wel also acces a pandas *dataframe* here? Then this mock client can be used in testing =]

vantage6-client/vantage6/tools/mock_client.py
```diff
@@ -146,3 +149,433 @@ def get_organizations_in_my_collaboration(self) -> list[dict]:
...
+            "type": str,
+            "input_data": dict
+        }
+        where database is the path/URI to the database, type is the database
+        type (as listed in node configuration) and input_data contains
+        the input data that is normally passed to the algorithm wrapper.
+    module : str
+        The name of the module that contains the algorithm.
+    image : str, optional
```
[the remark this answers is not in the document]
frankcorneliusmartin: I understand from the `AlgorithmClient` that these are obtained from the `token`, however they are also not supplied as arguments and are not used in this function. So I would remove them? And supply default values for the properties in the object? This so to avoid confusion about the input parameters to this class.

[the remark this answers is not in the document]
frankcorneliusmartin: Alternatively we need to be more specific in the argument description about this. This is also the case for `database`, `node_id`, `collaboration_id` and `organization_id`.

vantage6-client/vantage6/tools/mock_client.py
```diff
@@ -146,3 +149,433 @@ def get_organizations_in_my_collaboration(self) -> list[dict]:
...
+                    client_copy = deepcopy(self.parent)
+                    client_copy.node_id = org_id
+                    client_copy.organization_id = org_id
+                    result = method(self.parent, data, *args, **kwargs)
+                else:
+                    result = method(data, *args, **kwargs)
+
+                idx = 999  # we dont need this now
+                results.append(
+                    {"id": idx, "result": pickle.dumps(result)}
```
frankcorneliusmartin, line 300: Cant we remove the pickles?

vantage6-client/vantage6/tools/mock_client.py
```diff
@@ -146,3 +149,433 @@ def get_organizations_in_my_collaboration(self) -> list[dict]:
...
+            Returns
+            -------
+            list[dict]
+                The results of the task.
+            """
+            task = self.parent.tasks[task_id]
+            results = []
+            for result in task.get("results"):
+                # TODO in v4+, this is no longer a pickle
+                res = pickle.loads(result.get("result"))
```
frankcorneliusmartin, line 373: Also here

vantage6-client/vantage6/tools/mock_client.py
```diff
@@ -146,3 +149,433 @@ def get_organizations_in_my_collaboration(self) -> list[dict]:
                 "domain": f"mock-{i}.org",
             })
         return organizations
+
+
+# TODO in v4+, rename to ClientMockProtocol?
+class MockAlgorithmClient:
+    """
+    The MockAlgorithmClient is mimics the behaviour of the AlgorithmClient. It
```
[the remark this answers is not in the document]
bartvanb: done

vantage6-client/vantage6/tools/mock_client.py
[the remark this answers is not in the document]
bartvanb: - I removed `image` and `database` as arguments, but still providing them as attributes - that way the matching attributes of the AlgorithmClient are Mocked
- `node_id` was renamed (back) to `host_node_id` to match the attribute in real client
- `organization_id` and `collaboration_id` are matching the real attributes
- The id's can still be set (optionally) by the user creating the MockClient

vantage6-client/vantage6/tools/mock_client.py
bartvanb: Let's not do that in this PR, since this is going into 3.10. I added this to the checklist of #454

vantage6-client/vantage6/tools/mock_client.py
bartvanb: See previous

PR #704, opened 2023-06-02
Update node VPN IP on VPN connect and disconnect, instead of updating…

### LINKED ISSUE #520: [Bug] When a node comes online, VPN IP address is not always updated
**Describe the bug**
A node that uses VPN gets a certain IP address to communicate over VPN. A node without VPN has no IP address registered.

When a node first connects with VPN, but later connects without VPN, the IP address that is stored in the database is still the (then incorrect) VPN IP address. This should be fixed.

Also, we should check if the VPN IP is always saved to the database at the right times. It should be updated when the VPN connects or disconnects.

bartvanb:
… it on task startup

Also, renamed server_io (NodeClient) variable to client

Fix #520


frankcorneliusmartin (approved):
lgtm

PR #706, opened 2023-06-02
Startup the node with the same image version as the server is running on

### LINKED ISSUE #700: [Change Request] Node should adapt the version it pulls
**Description**
If a node is connecting to a server with version `x.y.z`, the node currently retrieves `x.latest`. However, we would like to change this to `x.y.latest`, as these are much more likely to be properly compatible.

- [x] Adapt pipeline to create also docker image that with `major.minor` which is always the latest of the patch versions
- [x] Adapt node to pull that one by default
- [x] Adapt docs
- [x] Add warning in the CLI if the node version is not equal to the CLI version
- [x] Check that CLI works if we can't connect to the server - node should still start if server is not running

bartvanb:
Also, delete an extra dot in the version for pre-releases

Fix #700 

**Don't merge this yet, we may need to upgrade the policy (now using exact server version**

### DIFF
--- a/vantage6/vantage6/cli/_version.py
+++ b/vantage6/vantage6/cli/_version.py
@@ -13,7 +13,7 @@ version_info = (3, 9, 0, 'candidate', __build__, 0)
 _specifier_ = {'alpha': 'a', 'beta': 'b', 'candidate': 'rc', 'final': ''}
 version = f'{version_info[0]}.{version_info[1]}.{version_info[2]}'
 pre_release = '' if version_info[3] == 'final' else \
-  '.'+_specifier_[version_info[3]]+str(version_info[4])
+  _specifier_[version_info[3]]+str(version_info[4])
 post_release = '' if not version_info[5] else f'.post{version_info[5]}'
 
 # Module version accessible using thomas.__version__
--- a/vantage6/vantage6/cli/node.py
+++ b/vantage6/vantage6/cli/node.py
@@ -35,7 +35,8 @@ from vantage6.common.globals import (
     STRING_ENCODING,
     APPNAME,
     DEFAULT_DOCKER_REGISTRY,
-    DEFAULT_NODE_IMAGE
+    DEFAULT_NODE_IMAGE,
+    DEFAULT_NODE_IMAGE_WO_TAG
 )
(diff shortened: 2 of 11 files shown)


vantage6/vantage6/cli/_version.py
```diff
@@ -13,7 +13,7 @@
 _specifier_ = {'alpha': 'a', 'beta': 'b', 'candidate': 'rc', 'final': ''}
 version = f'{version_info[0]}.{version_info[1]}.{version_info[2]}'
 pre_release = '' if version_info[3] == 'final' else \
-  '.'+_specifier_[version_info[3]]+str(version_info[4])
+  _specifier_[version_info[3]]+str(version_info[4])
```
frankcorneliusmartin, line 16: I dont understand how this worked before? I dont seem to find this '.' in the pypi releases?

vantage6/vantage6/cli/node.py
```diff
@@ -350,6 +351,32 @@ def cli_node_start(name: str, config: str, environment: str,
...
+            )['version']
+            major_minor = '.'.join(version.split('.')[:2])
+            image = (f"{DEFAULT_DOCKER_REGISTRY}/{DEFAULT_NODE_IMAGE_WO_TAG}"
+                     f":{major_minor}")
+        except Exception:
+            warning("Could not determine server version. Using default node "
+                    "image")
+            pass  # simply use the default image
+
+        if major_minor and not __version__.startswith(major_minor):
```
frankcorneliusmartin, line 372: <3

frankcorneliusmartin (approved):

vantage6/vantage6/cli/_version.py
bartvanb: Pypi uses the version as specified in the tag that we push, not derived from the code. I checked the code, and `__version__` is only used in setup.py and in printing versions in several places (eg. /version endpoint). Since we are not using setup.py but rather requirements.txt in our installation pipeline, there is no reason that something shouldn't function properly if the version is wrong

vantage6/vantage6/cli/node.py
bartvanb: :)

Issue #719, opened 2023-06-19
[Bug] `AlgorithmClient` object has no attribute `organization_id`


frankcorneliusmartin:
**Describe the bug**
This line requires `self.whoami` to be set, however this is only done when the authentication method is triggered. This does not happen for algorithm containers.

https://github.com/vantage6/vantage6/blame/7dd509c35f0cf19ca18e7d84462e3aee5a093100/vantage6-client/vantage6/client/algorithm_client.py#L49-L50

PR #720, opened 2023-06-19
Obtain organization id in AlgorithmClient from container token instea…

bartvanb:
…d of (unset) attribute in whoami


frankcorneliusmartin:
Fixes #719 .

PR #722, opened 2023-06-19
Change/sync clients

### LINKED ISSUE #669: [Change Request] Harmonize interfaces or split them more to avoid inconsistency bugs
There are several places where the docs mention client.task.xyz, but "client" actually means various things, e.g. a UserClient instance, an AlgorithmClient instance or the module client. This is not always clear.

What adds to difficulty in implementation is that the interfaces of some of the methods are different. 

- client.task.create requires different arguments depending on what "client" is, but arguments that intend to do the same have different names, e.g. organizations versus organization_ids and input versus input_

- client.result.get requires a result_id in the UserClient, but a task_id in the AlgorithmClient, which was not straightforward to debug. Also, they return diffent things, for the UserClient it returns a dict, while for the AlgorithmClient it returns the result value

It would help if either the interfaces are harmonized, such that they are exactly the same, at places where it is helpful to not care about whether you have a userclient/algorithclient/module and are made more distinct at places where it *does* matter which type of client you are working with.

bartvanb:
Fix #669

### DIFF
--- a/vantage6-client/vantage6/client/__init__.py
+++ b/vantage6-client/vantage6/client/__init__.py
@@ -1834,7 +1834,7 @@ class UserClient(ClientBase):
 
         @post_filtering(iterable=False)
         def create(self, collaboration: int, organizations: list, name: str,
-                   image: str, description: str, input: dict,
+                   image: str, description: str, input_: dict,
                    databases: list[str] = None) -> dict:
             """Create a new task
 
@@ -1851,7 +1851,7 @@ class UserClient(ClientBase):
                 Docker image name which contains the algorithm
             description : str
                 Human readable description
-            input : dict
+            input_ : dict
                 Algorithm input
             databases: list[str], optional
                 Database names to be used at the node
@@ -1861,9 +1861,10 @@ class UserClient(ClientBase):
             dict
                 [description]
             """
+            # TODO v4+ integrate post_task function in this function
             if databases is None:
                 databases = ['default']
-            return self.parent.post_task(name, image, collaboration, input,
+            return self.parent.post_task(name, image, collaboration, input_,
                                          description, organizations,
                                          databases)
 
--- a/vantage6-client/vantage6/client/algorithm_client.py
+++ b/vantage6-client/vantage6/client/algorithm_client.py
@@ -1,3 +1,4 @@
+from typing import Any
 import jwt
 import json as json_lib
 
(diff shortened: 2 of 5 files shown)


vantage6-client/vantage6/client/__init__.py
```diff
@@ -1861,9 +1861,10 @@ def create(self, collaboration: int, organizations: list, name: str,
             dict
                 [description]
```
frankcorneliusmartin, line 1862: Since your at it, could you add a description here?

vantage6-client/vantage6/client/algorithm_client.py
```diff
@@ -238,9 +281,9 @@ def create(
 
             Parameters
             ----------
-            input_ : bytes
+            input : bytes
```
frankcorneliusmartin, line 284: Shouldnt have removed the `_`

vantage6-client/vantage6/client/algorithm_client.py
```diff
@@ -252,10 +295,9 @@ def create(
             dict
                 Dictionary containing information on the created task
             """
-            if organization_ids is None:
-                organization_ids = []
-            self.parent.log.debug(
-                f"Creating new subtask for {organization_ids}")
+            if organizations is None:
```
frankcorneliusmartin, line 298: Use `if not organizations:`. I dont like the `is` operator as it checks if you are using the same object rather than the same value.

vantage6-client/vantage6/tools/decorators.py
frankcorneliusmartin, line 1: I am not sure how this works by reading the code. Is this so that the algorithm developer only patches the decorators with these special decorators? Or does he or she still need to supply additional arguments?

frankcorneliusmartin (changes requested):

vantage6-client/vantage6/client/__init__.py
bartvanb: done

vantage6-client/vantage6/client/algorithm_client.py
bartvanb: done

vantage6-client/vantage6/client/algorithm_client.py
bartvanb: Done. Not sure if I agree in this particular case but it's not too important

vantage6-client/vantage6/tools/decorators.py
bartvanb: The idea is that in the regular case you have:

```
@client
def some_func(client, *args, **kwargs)
```

Then the AlgorithmClient calls `some_func` with only the args and kwargs and the client is inserted. 

Now, the MockAlgorithmClient does the following call: `some_func(mock_client=my_mock_client_instance, *args, **kwargs)`. And because of these adaptations to the decorators, then the mock client instance is inserted instead of a regular AlgorithmClient.

The only backdraw of this is that algorithm functions will not be able to have arguments called `mock_client` or `mock_data` - but I would say that would be very coincendental

PR #732, opened 2023-06-22
Bugfix/slow start algorithm with vpn

### LINKED ISSUE #681: [Bug] starting container with vpn takes a lot of time
**Describe the bug**
Starting a container on a node that has vpn enabled takes about 3 minutes more than starting it on the same node with vpn disabled.

**To Reproduce**
With vpn_subnet in the node config, start the node, run a task. Use the portal UI to see start and finish time of the task.

Comment out the vpn_subnet line in the config. Repeat the above.
See that without the vpn configured the task takes about 3 minutes less.

frankcorneliusmartin:
Fix #681


bartvanb (approved):
lgtm

PR #734, opened 2023-06-23
Instead of 'online' and 'offline', show 'running' or 'not running' fo…

### LINKED ISSUE #733: [Change Request] `vnode list` confuses by using the term `online` and `offline` for a running container
**Description**
This implies that the node is online, but this is not necessary the case as it only means that the Docker container is running. So it would be better to change it to `running` and `stopped`.

bartvanb:
Fix #733


frankcorneliusmartin (approved):
LGTM

PR #735, opened 2023-06-26
Feature/start UI with vserver

### LINKED ISSUE #730: [Feature Request] Start UI locally with `vserver`
**Problem description**
The User Interface is a convenient way to interact with a server. However, the UI cannot be started in a convenient way but only with a `docker run` command that is found in the UI's README.

**Desired solution**
It would be convenient to start the UI also with:

``` bash
vserver start --with-ui
# OR
vserver start-ui
```

bartvanb:
Fix #730


vantage6/vantage6/cli/globals.py
```diff
@@ -35,3 +35,6 @@
 
 # Maximum time to start up RabbitMQ in seconds
 RABBIT_TIMEOUT = 300
+
+# Default port to start the UI on
+DEFAULT_UI_PORT = 4200
```
frankcorneliusmartin, line 40: Why 4200? I know it does not matter, but ill forget this. Could we make it 5001 of 5555?

frankcorneliusmartin (approved):
Minor comment, approved nonetheless

vantage6/vantage6/cli/globals.py
bartvanb: 4200 is just the default port for an Angular app, but indeed that is not obvious to the average vantage6 user. I changed it to 5001, so that is right next to the default v6 server port

PR #739, opened 2023-06-27
Feature/retrieve collab policies vnode new

### LINKED ISSUE #434: [Feature Request] `vnode new` should retrieve collaboration info from the server
**Problem description**
To configure the node with the command `vnode new` the user needs to fill out a lot of details which can be retrieved from the server. Sometimes it is not clear that the settings should be in sync with the settings at the server (e.g. encryption). We should consider these options as collaboration policies and retrieve them from the server. This could also help in the future as we want to have privacy policies per algorithm (e.g. min number of patients). 

**Desired solution**
Simplify the `vnode questionair`.

1. _server url and port_
2. _api-key_
3. --> retrieve collaboration policies/settings from the server
4. private key (only when the collaboration encryption policy is enabled)

bartvanb:
Fix #434

### DIFF
--- a/vantage6/vantage6/cli/configuration_wizard.py
+++ b/vantage6/vantage6/cli/configuration_wizard.py
@@ -4,6 +4,9 @@ import uuid
 from pathlib import Path
 
 from vantage6.common.globals import DATABASE_TYPES
+# TODO v4+ move NodeClient to other location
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Perfect (y) Thanks for the comment

PR #740, opened 2023-06-27
Update the version in the Dockerfiles where appropriate

bartvanb:
Partial fix of #454


frankcorneliusmartin (approved):
LGTM. I think we should add this to the build pipeline when a minor is released?

bartvanb:
> LGTM. I think we should add this to the build pipeline when a minor is released?

Nope, the function that was added is already included in the `set-version` function that is called by the build pipeline, so we already do this! See change in lines 212-214 in update-version.py

PR #741, opened 2023-06-27
Change/remove backwards compatibility2

bartvanb:
Partial solution to #454


bartvanb:
Wait for completing this until we merge the collaboration scope into this

PR #744, opened 2023-07-03
Feature/request validation marshmallow

### LINKED ISSUE #76: Endpoint arguments validation
The `reqparser` from flask is depreciated and they recommend using marshmellow to handle validation of input arguments:
https://marshmallow.readthedocs.io/en/stable/examples.html#quotes-api-flask-sqlalchemy

This could/should be applied to all `POST` and `PATCH` endpoints.

bartvanb:
Fix #76


vantage6-server/vantage6/server/resource/common/input_schema.py
frankcorneliusmartin, line 1: Return `-> None` are missing

vantage6-server/vantage6/server/resource/common/input_schema.py
```diff
@@ -0,0 +1,454 @@
...
+        ValidationError
+            If the organizations are not valid.
+        """
+        if not len(organizations):
+            raise ValidationError('At least one organization is required')
+        for organization in organizations:
+            if 'id' not in organization:
+                raise ValidationError(
+                    'Organization id is required for each organization')
+            if 'input' not in organization:
```
frankcorneliusmartin, line 373: I dont think this is per-se the case? What if you just want to send a container without any input?

vantage6-server/vantage6/server/resource/common/input_schema.py
frankcorneliusmartin: I think this looks really nice

vantage6-server/vantage6/server/resource/port.py
```diff
@@ -249,17 +251,22 @@ def post(self):
...
+        # validate request body
+        errors = port_input_schema.validate(data)
+        if errors:
+            return {'msg': 'Request body is incorrect', 'errors': errors}, \
+                HTTPStatus.BAD_REQUEST
 
         # The only entity that is allowed to algorithm ports is the node where
         # those algorithms are running.
-        run_id = data.get('run_id', '')
+        run_id = data.get('run_id')
```
frankcorneliusmartin, line 262: Wont this break the filter bellow? Or is is guaranteed to be there because of the schema validation? In this case I would like it to be `data['run_id']` as this make this assumption explicit

vantage6-server/vantage6/server/resource/recover.py
```diff
@@ -532,22 +556,20 @@ def post(self):
...
+
+        # validate request body
+        errors = reset_api_key_schema.validate(body)
+        if errors:
+            return {'msg': 'Request body is incorrect', 'errors': errors}, \
+                HTTPStatus.BAD_REQUEST
 
-        # find the node
-        node = db.Node.get(id)
+        id_ = body.get('id')
```
frankcorneliusmartin, line 567: I think `id` is garanteed, so use square brackets here also

vantage6-server/vantage6/server/resource/user.py
```diff
@@ -590,28 +575,23 @@ def patch(self, id):
                         "msg": "You cannot change the username of another user"
                     }, HTTPStatus.BAD_REQUEST
             user.username = data["username"]
-        if data["firstname"] is not None:
+        if data.get("firstname") is not None:
```
frankcorneliusmartin, line 578: I prefer `if not data.get("firstname")` over the `is` comparison. In this case it works, as `None` always refers to the same spot in memory..

frankcorneliusmartin (changes requested):
Minor comments, other than that it looks good! Nice enhancement of the API

vantage6-server/vantage6/server/resource/common/input_schema.py
bartvanb: :-)

vantage6-server/vantage6/server/resource/common/input_schema.py
```diff
@@ -0,0 +1,454 @@
...
+                                validate=Length(max=_MAX_LEN_STR_LONG))
+
+
+class Recover2FAInputSchema(_PasswordValidationSchema):
+    """ Schema for validating input for recovering 2FA. """
+    email = fields.Email()
+    username = fields.String(validate=Length(max=_MAX_LEN_NAME))
+
+    @validates_schema
+    def validate_email_or_username(self, data: dict, **kwargs):
```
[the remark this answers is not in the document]
bartvanb: You mean whether it is a valid email / username or if it is an email / username that exists in the database? The latter is always checked in the endpoints, and I think we should keep doing it there (e.g. non-existing user should give 404, not 400). 
If you mean the former, it is validated if something is an email by specifying `fields.Email()` two lines above this comment. 
Then, we could validate the username further by using the function `_validate_name()`, but I think in this particular case we have an endpoint to recover a password, and then you should put an existing username. I think we don't necessarily want to tell here that the username is invalid.

vantage6-server/vantage6/server/resource/common/input_schema.py
bartvanb: The 'input' always contains at least the `method` field that describes which algorithm to run, so there should always be an input field. But maybe we should change that, and give the method a separate field? Or do we envision a case where one task may call different methods for different organizations?
If we do change this, perhaps make a separate issue?

For now, I added a check if the `input` field contains a `method` field

vantage6-server/vantage6/server/resource/user.py
bartvanb: No, we actually changed it to this a few months ago for a good reason: if you patch a user and you want to give them `firstname=''`, with this check it would work. With the check you suggest, it wouldn't work because an empty string would evaluate to False

vantage6-server/vantage6/server/resource/common/input_schema.py
frankcorneliusmartin: You are assuming the wrapper is used, I don't think we should force it.

bartvanb: You're right indeed, good point!

frankcorneliusmartin (approved):

PR #746, opened 2023-07-04
Fix sphinx warnings

bartvanb:
Fix warnings in sphinx documentation


docs/use/server.rst
bartvanb, line 1: Note that this file is now in `docs/server/use.rst` - it was not imported and outdated but not previously deleted by accident

frankcorneliusmartin (approved):
lgtm

PR #747, opened 2023-07-04
The POST /task endpoint had some code of fields that are no longer us…

### LINKED ISSUE #742: [Change Request] Refactor task creation code server-side for master algorithm
Hi @bartvanb, following up on the conversation from several days ago about centralized aggregator where you asked me to create this issue.

By reading [the code here](https://github.com/vantage6/vantage6/blob/version/3.10.4/vantage6-server/vantage6/server/resource/task.py#L444-L464), it looked like it would be impossible for a user to run the master algorithm on a node that is not their organization's node. But on further inspection it seems like that 'master' attribute is [never sent on the POST at that level](https://github.com/vantage6/vantage6/blob/version/3.10.4/vantage6-client/vantage6/client/__init__.py#L507-L514). It is sent I believe, but [inside the serialized `organizations.input`](https://github.com/vantage6/vantage6/blob/version/3.10.4/vantage6-client/vantage6/client/__init__.py#L503), which might or might not be encrypted. So, I guess at the moment the server can't control on which node the master algorithm is allowed to run.

Perhaps it would be nice if node administrators had granular control over which part of an algorithm is allowed to run on their nodes (master, partial or both). But at the moment I don't believe this feature exists, and as you said, the expected behavior now is that users should be able to arbitrarily choose where the master runs (IIRC). So, maybe it's best if we remove [this check for 'master' on the server](https://github.com/vantage6/vantage6/blob/version/3.10.4/vantage6-server/vantage6/server/resource/task.py#L444-L462)? Provided I haven't missed anything, which is not unlikely :)

bartvanb:
…ed, this has been adapted

- Removed part of code that was outdated
- Updated docstrings

Fix #742

### DIFF
--- a/vantage6-server/vantage6/server/resource/common/swagger_templates.py
+++ b/vantage6-server/vantage6/server/resource/common/swagger_templates.py
@@ -19,10 +19,6 @@ swagger_template = {
                         "type": "string",
                         "description": "Human-readable task description"
                     },
-                    "input": {
-                        "type": "string",
-                        "description": "Task input"
-                    },
                     "name": {
                         "type": "string",
                         "description": "Human-readable task name"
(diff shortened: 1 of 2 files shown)


frankcorneliusmartin (approved):
I checked that this is indeed not used in the wrapper. LGTM

PR #751, opened 2023-07-07
Share node database labels and types with the central server to allow…

### LINKED ISSUE #750: [Feature Request] Share a node's database labels and types with central server
**Problem description**
At present, database labels are defined by node admins and not shared anywhere. If a user runs a task for multiple organizations, they must define a database label that should be present at all nodes, but it is impossible to verify that these databases exist at the nodes via vantage6. It would be helpful to have this information so that we could alert users that the database label they define doesn't exist.

Secondly, it would also be useful to share database types so users know which additional parameters they may need to define (e.g. if a node uses an excel database is should receive a worksheet name to read in).

This is also useful for the new researcher UI we will work on soon.

**Desired solution**
Share node database labels and types with server using system developed for #12.

bartvanb:
… for easier task creation

Fix #750

### DIFF
--- a/vantage6-node/vantage6/node/__init__.py
+++ b/vantage6-node/vantage6/node/__init__.py
@@ -1059,6 +1059,17 @@ class Node:
             config_to_share['allowed_orgs'] = \
                 policies.get('allowed_organizations')
 
+        # share node database labels and types
+        labels = []
+        types = {}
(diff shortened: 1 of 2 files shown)


frankcorneliusmartin (approved):
lgtm

PR #755, opened 2023-07-12
Bugfix/no pagination get rules was deleted

bartvanb:
This is adding back some of the code that was wrongly deleted  in a merge commit (https://github.com/vantage6/vantage6/commit/0edbb8fe748c317efef8ff0fccb96bdb66a4da9c)


bartvanb:
I just added an extra commit to fix a CLI unittest. I don't understand why that broke here (I think it should have broken before that part entered the 4.0 branch, but it needs to be fixed anyway.,

frankcorneliusmartin (approved):
LGTM

PR #763, opened 2023-07-18
Split algorithm tools from client - extra package

### LINKED ISSUE #662: [Change Request] Split vantage6-client into two packages
**Description**
The vantage6-client package currently contains both the client package as well as algorithm development tools. The latter depends on much more dependencies than the former, and it is inconvenient that a user that installs the python client also has to install dependencies for stuff like algorithm wrappers.

We should therefore consider separating them in different packages.

bartvanb:
Fix #662


bartvanb:
Ignore codacy, it's all just stuff that has been moved or is irrelevant

frankcorneliusmartin (approved):
LGTM

PR #765, opened 2023-07-19
Fix bug where a ServerConfiguration object instead of a dict was save…

bartvanb:
There was the following issue with new config files:

image

This was because the data saved to the yaml was not a dict but a Configuration Python object.
Probably this went wrong in deleting the environments.


frankcorneliusmartin (approved):
lgtm

PR #774, opened 2023-07-24
Fixed bug in patch/post node where base class function exists() was o…

bartvanb:
…verwritten by an exists() that required different parameters


frankcorneliusmartin (approved):
LGTM

PR #780, opened 2023-07-27
Removed no-longer-existing include_metadata option from python client…

### LINKED ISSUE #773: [Bug] `fields` arguments is not working as the `include_metadata=False` option is removed from the Python client

bartvanb:
Fix #773  - and apart from fixing `client.x.list(fields=[...])`, I also removed all the `include_metadata` arguments that were still in the Python client but were already removed for the server in v4


frankcorneliusmartin (approved):
LGTM

PR #781, opened 2023-07-27
Updated check when user A tries to change permissions of user B.

### LINKED ISSUE #443: [Feature Request] Hierarchy in scope for rule assignment
**Problem description**
If a user creates another user or modifies the permissions of another user, they are only allowed to assign rules to that user that they also have themselves. So currently, if you want to give a user permission to edit roles for their organization, you need to be allowed to do that yourselves. 
But let's say I have permission to edit roles globally, then it's weird I'm not allowed to assign someone permission to assign roles for their organization, just because I don't have the exact same rule.

**Desired solution**
Introduce a hierarchy in allowing rule assignment. 
If someone has collaboration scope access, they should be allowed to assign organization scope access.
If someone has global access, they should be allowed to assign collaboration or organization scope access.

**Additional context**
Implementing this might impact the visualization of rules in the UI. E.g. if you select a global scope somewhere, the matching organization scope should also be selected.

bartvanb:
Previously, user A would have to have the exact same permissions as user B. Now, if user A has permission to create users with scope collaboration, they will also be able to assign that permission with lower scope (i.e. organization) to user B.

Fix #443

### DIFF
--- a/vantage6-server/vantage6/server/permission.py
+++ b/vantage6-server/vantage6/server/permission.py
@@ -37,6 +37,11 @@ def print_operation(operation: Operation) -> str:
     -------
     str
         String representation of the operation
+
+    Raises
+    ------
+    ValueError
+        If the operation is not known
     """
     if operation.VIEW:
         return "view"
@@ -48,8 +53,10 @@ def print_operation(operation: Operation) -> str:
         return "delete"
     elif operation.SEND:
         return "send"
-    else:
+    elif operation.RECEIVE:
         return "receive"
+    else:
+        raise ValueError(f"Unknown operation {operation}")
 
 
 def print_scope(scope: Scope) -> str:
@@ -65,6 +72,11 @@ def print_scope(scope: Scope) -> str:
     -------
     str
         String representation of the scope
+
+    Raises
+    ------
+    ValueError
+        If the scope is not known
     """
     if scope.ORGANIZATION:
         return "organization"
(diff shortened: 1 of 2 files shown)


vantage6-server/vantage6/server/permission.py
```diff
@@ -21,6 +21,61 @@
...
+        return "view"
+    elif operation.EDIT:
+        return "edit"
+    elif operation.CREATE:
+        return "create"
+    elif operation.DELETE:
+        return "delete"
+    elif operation.SEND:
+        return "send"
+    else:
```
frankcorneliusmartin, line 51: I dont like this being implicit. I would `raise ValueError` on `else:`

vantage6-server/vantage6/server/permission.py
```diff
@@ -21,6 +21,61 @@
...
+    str
+        String representation of the scope
+    """
+    if scope.ORGANIZATION:
+        return "organization"
+    elif scope.COLLABORATION:
+        return "collaboration"
+    elif scope.GLOBAL:
+        return "global"
+    else:
```
frankcorneliusmartin, line 75: Same for this else.

frankcorneliusmartin (changes requested):

vantage6-server/vantage6/server/permission.py
bartvanb: Good point

vantage6-server/vantage6/server/permission.py
bartvanb: Done

PR #783, opened 2023-07-31
Feature/added UI permissions to user endpoint

bartvanb:
This PR adds a few additional details to the user endpoint so that the UI can be more efficient in determining permissions.

### DIFF
--- a/vantage6-server/vantage6/server/resource/user.py
+++ b/vantage6-server/vantage6/server/resource/user.py
@@ -109,7 +109,7 @@ def permissions(permissions: PermissionManager) -> None:
 # ------------------------------------------------------------------------------
 user_schema = UserSchema()
 user_input_schema = UserInputSchema()
-user_schema_with_ui_permissions = UserWithPermissionDetailsSchema()
+user_schema_with_permissions = UserWithPermissionDetailsSchema()
 
 class UserBase(ServicesResources):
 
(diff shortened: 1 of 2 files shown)


vantage6-server/vantage6/server/resource/user.py
```diff
@@ -495,6 +497,13 @@ def get(self, id):
                 type: integer
               description: User id
               required: true
+            - in: path
+              name: include_ui_permissions
```
frankcorneliusmartin, line 501: I don't like the name `include_ui_permissions` as this implies these permissions are for the UI. But these are actually backend permissions. I would change this to `include_permissions`.

vantage6-server/vantage6/server/resource/user.py
```diff
@@ -495,6 +497,13 @@ def get(self, id):
...
+              schema:
+                type: boolean
+              description: Whether or not to include extra permission info for
+                the user interface. You can only get these for yourself. By
+                default false.
```
frankcorneliusmartin, line 506: Why make this exclusive to your own user? I get that this is specifically written for the UI, but why limited it? If other users have sufficient permissions this should not be a problem?

frankcorneliusmartin (changes requested):

vantage6-server/vantage6/server/resource/user.py
bartvanb: Changed this

frankcorneliusmartin (approved):
LGTM

frankcorneliusmartin:
Tested using 

```python
client.request('user/1', method='get', params={'include_permissions': True})
client.request('user/1', method='get')
```

Output seems correct. Related issue: #832

PR #788, opened 2023-08-15
Give default roles with low permissions the extra permission to view …

bartvanb:
…all nodes from their collaboration.

Now that they don't have these permissions, they cannot know if the nodes in their collaboration are online when creating a task or so


frankcorneliusmartin (approved):
OK

PR #789, opened 2023-08-15
Change/config file location linux

### LINKED ISSUE #269: [Change Request] Storage location of server config files on Linux
**Description**
Server configuration files created through `vserver new` are currently stored in the folder `/xdg`. The directory is specified [here](https://docker-py.readthedocs.io/). It seems not to be good practice to store these files there. They should be probably stored in `/etc` instead. 
We should read up on what the best practices are and then apply it. Preferably, keep using the `appdirs` package to maintain clean cross-platform path construction.

bartvanb:
Fix #269

### DIFF
--- a/vantage6-common/vantage6/common/__init__.py
+++ b/vantage6-common/vantage6/common/__init__.py
@@ -291,6 +291,9 @@ def get_config_path(dirs: appdirs.AppDirs,
     """
     if system_folders:
         config_dir = dirs.site_config_dir
+        # the Appdirs package prefers to store the config in /etc/xdg, but
+        # we chose instead to put it in /etc/vantage6. We think this is more
+        # in accordance with the Unix File Hierarchy Standard for config files.
         if 'xdg' in config_dir:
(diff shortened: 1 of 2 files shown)


vantage6-common/vantage6/common/__init__.py
```diff
@@ -279,6 +272,32 @@ def check_config_writeable(system_folders: bool = False) -> bool:
...
+        Whether to get path to the system folders or the user folders.
+
+    Returns
+    -------
+    str
+        The path to the configuration directory.
+    """
+    if system_folders:
+        config_dir = dirs.site_config_dir
+        if 'xdg' in config_dir:
```
frankcorneliusmartin, line 294: Could you add a small comment here why we remove it and not use the default appdirs?

frankcorneliusmartin (approved):
Alternatively you could have wrapped `appdirs.AppDirs(APPNAME, "")`, that might have looked cleaner. But.. LGTM

PR #790, opened 2023-08-16
Raise errors when algorithm developers try to use the refresh_token()…

bartvanb:
… or authenticate() functions in the client, which are included in the BaseClient it inherits

Note that these were also in the old `ContainerClient`, and I just spotted they were missing in the new one.


frankcorneliusmartin (approved):
LGTM

PR #792, opened 2023-08-17
The example batch import file in the docs was empty - this is fixed h…

bartvanb:
…ere.

Something must have gone wrong in the migration to the new docs.
I got the file contents from the old docs and checked that the import still worked.


frankcorneliusmartin (approved):
LGTM

PR #795, opened 2023-08-18
Change/rabbitmq hostname check improved

### LINKED ISSUE #282: [Bug] Rabbitmq URI does not support hostnames and does require a very specific local hostname
**Describe the bug**

_Specific local hostname_
In the case you want to use a local rabbitmq container the _host_ part of the `rabbitmq_uri` needs to be in the following format: `vantage6-[SERVER_NAME]-rabbitmq`. If this name is incorrect this leads to a lot of errors when starting (and stopping) the server. 

_hostnames_
https://github.com/vantage6/vantage6/blob/01e55ea12be82ed9eddf3f7db11aa6d02d1b2bdb/vantage6/vantage6/cli/rabbitmq/__init__.py#L33

The assumption ip-address == not local address is not true when:
1. The rabbitmq service has a hostname
2. when using localhost

Another usefull thing would be that the CLI is more helpful when using a local rabbitmq instance (e.g. prefilling a valid URI). This would make testing easier and gives users a better hint how the URI should be formatted.

bartvanb:
Fix #282 

- Changed configuration structure RabbitMQ so that users can indicate to run it locally
- no longer requiring specific hostname for running locally but any valid docker container name will do
- updated the docs
- updated `vserver new`

### DIFF
--- a/vantage6/vantage6/cli/rabbitmq/queue_manager.py
+++ b/vantage6/vantage6/cli/rabbitmq/queue_manager.py
@@ -39,7 +39,7 @@ class RabbitMQManager:
     def __init__(self, ctx: ServerContext, network_mgr: NetworkManager,
                  image: str = None) -> None:
         self.ctx = ctx
-        self.queue_uri = self.ctx.config.get('rabbitmq_uri')
+        self.queue_uri = self.ctx.config.get('rabbitmq').get('uri')
         rabbit_splitted = split_rabbitmq_uri(self.queue_uri)
         self.rabbit_user = rabbit_splitted['user']
         self.rabbit_pass = rabbit_splitted['password']
@@ -89,7 +89,7 @@ class RabbitMQManager:
             ports=ports,
             detach=True,
             restart_policy={"Name": "always"},
-            hostname=f'{APPNAME}-{self.ctx.name}-rabbitmq',
+            hostname=self.host,
             labels={
                 f"{APPNAME}-type": "rabbitmq",
             },
--- a/vantage6/vantage6/cli/rabbitmq/queue_manager.py
+++ b/vantage6/vantage6/cli/rabbitmq/queue_manager.py
@@ -39,7 +39,7 @@ class RabbitMQManager:
     def __init__(self, ctx: ServerContext, network_mgr: NetworkManager,
                  image: str = None) -> None:
         self.ctx = ctx
-        self.queue_uri = self.ctx.config.get('rabbitmq').get('uri')
+        self.queue_uri = self.ctx.config['rabbitmq'].get('uri')
         rabbit_splitted = split_rabbitmq_uri(self.queue_uri)
         self.rabbit_user = rabbit_splitted['user']
         self.rabbit_pass = rabbit_splitted['password']
--- a/vantage6/vantage6/cli/server.py
+++ b/vantage6/vantage6/cli/server.py
@@ -44,7 +44,7 @@ from vantage6.cli.utils import (
     remove_file
 )
 from vantage6.cli.rabbitmq.queue_manager import RabbitMQManager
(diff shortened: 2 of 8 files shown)


vantage6/vantage6/cli/rabbitmq/queue_manager.py
```diff
@@ -39,7 +39,7 @@ class RabbitMQManager:
     def __init__(self, ctx: ServerContext, network_mgr: NetworkManager,
                  image: str = None) -> None:
         self.ctx = ctx
-        self.queue_uri = self.ctx.config.get('rabbitmq_uri')
+        self.queue_uri = self.ctx.config.get('rabbitmq').get('uri')
```
frankcorneliusmartin, line 42: I assume that this class is only invoked when `self.ctx.config.get('rabbitmq')` exists in the config. In that case, change it to:  `self.ctx.config['rabbitmq'].get('uri')` or when this is not the case and a `None` value is acceptable you could do: `self.ctx.config.get('rabbitmq', {}).get('uri')`

vantage6/vantage6/cli/server.py
```diff
@@ -162,6 +167,12 @@ def vserver_start(ctx: ServerContext, ip: str, port: int, image: str,
         port to listen on
     image : str
         Server Docker image to use
+    start_ui : bool
+        Start the graphical User Interface as well
+    ui_port : int
+        Port to listen on for the User Interface
+    start_rabbitmq : bool
+        Start RabbitMQ message broker as local container - use in development
```
frankcorneliusmartin, line 175: "only" missing, now it sounds mandatory

vantage6/vantage6/cli/server.py
```diff
@@ -812,18 +831,15 @@ def _start_rabbitmq(ctx: ServerContext, rabbitmq_image: str,
     network_mgr : NetworkManager
         Network manager object
     """
-    rabbit_uri = ctx.config.get('rabbitmq_uri')
+    rabbit_uri = ctx.config.get('rabbitmq').get('uri')
```
frankcorneliusmartin, line 834: Same here: `['rabbitmq']` or `('rabbitmq', {}).get(`

vantage6/vantage6/cli/server.py
```diff
@@ -862,7 +878,8 @@ def _stop_server_containers(client: DockerClient, container_name: str,
 
     # kill RabbitMQ if it exists and no other servers are using to it (i.e. it
     # is not in other docker networks with other containers)
-    rabbit_uri = ctx.config.get('rabbitmq_uri')
+    rabbit_uri = ctx.config.get('rabbitmq').get('uri') \
```
frankcorneliusmartin, line 881: And here also

frankcorneliusmartin (changes requested):
Could you take a look at the 'rabbitmq' key and make it more explicit if it is expected or optional? See specific comments

vantage6/vantage6/cli/server.py
bartvanb: added

vantage6/vantage6/cli/rabbitmq/queue_manager.py
bartvanb: You are right, so I changed it into the ['rabbitmq'] version

vantage6/vantage6/cli/server.py
bartvanb: Changed to first option as it is guaranteed to exist here

vantage6/vantage6/cli/server.py
bartvanb: Here I actually cannot guarantee the key exists, so went for the other variant that guarantees this doesn't crash

frankcorneliusmartin (approved):
LGTM now

PR #796, opened 2023-08-18
In CLI unittest, clean up the test private key file

bartvanb:
This was annoying me already for ages, thought I should do something about it :-)


frankcorneliusmartin (approved):
:trollface:

PR #802, opened 2023-08-22
On systems with older iptables (iptables-legacy), VPN configuration i…

bartvanb:
…s fixed by running iptables-legacy commands if regular fails

Caused by the changes in https://github.com/vantage6/vantage6/pull/725 - apparently we now only fixed it for systems with `iptables` and it no longer works for systems where the actual iptables rules are in `iptables-legacy`


frankcorneliusmartin (approved):
I think now the default is to use the outdated `iptables` command first right? In that case the fix should not be removed eventually, but the code should be updated to work with the new `iptables` command. If this is the case we should create an issue for this before merging this.

bartvanb:
No actually the new `iptables` is called first, then when that doesn't work, we try `iptables-legacy`. So the legacy command can probably be removed at some point

PR #805, opened 2023-08-23
Added code to be able to include both result and run with GET task

bartvanb:


vantage6-server/vantage6/server/model/task.py
```diff
@@ -66,6 +66,10 @@ class Task(Base):
     collaboration = relationship("Collaboration", back_populates="tasks")
     parent = relationship("Task", remote_side="Task.id", backref="children")
     runs = relationship("Run", back_populates="task")
+    # This second relationship is needed when constructing output JSON (
+    # including results with the task). It is marked 'view_only' to prevent
+    # write conflicts with the runs relationship.
+    results = relationship("Run", back_populates="task", viewonly=True)
```
frankcorneliusmartin, line 72: I fail to understand why you would need a second relationship for that? Why not use the `.runs` to retrieve them?

vantage6-server/vantage6/server/resource/__init__.py
```diff
@@ -70,7 +70,13 @@ def is_included(field) -> bool:
         bool
             True if the field is included, False otherwise
         """
-        return field in request.args.getlist('include')
+        # The logic below intends to find 'x' both in 'include=y&include=x' and
+        # 'include=x,y'.
```
frankcorneliusmartin, line 74: Thanks for the comment that it still support the old way =]

frankcorneliusmartin (approved):

vantage6-server/vantage6/server/model/task.py
bartvanb: In `output_schema.py`, where you define the output Marshmallow schema's, you do:
```
class TaskWithResultSchema(TaskSchema):
    """Returns the TaskSchema plus the correspoding results."""
    results = fields.Nested('ResultSchema', many=True)
```
And for including a nested schema there also has to be a relationship to base this on, which is the one defined above.

PR #806, opened 2023-08-28
Bugfix/vserver import users without permissions

### LINKED ISSUE #634: [Bug] Super role not assigned  any rules in `vserver import`
When running the `vserver import` cli-command to batch import configurations, even if a super-user is specified it is not assigned any rules. 

Fix: To update Super user role with appropriate rules. Look into root user when creating new server.

bartvanb:
Fix #634


bartvanb:
Just a thought: should we release this still for v3?

frankcorneliusmartin (approved):
Good find.

frankcorneliusmartin:
> Just a thought: should we release this still for v3?

Nah

Issue #777, opened 2023-07-25
[Feature Request] Generic preprocessing before running algorithm


bartvanb:
**Problem description**
The algorithm wrappers currently read the data from the database and feed it to the algorithm. Each algorithm is then responsible for implementing their own data processing, such as data transformation or filtering. It would be convenient if the wrapper have some simple, often-used preprocessing steps available that are generically useful for algorithms.

**Desired solution**
In creating tasks, there will be one or more fields looking something like
```
preprocessing={
  'transform': {
    'type': 'years-since'
    'input_column_name': 'birth_date',
    'output_column': 'age'
  },
  'filter': {
    'column_name': 'age',
    'min': 50
  },
}
```

In the example above, the intention is to calculate the age from the `birth_date` column and then filtering it to be at least 50.

harmbuisman:
To consider:
* need type casting between input from e.g. a UI (string?) and the type expected by the function arguments, should be handled automatically such that we ideally do not have to manage it on the level of the individual preprocessing functions

bartvanb:
Depends on #943 - it doesn't make sense implement preprocessing first as the implementation would change a lot after chaining is implemented.

bartvanb:
This is no longer relevant, as v5.0 will introduce sessions, where preprocessing will be done by dedicated algorithms and thus not be part of the wrapper

PR #807, opened 2023-08-30
First working version of preprocessing for algorithms

### LINKED ISSUE #804: [Change Request] Change POST task structure to allow for multiple databases
**Description**
The JSON structure of POSTing tasks of relevant fields in v3 is as follows:
```
{
    "name": "name", ...
    "organizations": [
        {
            "id": 2,
            "input": {
               ....,
               "query": "SELECT * from table"
            }
        }
    ],
    "databases": [
        "my_database_label",
        "my_second_db_label"
    ]
}
```

The problem with the above is that you can only define 1 query, but if you have two databases (and both are SQL), you need two queries. 

**Suggested solution**

Change to:
```
{
    "name": "name", ...
    "organizations": [
        {
            "id": 2,
            "input": {
                ...
		databases": [
		    {
		        "label": "some_label",
                        "query": "SELECT * FROM mytable;"
		     },
		    {
		        "label": "some_other_label",
                        "query": "SELECT * FROM my_other_table;"
		     },
		]
            } 
        }
    ]
}
```

The added advantages of doing it like that is that it will be possible to send different queries to different organizations (convenient for vertically partitioned data) and that it is extendable if we need more args such as `query` for one DB

It would even be possible to send multiple queries to the same database (!)

bartvanb:
Fix #804.
This is preliminary work towards #777 but not a final implementation

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
@@ -1,15 +1,14 @@
+import json
 import os
 import pandas as pd
 
 from functools import wraps
-from typing import Any
 
 from vantage6.algorithm.client import AlgorithmClient
 from vantage6.algorithm.tools.mock_client import MockAlgorithmClient
-from vantage6.algorithm.tools.wrap import load_input
 from vantage6.algorithm.tools.util import info, error, warn
 from vantage6.algorithm.tools.wrappers import select_wrapper
-
+from vantage6.algorithm.tools.preprocessing import preprocess_data
 
 def algorithm_client(func: callable) -> callable:
     """
@@ -114,7 +113,6 @@ def data(number_of_databases: int = 1) -> callable:
             # query to execute on the database
             input_file = os.environ["INPUT_FILE"]
             info(f"Reading input file {input_file}")
-            input_data = load_input(input_file)
 
             # read the labels that the user requested, which is a comma
             # separated list of labels.
@@ -133,7 +131,19 @@ def data(number_of_databases: int = 1) -> callable:
 
             for i in range(number_of_databases):
                 label = labels[i]
-                data_ = _get_data_from_label(label, input_data)
+                # read the data from the database
+                info("Reading data from database")
+                data_ = _get_data_from_label(label)
+
+                # do any data preprocessing here
+                info(f"Applying preprocessing for database '{label}'")
+                preprocess = json.loads(
+                    os.environ.get(f"{label.upper()}_PREPROCESSING")
+                )
+                if preprocess:
+                    data_ = preprocess_data(data_, preprocess)
+
+                # add the data to the arguments
                 args = (data_, *args)
 
             return func(*args, **kwargs)
@@ -141,7 +151,7 @@ def data(number_of_databases: int = 1) -> callable:
     return protection_decorator
 
 
-def _get_data_from_label(label: str, input_data: Any) -> pd.DataFrame:
+def _get_data_from_label(label: str) -> pd.DataFrame:
     """
     Load data from a database based on the label
 
@@ -149,8 +159,6 @@ def _get_data_from_label(label: str, input_data: Any) -> pd.DataFrame:
     ----------
     label : str
         Label of the database to load
-    input_data : Any
-        Input data from the input file
 
     Returns
     -------
@@ -174,5 +182,18 @@ def _get_data_from_label(label: str, input_data: Any) -> pd.DataFrame:
               f"label '{label}'. Please check the node configuration.")
         exit(1)
 
-    # Load the data from the database
-    return wrapper.load_data(database_uri, input_data)
+    if database_type == "excel":
+        # read sheet name from the environment variables. If not given, read
+        # the first sheet.
+        sheet_name = os.environ.get(f"{label.upper()}_SHEET_NAME")
+        return wrapper.load_data(database_uri, sheet_name=sheet_name)
+    elif database_type in ("sparql", "sql", "omop"):
+        query = os.environ.get(f"{label.upper()}_QUERY")
+        if not query:
+            error(f"Missing query for database with label '{label}'. Please "
+                  f"include it when creating the task.")
+            exit(1)
+        return wrapper.load_data(database_uri, query)
+    else:
+        # Load the data from the database without additional arguments
+        return wrapper.load_data(database_uri)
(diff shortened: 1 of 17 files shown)


docs/user/pyclient.rst
```diff
@@ -317,21 +317,24 @@ Here we assume that
...
 In this manual, we'll use the averaging algorithm from
 ``harbor2.vantage6.ai/demo/average``, so the second requirement is met.
-This container assumes a comma-separated (\*.csv) file as input, and will
-compute the average over one of the named columns. We'll assume the
-nodes in your collaboration have been configured to look at a
-comma-separated database, i.e. their config contains something like
+We'll assume the nodes in your collaboration have been configured to look as
+something like:
 
 ::
```
frankcorneliusmartin, line 323: `.. code:: yaml`

vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
```diff
@@ -128,6 +133,10 @@ def wait_for_results(self, task_id: int, *args, **kwargs) -> list:
         ----------
         task_id: int
             ID of the task for which the results should be obtained.
+        interval: float
+            Interval in seconds between checking for new results. This is
+            ignored in the mock client but included for compatibility with
```
frankcorneliusmartin, line 138: .. to match the signature of the AlgorithmClient.

vantage6-server/vantage6/server/resource/common/swagger_templates.py
```diff
@@ -49,18 +49,32 @@
...
                         }
-                    ]
+                    ],
+                    "databases": [
+                        {
+                            "label": "database-label",
+                            "query": "SELECT * FROM table",
+                            "preprocessing": [
+                                {
+                                    "type": "filter_range",
```
frankcorneliusmartin, line 68: I would consider adding `function` as key. In this way you are able to group logical preprocessing groups: `filter`, `impute`, `engineer`, `convert`, etc. (effectively it is not different that prefixing them with `filter_`, etc.)

vantage6-server/vantage6/server/resource/common/input_schema.py
```diff
@@ -371,6 +371,46 @@ def validate_organizations(self, organizations: list[dict]):
...
+                # TODO we may add further validation on the preprocessing
+                # parameters when that is completed
+                for prepro in database['preprocessing']:
+                    if not 'type' in prepro:
+                        raise ValidationError(
+                            f"Database preprocessing {prepro} is missing a "
+                            "'type'"
+                        )
+            allowed_keys = {'label', 'preprocessing', 'query', 'sheet_name'}
+            if database.keys() - allowed_keys:
```
frankcorneliusmartin, line 408: The error message is incorret in the case f `database.keys() < allowed_keys`. I would check that `database.keys()` is a subset of `allowed_keys`:  

```python
set(database.keys()).issubset(set(allowed_keys))
```

frankcorneliusmartin (changes requested):

vantage6-server/vantage6/server/resource/common/swagger_templates.py
bartvanb: I don't fully understand what you mean. The `filter_range` is actually a preprocessing function - I guess we should change `type` to `function`. I can see we may use some kind of grouping of the preprocessing functions in the frontend but I don't think the user should have to specify the group

bartvanb: Btw I first made it that the structure was something like `{"filter_range": { <dict of parameters for filtering range> }}` but changed it to this list structure because that allows you to use the same preprocessing function multiple times.

docs/user/pyclient.rst
bartvanb: done

vantage6-server/vantage6/server/resource/common/input_schema.py
bartvanb: I think the former *is* actually giving the correct response, but I do prefer the set solution, as it is a bit more verbose and therefore less confusing. I now set to:
```
if not set(database.keys()).issubset(set(allowed_keys)):
```
Which basically is just the same as your comment except for the `not`.

frankcorneliusmartin (approved):
LGTM

vantage6-server/vantage6/server/resource/common/swagger_templates.py
frankcorneliusmartin: I was considering:

```json
{
    "type": "filter",
    "function": "range"
}
```
which is effectively the same as what you proposed. However that you can create library classes:

```python
class Filter:
    def range(...):
    ...
```

bartvanb: Good suggestion, let's have a look in v4.1 to incorporate this

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -133,24 +135,36 @@ def decorator(mock_data: list[pd.DataFrame] = None, *args,
...
                 label = labels[i]
-                data_ = _get_data_from_label(label, input_data)
+                # read the data from the database
+                info("Reading data from database")
+                data_ = _get_data_from_label(label)
+
+                # do any data preprocessing here
+                info(f"Applying preprocessing for database '{label}'")
+                preprocess = json.loads(
+                    os.environ.get(f"{label.upper()}_PREPROCESSING")
```
frankcorneliusmartin, line 145: This results in an error when there is no JSON to load?

frankcorneliusmartin:
```python
2023-09-14 09:47:48 - task (4)       - INFO     - info > wrapper for v6-average-py
info > Reading input file /mnt/data/task-000000006/input
info > Dispatching ...
info > Module 'v6-average-py' imported!
info > Reading input file /mnt/data/task-000000006/input
info > Reading data from database
info > Using '/mnt/data/default.csv' with label 'default' as database
info > Applying preprocessing for database 'default'
error > Error encountered while calling partial_average: the JSON object must be str, bytes or bytearray, not NoneType
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/vantage6/algorithm/tools/wrap.py", line 108, in _run_algorithm_method
    result = method(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/vantage6/algorithm/tools/decorators.py", line 144, in decorator
    preprocess = json.loads(
  File "/usr/local/lib/python3.10/json/__init__.py", line 339, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not NoneType
error > None
```

PR #822, opened 2023-09-12
The field init_user_id was not filled out for tasks created via vserv…

bartvanb:
…er import

This may lead to issues further downstream because this field is normally filled


bartvanb:
Superseded by #828

PR #844, opened 2023-09-14
Bugfix/paginate organizations

### LINKED ISSUE #392: [Feature Request] Extend API to enable pagination in UI
**Problem description**
We would like to use pagination in the UI to make everything go a lot faster, but we're missing some things on the API to make that change. This should be included in v4.

Features that we definitely need:
- some type of `include` argument where you include complete resources. E.g. `/node?include=collaboration,organization` should yield the node with inside the full json of the organization and collaboration. (Note that this could lead to difficult authorization checks)
- A `count` option to get the total number of resources without needing to get them all fully

We should consider if these are all necessary features, or if there are more

bartvanb:
Fix #392

### DIFF
--- a/vantage6-server/vantage6/server/resource/common/pagination.py
+++ b/vantage6-server/vantage6/server/resource/common/pagination.py
@@ -186,13 +186,8 @@ class Pagination:
 
         # check if pagination is desired, else return all records
         if paginate:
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Ah I see, I could have added `page`. Then it would have worked

PR #847, opened 2023-09-14
Use same minor version as server when starting the node

### LINKED ISSUE #700: [Change Request] Node should adapt the version it pulls

bartvanb:
Fix #700

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -232,6 +232,8 @@ jobs:
     needs: [version, infrastructure-images]
     env:
       version: ${{ needs.version.outputs.version }}
+      major: ${{ needs.version.outputs.major }}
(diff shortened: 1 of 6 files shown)


frankcorneliusmartin (approved):
OK

PR #881, opened 2023-09-29
Fixed broken data loading in mockClient

### LINKED ISSUE #872: [Bug] Passing dataframes to the AlgorithmMockClient broke in 4.0
**Describe the bug**
Previously it was possible to pass dataframes directly to the mockclient, in 4.0 this broke.

The problem is in this piece of code:
https://github.com/vantage6/vantage6/blob/257c68bb952dff137f72a3b4605674ca39dced2f/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py#L99C1-L106C1

It can be fixed by the following conditional check, which I added to the preprocessing branch as well (needs import of pandas as pd):

```

    database_handle = dataset.get("database")
    if isinstance(database_handle, pd.DataFrame):
        df = database_handle
    else:
        df = load_data(
            database_uri=database_handle,
            db_type=dataset.get("db_type"),
            query=dataset.get("query"),
            sheet_name=dataset.get("sheet_name"),
        )

```
Please fix it, because without it we cannot upgrade to 4.0 as our unit tests depend on this to work.

bartvanb:
- Handle pd.Dataframe properly
- correct 'type' to 'db_type'

Fix #872

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
@@ -1,9 +1,10 @@
 import json
 import logging
-
 from typing import Any
 from importlib import import_module
 from copy import deepcopy
+
+import pandas as pd
 
 from vantage6.algorithm.tools.wrappers import load_data
 from vantage6.algorithm.tools.util import info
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
LGTM

PR #882, opened 2023-09-29
Propagate signal that data decorator is used in client decorator

### LINKED ISSUE #874: [Bug] decorator problem with MockAlgorithmClient method
**Describe the bug**
When using the MockAlgorithmClient, the order of decorators of the called method leads to an error:
@algorithm_client
@data()

KeyError: 'INPUT_FILE' as it tries to get INPUT_FILE from os.environment.

The following order however works:
@data()
@algorithm_client

**To Reproduce**
you can reproduce it by setting the wrong order in the mock_page.__init__ file in the preprocessing branch and run make test

bartvanb:
Also removed superfluous code block in data decorator

Fix #874


frankcorneliusmartin (changes requested):
use `functools.wraps` + alias

frankcorneliusmartin (approved):
LGTM

PR #885, opened 2023-10-03
Feature/extemding wrapper for omop

frankcorneliusmartin:
Could we merge this part now? That way we can start using the OMOP base image for creating algorithms. Dont delete the branch.


docker/algorithm-base.Dockerfile
```diff
@@ -1,5 +1,6 @@
-FROM python:3.10-slim-buster
+FROM python:3.11-slim-bullseye
```
bartvanb, line 1: If we do this, we need to modify the docs that algorithms should be written in python 3.11.
Also, what's the diff between bullseye and buster?

docker/algorithm-base.Dockerfile
```diff
@@ -1,5 +1,6 @@
...
 
+# install dependencies for the ohdsi tools part of the wrapper
```
bartvanb, line 3: This comment seems not to belong?

docker/algorithm-omop-base.Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN Rscript -e "install.packages('DatabaseConnector')"
+RUN Rscript -e "install.packages('drat')"
+RUN Rscript -e "drat::addRepo('OHDSI'); install.packages('FeatureExtraction')"
+RUN Rscript -e "install.packages('remotes')"
+RUN Rscript -e "remotes::install_github('ohdsi/CirceR')"
+RUN Rscript -e "remotes::install_github('OHDSI/CohortDiagnostics')"
+RUN Rscript -e "remotes::install_github('OHDSI/CohortGenerator')"
+
+
+ENV LD_LIBRARY_PATH=/usr/lib/jvm/java-17-openjdk-amd64/lib/server/
```
bartvanb, line 33: Maybe you could add a comment what this does

bartvanb (changes requested):

docker/algorithm-base.Dockerfile
frankcorneliusmartin: That is the version of the Debian distribution. I needed this for some R / Java related dependency. This OS version change should not affect algorithms. 

I will check if it also works with python:3.10, else I will update the docs.

docker/algorithm-omop-base.Dockerfile
frankcorneliusmartin: I added a comment and a fixme, as this is not such a great solution. However i also do not know how with deal with this properly.

docker/algorithm-base.Dockerfile
frankcorneliusmartin: Yes you ar right.

docker/algorithm-base.Dockerfile
frankcorneliusmartin: I ended up reverting to `3.10` just to avoid compatibility issues

bartvanb: Clear!

bartvanb (approved):

PR #898, opened 2023-10-16
Specify algorithmClient boolean request parameters only if they are true

bartvanb:
Flask proxy server handles any other value as true including 'False'


frankcorneliusmartin (approved):
LGTM

PR #902, opened 2023-10-17
Feature/extemding wrapper for omop

frankcorneliusmartin:


frankcorneliusmartin:
## Codacy issues
1. Outdated linter https://stackoverflow.com/questions/72732058/match-case-invalid-syntax-with-spyder
2. And some `pip install ...` without pinned version, this is intentional for now

docker/algorithm-ohdsi-base.Dockerfile
```diff
@@ -1,4 +1,11 @@
-FROM harbor2.vantage6.ai/infrastructure/algorithm-base:latest
+ARG TAG=latest
+ARG BASE=4.0
+# ARG OHDSI_VERSION=0.3.2
```
bartvanb, line 3: should we remove this?

bartvanb: and line 7 as well? Or why not?

docs/node/node_config.yaml
```diff
@@ -17,9 +17,9 @@ vpn_subnet: 10.76.0.0/16
 algorithm_device_requests:
   gpu: false
 
-# add additional environment variables to the algorithm containers.
-# this could be usefull for passwords or other things that algorithms
-# need to know about the node it is running on
+# Add additional environment variables to the algorithm containers. In case
+# you want to supply database specific environment you should use `env` key in
+# the `database` settings.
```
bartvanb, line 22: Can we keep the original example here? Also, the sentence you added is not entirely clear to me: which `env` key and which database settings? Can you be more specific?

bartvanb: Right, I see now that these are added below. In that case, maybe we restore this to original as I think that's clearer: when the example for OMOP is already there, it's obvious not to use this for OMOP

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -84,7 +117,7 @@ def data(number_of_databases: int = 1) -> callable:
     """
     Decorator that adds algorithm data to a function
 
-    By adding @data to a function, one or several pandas dataframes will be
+    By adding `@data` to a function, one or several pandas dataframes will be
```
bartvanb, line 120: This should actually be `@data()`

bartvanb: or with `N` inside the parentheses

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -168,6 +199,215 @@ def decorator(*args, mock_data: list[pd.DataFrame] = None,
...
+            Wrap the function with the database connection
+            """
+            labels = _get_user_database_labels()
+            if len(labels) != 1:
+                error(f"User provided {len(labels)} databases, but algorithm "
+                      f"requires 1 database connection. Exiting...")
+                exit(1)
+
+            match type_:
+                case "OMOP":
```
bartvanb, line 234: maybe case insentivity would be nice

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -168,6 +199,215 @@ def decorator(*args, mock_data: list[pd.DataFrame] = None,
...
+    -------
+    callable
+        OHDSI Database Connection object
+    """
+
+    # check that the OHDSI package is available in this container
+    if not OHDSI_AVAILABLE:
+        error("OHDSI/DatabaseConnector is not available.")
+        error("Did you use the correct algorithm base image to "
+              "build this algorithm?")
```
bartvanb, line 385: Would be nice to specify which one they should use

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -168,6 +199,215 @@ def decorator(*args, mock_data: list[pd.DataFrame] = None,
...
+        exit(1)
+
+    # check that the required environment variables are set
+    if "DBMS" not in os.environ:
+        error("Environment variable 'DBMS' is not set. Exiting...")
+        exit(1)
+    if "USER" not in os.environ or "PASSWORD" not in os.environ:
```
bartvanb, line 392: I prefer splitting these into separate `if`s because then the user doesn't have to guess what they are missing

vantage6-node/vantage6/node/docker/task_manager.py
```diff
@@ -518,6 +518,12 @@ def _setup_environment_vars(self, algorithm_env: dict,
             type_var_name = f'{label.upper()}_DATABASE_TYPE'
             environment_variables[type_var_name] = db['type']
 
+            # Add optional database parameter settings, these can be used by
+            # the algorithm (wrapper).
+            if 'env' in db:
+                for key in db['env']:
+                    environment_variables[key.upper()] = db['env'][key]
```
bartvanb, line 525: This is a bit tricky - should we worry that people can overwrite other envvars in this way? We could work with an include list instead of doing everything (but I see the chances are small)

bartvanb (changes requested):
looks good, small changes

docker/algorithm-ohdsi-base.Dockerfile
frankcorneliusmartin: Added a note why I kept that

vantage6-node/vantage6/node/docker/task_manager.py
bartvanb: Yes sounds good! I do feel like we may be able to think of something more specific than `SPECIAL`, e.g. something like `DB_PARAM`. But then again, `SPECIAL` also indicates such cases nicely, although then we should be sure to make clear in the comments/user feedback what this is and why they cannot use it. So let's use `SPECIAL` with good comments?

frankcorneliusmartin: Done, i prefixed them with `DB_PARAM` 👍

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -209,12 +210,31 @@ def database_connection(type_: str) -> callable:
...
     ----------
-    type_ : str
-        Type of database to connect to. Currently only "OMOP" is supported.
+    types : list[str]
+        List of types of databases to connect to. Currently only "OMOP" is
+        supported.
+    include_metadata : bool
+        Whether to include metadata in the function arguments. This metadata
+        contains the database name, CDM schema, and results schema. Default is
+        True.
```
bartvanb, line 219: I'm not too familiar with OMOP, but usually I wouldn't expect to need this metadata in an algorithm for a SQL DB. Just to check: do you think the majority of algorithms will need this?

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -225,48 +245,37 @@ def decorator(*args, **kwargs) -> callable:
...
-            return run_metadata
-        case "omop":
-            return omop_metadata
-        case _:
-            info(f"Unknown metadata type: {type_}. Exiting.")
-            exit(1)
-
-
-def run_metadata(func: callable, *args, **kwargs) -> callable:
+def metadata(func: callable, *args, **kwargs) -> callable:
```
bartvanb, line 278: I could be wrong, but I wouldn't expect the arguments here - then calling it would be `@metadata(func, ...)` right?

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
```diff
@@ -387,15 +391,18 @@ def _create_omop_database_connection(label: str) -> callable:
...
+        _check_environment_var_exists_or_exit(f'{label_}_DB_PARAM_{var}')
 
     info("Reading OHDSI environment variables")
     dbms = os.environ["DB_PARAM_DBMS"]
-    uri = os.environ[f"{label.upper()}_DATABASE_URI"]
-    user = os.environ["DB_PARAM_USER"]
-    password = os.environ["DB_PARAM_PASSWORD"]
+    uri = os.environ[f"{label_}_DATABASE_URI"]
+    user = os.environ[f"{label_}_DB_PARAM_USER"]
+    password = os.environ[f"{label_}_DB_PARAM_PASSWORD"]
```
bartvanb, line 405: I would have a slight preference to change `{label}_DB_PARAM` into `DB_PARAM_{label}` because then all the 'special' envvars start with the same prefix. If you disagree, feel free to leave as is

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
bartvanb: Also, as this decorator has now changed, shouldn't that lead to a docs update?

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
frankcorneliusmartin: In the OMOP case, yes. As many of the OHDSI function that we use in the algorithm need certain settings that are in this metadata. That is why I set it to default to `True`. In light of adding different connections this might get inconvenient? Let me know.

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
frankcorneliusmartin: > I could be wrong, but I wouldn't expect the arguments here - then calling it would be `@metadata(func, ...)` right?

Yes, that was a remainder of the change. I removed them. 



> Also, as this decorator has now changed, shouldn't that lead to a docs update?

I tried to get away with the fact that is was documented in the `function-docs`. I added a section in `Implementing the algorithm functions`

vantage6-algorithm-tools/vantage6/algorithm/tools/decorators.py
frankcorneliusmartin: I did this do be consistent with the `*_DATABASE_URI` now all settings are per database grouped. I think this is also nice since these are database dependent. So I am going to leave it like this.

bartvanb (approved):

PR #905, opened 2023-10-17
Clarified and fixed errors in docstrings of algorithmClient functions…

### LINKED ISSUE #900: Unexpected behaviour algorithm_client.vpn.get_parent_address
algorithm_client.vpn.get_parent_address() returns a list rather than a string which I expected from the name or a dict which the docstring suggests, e.g. I now get this output:
 [{'port': 49167, 'label': '= message_queue', 'ip': 'XXX, 'organization_id': 3, 'task_id': 118, 'parent_id': None},
  {'port': 49168, 'label': '= algorithm_communication', 'ip': 'XXX', 'organization_id': 3, 'task_id': 118, 'parent_id': None},
  {'port': 49169, 'label': '= other_label', 'ip': 'XXX', 'organization_id': 3, 'task_id': 118, 'parent_id': None}]

I would have expected a string with the address in it. We have multiple ports exposed in the docker config file with labels that represent what we see here.

The docstring says the return type is a dict.

Suggestions:
* if this is a bug, then fix
* update the docstring to reflect the expected output, including the keys of the dict part that is returned
* rename the function to something like get_parent_address_info to better reflect that one gets more than just the ip address
* the label, e.g. '= algorithm_communication' seems a bug, I think it should be 'algorithm_communication'

The same applies to the other get_address functions.

bartvanb:
Fix #900

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
@@ -474,9 +474,6 @@ class AlgorithmClient(ClientBase):
 
             return results['addresses']
 
-        # FIXME consider renaming this and similar functions in v5+: users
-        # reported they expected a '{ip}:{port}' string, not a dict - we should
-        # evaluate what is most effective though complete
         def get_parent_address(self) -> dict:
             """
             Get the IP address and port number of the parent of the current
--- a/vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
@@ -437,8 +437,9 @@ class AlgorithmClient(ClientBase):
             Returns
             -------
             list[dict] | dict
-                List of dictionaries containing the IP address and port number,
-                and other information to identify the containers. If obtaining
+                List of dictionaries containing the IP address and port
+                number(s) of the different communication channels, and
+                additional data to identify them. If obtaining
                 the VPN addresses from the server fails, a dictionary with a
                 'message' key is returned instead.
             """
@@ -470,6 +471,9 @@ class AlgorithmClient(ClientBase):
 
             return results['addresses']
 
+        # FIXME consider renaming this and similar functions in v5+: users
+        # reported they expected a '{ip}:{port}' string, not a dict - we should
+        # evaluate what is most effective though complete
         def get_parent_address(self) -> dict:
             """
             Get the IP address and port number of the parent of the current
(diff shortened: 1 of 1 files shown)


vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -437,8 +437,9 @@ def get_addresses(
             Returns
             -------
             list[dict] | dict
-                List of dictionaries containing the IP address and port number,
-                and other information to identify the containers. If obtaining
+                List of dictionaries containing the IP address and port
+                number(s) of the different communication channels, and
+                additional data to identify them. If obtaining
```
frankcorneliusmartin, line 442: Maybe its good to mention that one algorithm could have multiple addresses and that therefor the additional data is required. Now it still seems a bit vague to me. (Maybe put this in the body and not in the return type)

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -470,18 +471,22 @@ def get_addresses(
 
             return results['addresses']
 
+        # FIXME consider renaming this and similar functions in v5+: users
+        # reported they expected a '{ip}:{port}' string, not a dict - we should
+        # evaluate what is most effective though complete
```
frankcorneliusmartin, line 476: In that case this function should have an adduitional argument `label` as well.

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
frankcorneliusmartin: Same goes for the very similar docstrings

frankcorneliusmartin (changes requested):

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: Yes indeed. This was to address a question by Harm that he expected a string to be returned for a function with this name. I can see that the dictionary is bloated now, maybe we can get away with just `ip, port, label` and maybe `node_id` or `organization_id`. But I guess that would also be v5

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: Changed them again

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: removed this now: the new docstrings should be clear enough on what to expect - and Harm indicated this was the main ponit.

frankcorneliusmartin (approved):
LGTM

PR #908, opened 2023-10-19
Feature/column name endpoint

### LINKED ISSUE #778: [Feature Request] Collect database column names
**Problem description**
It is very useful to share the column names of the databases when creating new tasks via the UI. This is e.g. required for the planned preprocessing tools (#777) but can also be convenient to check between nodes when algorithms require a column name as input.

**Desired solution**

Share column names. Several options:
- Doing this via a small basic algorithm that loads the data from the wrapper. Plus side is it's easy to implement, downside that it takes more time
- Doing this by calling wrapper code. Easy, but can still take time for larger queries
- Doing this over socket
- Creating an API endpoint

bartvanb:
Fix #778

### DIFF
--- a/vantage6-common/vantage6/common/globals.py
+++ b/vantage6-common/vantage6/common/globals.py
@@ -33,3 +33,6 @@ PING_INTERVAL_SECONDS = 60
 # start trying to refresh the JWT token of the node 10 minutes before it
 # expires.
 NODE_CLIENT_REFRESH_BEFORE_EXPIRES_SECONDS = 600
+
+# The basics image can be used (mainly by the UI) to collect column names
+BASICS_IMAGE = 'harbor2.vantage6.ai/algorithms/basics'
--- a/vantage6-common/vantage6/common/globals.py
+++ b/vantage6-common/vantage6/common/globals.py
@@ -35,4 +35,4 @@ PING_INTERVAL_SECONDS = 60
 NODE_CLIENT_REFRESH_BEFORE_EXPIRES_SECONDS = 600
 
 # The basics image can be used (mainly by the UI) to collect column names
-BASICS_IMAGE = 'harbor2.vantage6.ai/algorithms/basics'
+BASIC_PROCESSING_IMAGE = 'harbor2.vantage6.ai/algorithms/basics'
--- a/vantage6-node/vantage6/node/docker/docker_manager.py
+++ b/vantage6-node/vantage6/node/docker/docker_manager.py
@@ -19,7 +19,7 @@ from pathlib import Path
 from vantage6.common import logger_name
 from vantage6.common import get_database_config
 from vantage6.common.docker.addons import get_container, running_in_docker
-from vantage6.common.globals import APPNAME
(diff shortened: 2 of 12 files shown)


vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -654,3 +663,33 @@ def kill_tasks(self, org_id: int,
...
+        Returns
+        -------
+        list[str]
+            List of column names
+        """
+        db = self.databases.get(label)
+        if not db:
+            self.log.error("Database with label %s not found", label)
+            return []
+        if not db['is_file']:
```
frankcorneliusmartin, line 687: sqlite databases are also file-based, not that we use them a lot. But in that case this will not work. Consider checking the type instead.

vantage6-server/vantage6/server/resource/ui/column.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        data = request.get_json()
+
+        # validate request body
+        errors = column_name_input_schema.validate(data)
+        if errors:
+            return {'msg': 'Request body is incorrect', 'errors': errors}, \
+                HTTPStatus.BAD_REQUEST
+
+        collaboration_id = data.pop('collaboration_id')
```
frankcorneliusmartin, line 118: Why use `pop`? Now you also modify `data` as side-effect?

vantage6-common/vantage6/common/globals.py
```diff
@@ -33,3 +33,6 @@
 # start trying to refresh the JWT token of the node 10 minutes before it
 # expires.
 NODE_CLIENT_REFRESH_BEFORE_EXPIRES_SECONDS = 600
+
+# The basics image can be used (mainly by the UI) to collect column names
+BASICS_IMAGE = 'harbor2.vantage6.ai/algorithms/basics'
```
frankcorneliusmartin, line 38: `BASICS_` is a bit too vague for me. Could we do `BASIC_PROCESSING_IMAGE`, or something in that line?

frankcorneliusmartin (approved):
Only minor changes needed

vantage6-common/vantage6/common/globals.py
bartvanb: Done

vantage6-server/vantage6/server/resource/ui/column.py
bartvanb: Well spotted! At some point, when I was considering a socket event, I changed it to pop so that I could send the rest of the args as data via the socket event (collab id was superfluous there as that would have been the socket room)

bartvanb: So changed it

vantage6-node/vantage6/node/docker/docker_manager.py
bartvanb: I would like to keep this check in case a CSV database points to a non-existing file. But I have also added another check now to ensure that this is only executed for supported data types (i.e. csv and sparql)

PR #927, opened 2023-11-02
Set READTHEDOCS environment variable for the 'make devdocs' target

bartvanb:


Makefile
```diff
@@ -203,5 +203,9 @@ publish:
 test:
 	coverage run --source=vantage6 --omit="utest.py","*.html","*.htm","*.txt","*.yml","*.yaml" utest.py
 
+# the READTHEDOCS envvar is set for this target to circumvent a monkey patch
+# that would get stuck indefinitely when running the sphinx-autobuild package.
+# Note that the value of the envvar does not matter, just that it is set.
+devdocs: export READTHEDOCS = Yes
```
frankcorneliusmartin, line 209: This does not seem legit. It seems you are redefining `devdocs`. Also does this work on windows?

frankcorneliusmartin (changes requested):

Makefile
bartvanb: As discussed, this works for me (although it looks weird). Consensus is that Frank should try if it works for him locally, then we merge it

frankcorneliusmartin (approved):
LGTM

PR #929, opened 2023-11-02
Update section on running a node or server locally with --mount-src o…

bartvanb:
…r --image

Also changed the title so that it is not confusing with the unit tests, and moved the section


frankcorneliusmartin (approved):
LGTM

PR #932, opened 2023-11-07
Error out in client if using invalid characters for database labels

lsago:
Algorithm images are usually built on top of the algorithm wrapper [0]. This is using python:3.10-slim-bullseye as base image at the moment, which as a Debian based image has dash as its `/bin/sh`. Docker will use that shell when using the CMD instruction in "shell form" [1] in the Dockerfile, like in the average algorithm example [2].

At the moment, environment variables are used to pass to the algorithm container the details of the database to use. This means that the database label will become an environment variable name.

Dash only allows valid identifiers: "a letter or underscore followed by zero or more letters, underscores, and digits" [3].

[0]: https://github.com/vantage6/vantage6/blob/version/4.0.3/vantage6-algorithm-tools/vantage6/algorithm/tools/wrap.py
[1]: https://docs.docker.com/engine/reference/builder/
[2]: https://github.com/IKNL/v6-average-py/blob/bb34f43/Dockerfile#L20
[3]: https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/parser.c?h=v0.5.12#n1483

### DIFF
--- a/vantage6-client/tests/test_client.py
+++ b/vantage6-client/tests/test_client.py
@@ -38,6 +38,43 @@ class TestClient(TestCase):
 
         assert results == [{'result': {'some_key': 'some_value'}}]
 
+    def test_parse_arg_databases(self):
+        dbs_in = [{"label": "dblabel"}]
+        dbs_out = UserClient.Task._parse_arg_databases(dbs_in)
+        self.assertEqual(dbs_in, dbs_out)
+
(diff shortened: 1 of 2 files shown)


bartvanb (approved):
Good work!

bartvanb:
Added to `release/4.2` branch so that we release it then

PR #948, opened 2023-12-08
Default roles were not visible to users without global scope to view …

bartvanb:
…roles on endpoint GET /role/<id>.

Now, they are visible to anyone with at least organization scope, same as is the case for the endpoint GET /role


frankcorneliusmartin (approved):
OK

Issue #951, opened 2023-12-19
[Bug] Log is not persisted for server


mellesies:
**Describe the bug**
The server log file is not persisted in the location pointed towards by `vserver files`.

**To Reproduce**
1. Start a vantage6 server
2. Run `vserver files`
3. The output lists a location for the server log
4. The file is not in the specified location (or anywhere else, as far as I can tell)

**Expected behavior**
Logs should be persisted and retrievable.

**Screenshots**
<img width="1179" alt="image" src="https://github.com/vantage6/vantage6/assets/9558262/5687046c-dbfa-4ce0-b370-b05a2dea04ad">

**Details**
 - OS: Ubuntu 22.04
 - Version 
   * vantage6==4.1.2
   * vantage6-client==4.1.2
   * vantage6-common==4.1.2
- Docker deamon version: 24.0.7

PR #953, opened 2023-12-19
Server logging didn't work anymore as server log file was created in …

bartvanb:
…different place in the container

Also 2 other improvements:
- ensure the appContext init is run when starting the server
- No longer mount all node logs into container for a single node

This supersedes #952 and instead of #946 it merges the changes into main - branch was created by cherrypicking the commit from the latter PR


bartvanb:
this fixed #951

PR #958, opened 2023-12-20
Change hardcoded email address to the one set in the configuration fi…

### LINKED ISSUE #956: [Bug] Support emails are always sent from hardcoded email address
**Describe the bug**
Emails are always sent from support at vantage6.

**To Reproduce**
1. Set up SMTP in server config
2. Set up support `support_email` config key
3. Request a password reset
4. Email comes from [harcoded email address](https://github.com/vantage6/vantage6/blob/2502eb6dcb60895aa69f3b48d007d21e9edb0586/vantage6-server/vantage6/server/resource/recover.py#L256).

**Expected behavior**

Either the support_email address, or the sender emails address, or something else. Whatever makes the most sense.

**Details**
 - OS: Linux
 - Vantage6 version 4.1.2

bartvanb:
…le (if provided, else default)

Fix #956

### DIFF
--- a/vantage6-server/vantage6/server/resource/recover.py
+++ b/vantage6-server/vantage6/server/resource/recover.py
@@ -246,9 +246,8 @@ class RecoverPassword(ServicesResources):
             {"id": str(user.id)}, expires_delta=expires
         )
 
-        email_info = self.config.get("smtp", {})
-        email_sender = email_info.get("username",
-                                      DEFAULT_SUPPORT_EMAIL_ADDRESS)
+        email_sender = smtp_settings.get("username",
+                                         DEFAULT_SUPPORT_EMAIL_ADDRESS)
         support_email = self.config.get("support_email", email_sender)
 
         self.mail.send_email(
--- a/vantage6-server/vantage6/server/resource/recover.py
+++ b/vantage6-server/vantage6/server/resource/recover.py
@@ -253,7 +253,7 @@ class RecoverPassword(ServicesResources):
 
         self.mail.send_email(
             f"Password reset {APPNAME}",
-            sender="support@vantage6.ai",
+            sender=email_sender,
             recipients=[user.email],
             text_body=render_template(
                 "mail/reset_token.txt", token=reset_token,


lsago:
Thanks for the quick fix!

I was thinking if `email_sender` is what made the most sense, in case the user is setting up SMTP with a service like mailchimp, mailtrap.io, etc., where the 'username' might not necessarily be an email address or at least not one they might want vantage6 to use / reveal in outgoing emails.

bartvanb:
`email_sender` in the end is what is defined under `username` in your server configuration file in vantage6. I guess this was put as `username` in the config file because that's what Flask-mail also calls it (see [here](https://github.com/vantage6/vantage6/blob/main/vantage6-server/vantage6/server/__init__.py#L229)). 

I do get your confusion and I would be fine with renaming the option in the configuration file. To keep all v4 versions working together, we cannot simply rename it now, so then we should either wait till v5 or support both. 
Feel free to open an issue. I don't think this is at the top of our priorities so I doubt we will implement this soon. If you want to support it in v4, feel free to open a PR.

lsago:
Yeah, I think that's what I understood. The (potential) issue that I see is that for some of those services (Amazon SES, mailtrap, etc), the "user" you use to authenticate against their SMTP server may not be an email address. Or it might, but not necessarily one the server administrator wants to use / reveal in outgoing emails.

As in, this would be what they'd had to set up in the v6 server config yaml, for example:
```yaml
smtp:
  port: 587
  server: mail.service.example.org
  username: dd3317beef401234567f
  password: 7755cafeacafeb114
```

bartvanb:
Yes I see what you mean, but I think that the username/password is only used for the authentication and does not appear in the email body. So if that would log in to the email account for `my-account@abc.org`, that would be the email address that people receive an email from, right?
Have you tried your account to get an email and if so what does it look like? 

PS I hope that's not your real password ;-)

lsago:
It does not appear in the email body, but it will appear in the email headers. So, the user will see the email coming `From` the `username` that the server admin put in the config file. Which for some of those services (Amazon SES, mailtrap, etc) it might look like the example I put above (`dd3317beef401234567f`). But will test your code on this branch just to make sure :-)

[Not a real user/password, no. Random string, with extra `cafe` and `beef` on top :-)]

lsago:
Forgot to reply to this:

> So if that would log in to the email account for my-account@abc.org, that would be the email address that people receive an email from, right?

Not necessarily. I think. One thing is authenticating to the SMTP server, and once you are logged in, a different thing is the "from" email address you choose to use when sending an email. For instance, before the change of hardcoded `sender=` email, the email sent for a password reset appeared to come from that hardcoded email address (`support@vantage6`..), even though the SMTP server that I used is not associated to the vantage6.ai domain in any way.

I think it's likely this would get flagged as spam by most receiving email servers (I could imagine something like.. "Hey, this email says it's `From` jane@example.org but the email actually comes from a server that is not authorized / associated to the the domain `example.org`", not sure about details here though).

bartvanb:
Right, gotcha. The email headers may indeed be a problem there. Note that we are using openxchange email server hosted on eurodns and we don't have this issue. So that could be a workaround.

Nevertheless I recognize this can be improved. Can you create an issue for this? It would be helpful if you can show that your email headers are not good, causing emails to go to spam. We are also very much open to contributions on such an issue.

Meanwhile, let's still merge this PR since it solves a different (though related) issue

frankcorneliusmartin (approved):

Issue #961, opened 2023-12-20
[Feature Request] Options to add trusted CA certificate to node


lsago:
**Problem description**

A vantage6 node running on a host behind a transparent HTTPS proxy fails to connect to the vantage6 server because the node does not trust the necessary CA certificate set up for this transparent proxy.

**Desired solution**

A way for a node administrator to add a CA certificate for the node to trust, or the set of certificates to trust (?).

**Additional context**

In Python, [`requests` uses a separate rootstore](https://requests.readthedocs.io/en/latest/user/advanced.html#ca-certificates) from the system's (e.g. `/etc/ssl`). But I believe [environment variable `REQUESTS_CA_BUNDLE`](https://requests.readthedocs.io/en/latest/user/advanced.html#ssl-cert-verification) can be used to specify a different rootstore. Of course, this CA bundle / rootstore should exist in the container's file system, perhaps by means of volume mapping. 

Note too, that even when running `v6 node new` on the host (no docker), this will fail due to the configuration's wizard attempt to fetch data from the server. But in this case, the user can manually set the `REQUESTS_CA_BUNDLE` environment variable beforehand.

**Note**

Perhaps after we provide users with an option to run a node only using docker-compose, this feature will be implicit.

PR #963, opened 2024-01-05
Add node_extra_mounts & node_extra_env config options

lsago:
node_extra_mounts: allows the node administrator to mount on the node additional volumes.

node_extra_env: allows the node administrator to set additional environment variables.

Combined, these options can be used to address #961 (https://github.com/vantage6/vantage6/issues/961)


docs/node/node_config.yaml
```diff
@@ -28,6 +28,18 @@ algorithm_env:
   # the value 'Alice' inside the algorithm container
   player: Alice
 
+# Add additional environment variables to the node container. This can be useful
+# if you need to modify the configuration of certain python libraries that the
+# node uses. For example, if you want to use a custom CA bundle for the requests
+# library you can specify it here.
+node_extra_env:
+  - REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
```
frankcorneliusmartin, line 36: Could you change this to key: value pairs? This way its also consistent with the `algorithm_env` setting. 

```yaml
node_extra_env:
    REQUESTS_CA_BUNDLE: /etc/ssl/certs/ca-certificates.crt
```

docs/node/node_config.yaml
frankcorneliusmartin, line 1: I am not super happy with the `node_` prefix as this is the _node_ configuration file. Although I see that you did that to be explicit so that this is not confused with the `algorithm_env`. @bartvanb what do you think?

frankcorneliusmartin (changes requested):

docs/node/node_config.yaml
bartvanb: I think it is a good prefix considering what you say about `algorithm_env`. I think it's good to be clear and don't see a good alternative for `node_` so I'm happy leaving it as is

docs/node/node_config.yaml
lsago: Yup, this is better IMO too. Done.

bartvanb (commented):
@lsago I think it would be nice to check also which environment variables people put into their node configuration file. Specifically that they do not overwrite any of the default env vars that are defined in the node (which I think would be `PRIVATE_KEY`, `PROXY_SERVER_HOST`, `PROXY_SERVER_PORT` and `<some-label>_DATABASE_URI`)

Also would be nice to only allow identifiers to be env var keys

lsago:
> @lsago I think it would be nice to check also which environment variables people put into their node configuration file. Specifically that they do not overwrite any of the default env vars that are defined in the node (which I think would be `PRIVATE_KEY`, `PROXY_SERVER_HOST`, `PROXY_SERVER_PORT` and `<some-label>_DATABASE_URI`)

Code was checking for this but only throwing a warning, I will make it error out. It's checking against `env` dict which should include the vars you mention and any other that might come up. If I'm not mistaken?

> 
> Also would be nice to only allow identifiers to be env var keys

Will do. I think libraries expecting non-valid-identifier characters in env var names should be extremely rare (`dash` won't even accept them).

I'm being more permissive here because contrary to client controlled input, this (`node_extra_env`) is controlled solely by the node admin.

bartvanb:
> Code was checking for this but only throwing a warning, I will make it error out. It's checking against env dict which should include the vars you mention and any other that might come up. If I'm not mistaken?
Sorry I missed that!

> Will do. I think libraries expecting non-valid-identifier characters in env var names should be extremely rare (dash won't even accept them).
> I'm being more permissive here because contrary to client controlled input, this (node_extra_env) is controlled solely by the node admin.
Yeah I think so too. I'm also fine with not checking identifiers but at very least there should be no equals signs in env vars.

lsago:
> > Code was checking for this but only throwing a warning, I will make it error out. It's checking against env dict which should include the vars you mention and any other that might come up. If I'm not mistaken?
> 
> Sorry I missed that!
> 

No prob! Have made error out just in case – warning was easy to miss anyway.

> > Will do. I think libraries expecting non-valid-identifier characters in env var names should be extremely rare (dash won't even accept them).
> > I'm being more permissive here because contrary to client controlled input, this (node_extra_env) is controlled solely by the node admin.
> 
> Yeah I think so too. I'm also fine with not checking identifiers but at very least there should be no equals signs in env vars.

Done

bartvanb (approved):

bartvanb (commented):
Just noticed the unit tests are not successful, can you have a look there.

Concerning your remaining two points: I agree that would be better - on the whole the entire `start` command should be refactored as it is way too complex for a single function. Feel free to address them here or create an issue for refactoring (but that one I'd make a bit broader to also include a few other improvements)

lsago:
Have pushed a commit fixing test, commit message should hopefully explain.

I agree `start` command could use a refactor. I propose we address it in a different PR however, for the sake of releasing this feature sooner rather than later :).
I'm also keeping in mind that once we were to go full-dockerized for the node, we might want to re-write these cli tools to work with docker compose under the hood perhaps. .. still intermittently working on this every now and again :/

bartvanb (approved):

bartvanb:
> I agree start command could use a refactor. I propose we address it in a different PR however, for the sake of releasing this feature sooner rather than later :).
I'm also keeping in mind that once we were to go full-dockerized for the node, we might want to re-write these cli tools to work with docker compose under the hood perhaps. .. still intermittently working on this every now and again :/

A good reason to wait with refactoring `v6 node start` indeed.

Since we already released 4.2, I just updated the base branch to 4.3. However, I was wondering if you really need this soon - in that case we can merge to `main` and include it in 4.2.2 since it's only a small change. 
Also, with the new branch there is now a conflict, probably because the black code style is not yet applied on your feature branch. Let me know if I can help there.

lsago:
I would vote for merging this into main for 4.2.2 if possible, as I know of a couple of users that are waiting for this option.
In that case, I don't see potential merge conflicts with main at the moment.
What do you think?

bartvanb:
Merged. As soon as the other 3 small PRs that are currently open are merged, we will release 4.2.2

PR #970, opened 2024-01-10
Corrected wrong configuration documentation

frankcorneliusmartin:


docs/node/node_config.yaml
```diff
@@ -139,22 +140,22 @@ ssh-tunnels:
...
+      # connect.
+      bind:
+        ip: 0.0.0.0
+        port: 8000
+
+      # The port and ip on the remote machine. If the data source runs
+      # on this machine, the ip most likely is 127.0.0.1.
+      dest:
+        ip: 127.0.0.1
+        port: 8000
```
bartvanb, line 158: Now this is indented outside the SSH settings. That feels odd to me - are you sure this is correct?

frankcorneliusmartin: Yes this is correct. It is about the `tunnel` and not about the `ssh` connection.

bartvanb (approved):

PR #987, opened 2024-01-18
Fix unit tests to post tasks - they failed after introducing new inpu…

bartvanb:
Fix unit tests to post tasks - they failed after introducing new input check in post task


frankcorneliusmartin (approved):
LGTM

PR #1009, opened 2024-01-25
Node logs were not persisted due to error in mount location

### LINKED ISSUE #993: [Bug] Logfile only contains `INFO` (or higher?) logs
**Describe the bug**
It seems that I only get `INFO` logging in the logfile even when specified that the log level should be DEBUG in the config file.

```yaml
logging:
  backup_count: 5
  datefmt: '%Y-%m-%d %H:%M:%S'
  file: testing420_node_3.log
  format: '%(asctime)s - %(name)-14s - %(levelname)-8s - %(message)s'
  level: DEBUG
  max_size: 1024
  use_console: true
```

I do see the debug logs when I `attach` the node.

**To Reproduce**
1. start node with debug logging
2. observe that only info or higher is stored in the logfile

**Expected behavior**
I expect the same logging to be present in the logfile as in the attached form.

**Details**
 - Version 4.2.1

bartvanb:
Fix #993 (at least I see the debug logs now again, but before this fix I actually saw nothing at all instead of just INFO and higher)

### DIFF
--- a/vantage6/vantage6/cli/node/start.py
+++ b/vantage6/vantage6/cli/node/start.py
@@ -197,7 +197,7 @@ def cli_node_start(
     # FIXME: should obtain mount points from DockerNodeContext
     mounts = [
         # (target, source)
-        (f"/mnt/log/{name}", str(ctx.log_dir)),
+        ("/mnt/log", str(ctx.log_dir)),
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Lets leave the issue open to check in the prod environment

PR #1013, opened 2024-01-29
Added server_url configuration option to server configuration file an…

bartvanb:
…d use it to communicate with algorithm store

### DIFF
--- a/vantage6-server/vantage6/server/resource/algorithm_store.py
+++ b/vantage6-server/vantage6/server/resource/algorithm_store.py
@@ -201,6 +201,27 @@ class AlgorithmStoreBase(ServicesResources):
             headers=headers,
         )
 
+    def _get_server_url(self, server_url_from_request: str | None) -> str:
+        """ "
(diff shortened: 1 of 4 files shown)


vantage6-server/vantage6/server/resource/algorithm_store.py
```diff
@@ -201,6 +201,27 @@ def _execute_algo_store_request(
...
+        server_url_from_request : str | None
+            Server url from the request data.
+
+        Returns
+        -------
+        str | None
+            The server url
+        """
+        return (
+            self.config["server_url"]
```
frankcorneliusmartin, line 220: Why not use `self.config.get("server_url", server_url_from_request)`

frankcorneliusmartin (changes requested):

vantage6-server/vantage6/server/resource/algorithm_store.py
bartvanb: that's better indeed

PR #1023, opened 2024-01-31
Bugfix/fixing header issues

frankcorneliusmartin:
I am not completely sure that we need the `server_url` to be forwarded, see TODO in the code


vantage6-server/vantage6/server/resource/algorithm_store.py
```diff
@@ -103,7 +103,9 @@ def _request_algo_store(
             response = self._execute_algo_store_request(
                 algo_store_url, server_url, endpoint, method, force
             )
-        except requests.exceptions.ConnectionError:
+        except Exception as e:
+            log.warning("Request to algorithm store failed")
+            log.exception(e)
```
bartvanb, line 108: This will lead to printing an exception every time if you are running both server and algorithm store on localhost. Maybe we should check here if it is localhost before printing the exceptions to prevent a lot of error logs that we know will happen

vantage6-server/vantage6/server/resource/algorithm_store.py
```diff
@@ -176,9 +186,10 @@ def _execute_algo_store_request(
         if force:
             param_dict["force"] = True
 
-        # add server_url header
-        headers = {k: v for k, v in request.headers.items()}
-        headers["server_url"] = server_url
+        # set headers
+        # TODO FM 31-01-204: I dont think this is used by any of the endpoints? So
+        # we might be able to remove this.
```
bartvanb, line 191: Yes it is, see https://github.com/vantage6/vantage6/blob/release/4.3/vantage6-algorithm-store/vantage6/algorithm/store/resource/__init__.py#L31

bartvanb (changes requested):
Few minor comments

vantage6-server/vantage6/server/resource/algorithm_store.py
frankcorneliusmartin: I adjusted the comment, thanks for the link.

vantage6-server/vantage6/server/resource/algorithm_store.py
```diff
@@ -103,7 +103,8 @@ def _request_algo_store(
             response = self._execute_algo_store_request(
                 algo_store_url, server_url, endpoint, method, force
             )
-        except requests.exceptions.ConnectionError:
+        except Exception as e:
+            # lets try again before logging the exception
```
bartvanb, line 107: I guess my previous comment was not really clear. I think we *should* log here but *only* if the request is not directed at localhost (or 127.0.0.1). Otherwise you would not get any logs in that case, and you still prevent to get unnecessary errors in the logs for the localhost case

bartvanb (commented):
One comment from last time was not entirely clear I think

vantage6-server/vantage6/server/resource/algorithm_store.py
bartvanb: Also note that codacy is complaining that `e` is an unused variable

bartvanb:
Closing this as it was easier to reimplement (done in #1035)

PR #1037, opened 2024-02-08
Change/refactor common server code

### LINKED ISSUE #979: [Change Request] Remove algorithm store dependency on `vantage6.server`
**Description**
The algorithm store has been set up quickly and imports several modules/functions/etc from `vantage6.server`. It should be clear that this server functionality is used by the algorithm store, so all functionality from the server that is used by the algorithm store should be put into a separate folder, e.g. `vantage6-common-api`

bartvanb:
This introduces a new package `vantage6-backend-common` to contain functionality that both server and algorithm store use

Fix #979


bartvanb:
codacy issues checked - we'll ignore those

frankcorneliusmartin (approved):

PR #1042, opened 2024-02-08
Bugfix/startup node from relative path config

### LINKED ISSUE #870: [Change Request] CLI `--config` flag requires full path to YAML file
**Description**
If you provide a custom location for a config file, e.g. `vserver start -c myconfig.yaml`, the YAML file requires a full path. It would be nicer if it only requires a correct path.
This probably should also be changed for `vnode` commands.

Right now there is an error, as reported by Severin Elvatun in Discord:
image

bartvanb:
Fix #870

### DIFF
--- a/vantage6/vantage6/cli/context/node.py
+++ b/vantage6/vantage6/cli/context/node.py
@@ -61,7 +61,7 @@ class NodeContext(AppContext):
             Node context object
         """
         return super().from_external_config_file(
-            os.path.abspath(path), InstanceType.NODE, system_folders
+            Path(path).resolve(), InstanceType.NODE, system_folders
         )
 
     @classmethod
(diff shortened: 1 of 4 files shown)


vantage6/vantage6/cli/context/node.py
```diff
@@ -61,7 +61,7 @@ def from_external_config_file(
             Node context object
         """
         return super().from_external_config_file(
-            path, InstanceType.NODE, system_folders
+            os.path.abspath(path), InstanceType.NODE, system_folders
```
frankcorneliusmartin, line 64: I prefer the new object oriented way of doing this. 
`Path(path).resolve()`

Note that this method raises a `FileNotFoundError` when the file does not exist (`os.path.abspath` returns an empty string).

frankcorneliusmartin (approved):

Issue #724, opened 2023-06-20
[Bug] VPN setup fails on Ubuntu 22


bartvanb:
**Describe the bug**
On **Ubuntu 22**, setting up host network rules for VPN fails:

```
Exception: Command 'sh -c "iptables -I DOCKER-USER 1 -d 10.76.0.0/16 -i br-62d8457ae01e -j ACCEPT; 
iptables -I DOCKER-USER 1 -s 10.76.0.0/16 -o br-62d8457ae01e -j ACCEPT; "' in image 
'harbor2.vantage6.ai/infrastructure/vpn-configurator' returned non-zero exit status 1: b'iptables: No 
chain/target/match by that name.\niptables: No chain/target/match by that name.\n'
```

This happens when running commands from the `vpn-configurator` image to set up host networking rules. 

Probably something changed in the OS to not allow this. We know it works in Windows and Ubuntu 18 (not sure about Ubuntu 20).

**To Reproduce**

Start node on Ubuntu 22 with VPN. Error is shown in logs

**Expected behavior**
VPN works on Ubuntu 22

bartvanb:
It appears to be the case that docker images that are not based on ubuntu 22 cannot access the iptables in ubuntu 22 (see e.g. [this stackoverflow](https://stackoverflow.com/questions/74074082/manage-host-iptables-from-docker-container)).

A solution would therefore be to base the `vpn-configurator` image on ubuntu 22 instead of alpine. I tested this and it works with both Ubuntu 22 and Windows WSL2.

I also made a temp image `harbor2.vantage6.ai/infrastructure/vpn-configuration:ubuntu22` that should work on Ubuntu22, and will link a PR soon

frankcorneliusmartin:
1. Why does this exist? https://harbor2.vantage6.ai/harbor/projects/3/repositories/vpn-configuration (instead of vpn-configurator), should we  move/remove this.
2. I found in the release notes that with 3.10.4 we made some changes to the VPN stuff. Als in harbor i found a big difference in image size between 3.10.3 and 3.10.4 for the vpn-configurator https://docs.vantage6.ai/en/main/release_notes.html#id3
3. I replaced `harbor2.vantage6.ai/infrastructure/vpn-configurator:latest` with `harbor2.vantage6.ai/infrastructure/vpn-configurator:3.10.3`
4. I also think we should match versions of the support images with the node version image. This should help us avoid these problems and make is easier to debug.

bartvanb:
[...]

PR #1044, opened 2024-02-09
Change/vpn fixes ubuntu22

bartvanb:
(Partial?) fix for #724 

- building major+minor version support images
- no longer building support images twice each time
- use major+minor version support images that match node version by default

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -176,10 +176,6 @@ jobs:
         name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
-      - # vpn images
-        name: Build Docker support images
-        run: make support-image TAG=$version
-
       - # we only release a new base image if it is a major or minor release
         name: Build Docker base image
         if: ${{ env.patch == 0 }}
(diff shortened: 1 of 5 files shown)


frankcorneliusmartin (changes requested):
Ok. Good that these images are now also versioned the samen. I am however a little bit worried that this does not fix the VPN issues as this was a problem with the 'newer' VPN images?

From the issue https://github.com/vantage6/vantage6/issues/724: 

> I replaced harbor2.vantage6.ai/infrastructure/vpn-configurator:latest with harbor2.vantage6.ai/infrastructure/vpn-configurator:3.10.3 then it worked again.

bartvanb:
> Ok. Good that these images are now also versioned the samen. I am however a little bit worried that this does not fix the VPN issues as this was a problem with the 'newer' VPN images?

> From the issue https://github.com/vantage6/vantage6/issues/724:

> > I replaced harbor2.vantage6.ai/infrastructure/vpn-configurator:latest with harbor2.vantage6.ai/infrastructure/vpn-configurator:3.10.3 then it worked again.

Honestly I am not sure and think it would be best to test a release candidate if all issues are solved. There was definitely an explanation for older version to work: at some point, we changed from the old version of `iptables` to the new version. At that point, people with the older version had to use the old image. Then, we changed again to support both the old and the new in the commit I mentioned. But I'm not 100% sure this is the case, but best to find out by testing IMO

PR #1046, opened 2024-02-15
Added allowed_algorithms to configuration wizard and add warning mess…

bartvanb:
…ages if it is not defined

To prevent the configuration wizard from becoming to long, we may remove the VPN configuration option by default (as not many people use that)

### DIFF
--- a/vantage6/vantage6/cli/configuration_wizard.py
+++ b/vantage6/vantage6/cli/configuration_wizard.py
@@ -96,6 +96,32 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
             default="10.76.0.0/16",
         ).ask()
 
+    is_policies = q.confirm(
+        "Do you want to add limit the algorithms allowed to run on your node? This "
+        "should always be done for production scenarios.",
+        default=True,
+    ).ask()
+    if is_policies:
+        allowed_algorithms = []
+        print("Below you can add algorithms that are allowed to run on your node.")
+        print("You can use regular expressions to match multiple algorithms.")
+        print("Examples:")
+        # pylint: disable=W1401
+        # flake8: noqa: W605
+        print("^harbor2\.vantage6\.ai/demo/average    Allow the demo average algorithm")
+        print(
+            "^harbor2\.vantage6\.ai/algorithms/.*   Allow all algorithms from "
+            "harbor2.vantage6.ai/algorithms"
+        )
+        while True:
+            algo = q.text(message="Enter your algorithm expression:").ask()
+            allowed_algorithms.append(algo)
+            if not q.confirm(
+                "Do you want to add another algorithm?", default=True
+            ).ask():
+                break
+        config["policies"] = {"allowed_algorithms": allowed_algorithms}
+
     config["logging"] = {
         "level": res,
         "use_console": True,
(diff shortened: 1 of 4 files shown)


vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -96,6 +96,32 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
...
+        default=True,
+    ).ask()
+    if is_policies:
+        allowed_algorithms = []
+        info("Below you can add algorithms that are allowed to run on your node.")
+        info("You can use regular expressions to match multiple algorithms.")
+        info("Examples:")
+        # pylint: disable=W1401
+        # flake8: noqa: W605
+        info("^harbor2\.vantage6\.ai/demo/average    Allow the demo average algorithm")
```
frankcorneliusmartin, line 111: Maybe also add an example by pinning a specific hash

vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -96,6 +96,32 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
...
+        info(
+            "^harbor2\.vantage6\.ai/algorithms/.*   Allow all algorithms from "
+            "harbor2.vantage6.ai/algorithms"
+        )
+        while True:
+            algo = q.text(message="Enter your algorithm expression:").ask()
```
frankcorneliusmartin, line 117: I would make this `regex`

vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -96,6 +96,32 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
...
+            allowed_algorithms.append(algo)
+            if not q.confirm(
+                "Do you want to add another algorithm?", default=True
```
frankcorneliusmartin, line 120: Do you want to add another regex. As one regex does not neccasaraly mean one algorithm

frankcorneliusmartin (commented):
Can you also address this message, this log has been replaced by logging in the `_setup_policies`

frankcorneliusmartin (changes requested):
Minor changes

vantage6/vantage6/cli/configuration_wizard.py
bartvanb: Done

vantage6/vantage6/cli/configuration_wizard.py
bartvanb: Since I also added functionality to explicitly support normal strings now, I will leave it at `expression`

vantage6/vantage6/cli/configuration_wizard.py
bartvanb: See above, I changed this into `algorithm expression` now

vantage6/vantage6/cli/configuration_wizard.py
frankcorneliusmartin: Ah I see now. However I dont think your `_is_regex_pattern` method is correct?

```python
>>> pat = "henk*"
>>> re.match(pat, pat)
<re.Match object; span=(0, 4), match='henk'>
>>> not re.match(pat, pat)
False
>>> pat = "^henk*"
>>> not re.match(pat, pat)
True
```

A regular string is also a regex pattern so how would you be able to distinguish between them? You could assume that when it does not contain any special regex chars it is a regular string, but then you are still not sure. Maybe add the type to the config / questionary?

Or does it make sense to make these assumptions?

bartvanb: I prefer not to add it to the questionaire, so as not to complicate the questionaire further. Maybe it would be possible to add it as a configuration setting. 

For now, I have improved the implementation of checking regexes and also that of converting string to regex

frankcorneliusmartin (approved):

PR #1053, opened 2024-02-19
Remove option to change organization and collaboration id of existing…

bartvanb:
… node. Add option clear_ip that was missing from the python client


frankcorneliusmartin (approved):
Nice that looks a lot cleaner too =]

PR #1067, opened 2024-02-26
Change/remove unused omop wrapper

frankcorneliusmartin:


bartvanb (approved):
I just made a commit which removes a few other references to OMOP needing a query, with that all is approved

PR #1112, opened 2024-03-18
Change/integrate UI in main repo

bartvanb:


docs/server/optional.rst
```diff
@@ -60,6 +58,9 @@ own environment.
           - SERVER_URL=https://<url_to_my_server>
           - API_PATH=/api
 
+Alternatively, you can also run the UI locally with Angular. In that case, follow the
```
frankcorneliusmartin, line 61: This repo is going to be archived and likely going to be outdated. I would not mentiond it?

frankcorneliusmartin (approved):

docs/server/optional.rst
bartvanb: As discussed, the link has in fact been updated to the new address (but tricky to see with all those `vantage6/vantage6/vantage6` :-P)

Issue #1155, opened 2024-03-20
[Bug] Server you are trying to authenticate with is not whitelisted


frankcorneliusmartin:
**Describe the bug**
I keep getting the message: 
`{'msg': 'Server you are trying to authenticate with is not whitelisted'}`

Even though the user is added in the store config file as root user.

**Details**
 - Version 4.3.1

PR #1162, opened 2024-03-21
If server url ends with :443 (if https) or :80 (http) as passed to al…

bartvanb:
…gorithm store, check if it is whitelisted without the port

Also added some more data to the output of the endpoints retrieving users and whitelisted servers


bartvanb:
Fix #1155

Issue #1094, opened 2024-03-14
[Feature Request] Add ability to add hostname mappings to the node.


lsago:
**Problem description**

In Linux (no Docker Desktop), `host.docker.internal` to talk directly to the host from within a container does not work. If using docker compose one can use [`extra_hosts`](https://github.com/compose-spec/compose-spec/blob/master/spec.md#extra_hosts) to create that mapping ( `host.docker.internal:host-gateway`).

If I have an server running behind an HTTPS reverse-proxy, and I want to set up a node in the same LAN that connects to it using HTTPS (without leaving the LAN), it would be nice (for testing purposes) if I could remap the public domain for which it serves the certificate to the LAN (private) IP of the server.

**Desired solution**
Being able to set these mappings in the config yaml for the node, for example.

lsago:
Will push code soon.

PR #1167, opened 2024-03-23
Add node_extra_hosts option in node config

lsago:
This addresses https://github.com/vantage6/vantage6/issues/1094

I'm PR'in to main, but feel free to change that.

### DIFF
--- a/docs/node/node_config.yaml
+++ b/docs/node/node_config.yaml
@@ -42,7 +42,8 @@ node_extra_mounts:
 
 node_extra_hosts:
   # In Linux (no Docker Desktop) you can use this (special) mapping to access
-  # the host from the node
+  # the host from the node.
+  # See: https://docs.docker.com/reference/cli/docker/container/run/#add-host
   host.docker.internal: host-gateway
   # For testing purposes, it can also be used to map a public domain to a
   # private IP address, allowing you to avoid breaking TLS hostname verification
--- a/docs/node/node_config.yaml
+++ b/docs/node/node_config.yaml
@@ -40,6 +40,14 @@ node_extra_env:
 node_extra_mounts:
   - /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro
 
+node_extra_hosts:
+  # In Linux (no Docker Desktop) you can use this (special) mapping to access
+  # the host from the node
+  host.docker.internal: host-gateway
+  # For testing purposes, it can also be used to map a public domain to a
+  # private IP address, allowing you to avoid breaking TLS hostname verification
+  v6server.example.com: 192.168.1.10
+
 # specify custom Docker images to use for starting the different
 # components.
 # OPTIONAL
--- a/vantage6-node/vantage6/node/__init__.py
+++ b/vantage6-node/vantage6/node/__init__.py
@@ -498,9 +498,9 @@ class Node:
             self.log.warn(
                 f"You are trying to reach the server at {self.client.host}."
                 " As your node is running inside a Docker container, it cannot"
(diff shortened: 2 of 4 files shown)


vantage6-node/vantage6/node/__init__.py
```diff
@@ -498,9 +498,9 @@ def __print_connection_error_logs(self):
             self.log.warn(
                 f"You are trying to reach the server at {self.client.host}."
                 " As your node is running inside a Docker container, it cannot"
-                " reach localhost on your host system. Probably, you have to "
-                "change your serverl URL to http://host.docker.internal "
-                "(Windows/MacOS) or http://172.17.0.1 (Linux)."
+                " reach localhost on your host system. Probably, you have to"
+                " change your serverl URL to http://host.docker.internal (Windows/MacOS)"
+                ' or look into node config option "node_extra_hosts" (Linux).'
```
bartvanb, line 503: I'm wondering if `http://172.17.0.1` still works (or did it never work?). Should we provide that as alternative here?

bartvanb: Maybe we could specify that host.docker.internal is for when you're using docker desktop (if that's correct, I'm not sure)

lsago: I think `172.17.0.1` _may_ not always work. For example, if the subnet `172.17.0.0/24` is already being used for another purpose a newly started node container will get (my guess) a different subnet (e.g. 172.21.0.0/24). I believe it's also possible for a user to [configure their docker daemon to use a different subnet for new containers](https://docs.docker.com/network/drivers/bridge/#configure-the-default-bridge-network).

Docker allows for [extra host mappings](https://docs.docker.com/reference/cli/docker/container/run/#add-host), and it [treats the target 'host-gateway' string as a special string (moby)](https://github.com/moby/moby/blob/8b79278316b532d396048bc8c2fa015a85d53a53/opts/hosts.go#L30-L33) that will point to the host. This special treatment of 'host-gateway' is in moby, not Docker Desktop. But I believe Docker desktop probably passes this mapping by default, whereas plain docker does not.

In terms of the warning message, I decided to keep it short and just point the Linux user to the "node_extra_hosts". But if you'd rather we explain further, I can add some clarifications to it. At the very least, I'll fix that typo s/serverl/server/ now :p

bartvanb: Yeah I would like to mention that `172.17.0.1` is usually OK in the message. That will be a good solution for most node admins. 
Good that you spotted the typo, I can't believe I never saw it

docs/node/node_config.yaml
```diff
@@ -40,6 +40,14 @@ node_extra_env:
 node_extra_mounts:
   - /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro
 
+node_extra_hosts:
+  # In Linux (no Docker Desktop) you can use this (special) mapping to access
+  # the host from the node
+  host.docker.internal: host-gateway
```
bartvanb, line 46: I'm thinking it may be nice to provide a sensible example value here - that would be 172.17.0.1 right? If that's correct, maybe you could explain it a bit in the comment

lsago: I think the most common use case it to actually type in '`host-gateway`', as that's the special string the moby knows to point to the host. For an example with an IP, there's the one below. Or am I misinterpreting your request here?
I've added a [link](https://docs.docker.com/reference/cli/docker/container/run/#add-host) to explain a bit further.

vantage6-node/vantage6/node/__init__.py
lsago: Done.

PR #1186, opened 2024-04-05
Ensure that non-string environment variables are properly handled by …

bartvanb:
…the node


frankcorneliusmartin (approved):
Environment vars are always strings, so I approve

Issue #1188, opened 2024-04-07
[Change Request] Remove `pull_if_newer` mechanism


lsago:
**Description**
At the moment we have a function `pull_if_newer` that to avoid overwriting (well, technically retagging, I believe) an "older" (upstream) image. See [issue](https://github.com/vantage6/vantage6/issues/180) where *I think* this came from.

IMHO, I don't think the added complexity to solely deal with this today, when multiple registries are typically used for algorithms, is worth the continued effort.

These are some of the reasons, as a I see it:
* As [Frank pointed out](https://github.com/vantage6/vantage6/issues/180#issuecomment-1127623068), using a different tag locally is what is recommended by the moby community.
* IMHO, it would make the most sense to [only tag with the upstream registry name right before actually wanting to push](https://github.com/moby/moby/issues/10291#issuecomment-71675093).
* `pull_if_newer` system was (?) implemented at a time [when harbor seemed](https://github.com/vantage6/vantage6/issues/180#issuecomment-1127623073) to be only registry in use by vantage6. These days I think users (myself included) are using other registries for algorirhtms like Azure Container Registry or GitHub Container Registry.
* IMHO, vantage6 shouldn't be tied to a specific registry implementation like harbor – especially to parts of it that are outside of the [OCI distribution spec](https://github.com/opencontainers/distribution-spec).

FWIW, to deal with the [issue](https://github.com/vantage6/vantage6/issues/180) for algorithms I might be developing on locally, on the same machine where my node is running (testing), I think naming the image with a local tag seems very acceptable. Although most people I believe use the MockClient. Perhaps using 'localhost/name-of-image' could be an idea, or local.vantage6.ai .. or anything the user can rest assure will not actually exist out there.

For server/node/support images, this is where my bias shows :). I never have to deal with this, as if I ever make changes to the node/server code, I always just test it locally with `--mount-src`. And if I push it somewhere, it's not to the official registry.

Perhaps related to this issue, I'd argue that running something like `make base-image` should limit itself to building the image. Pushing to a production registry seems unexpected from just running that command. Perhaps we can find a series of `make` incantations that work for everybody – local tags, `make push`, .. or whatever we might come up with :)

Source code:
* https://github.com/vantage6/vantage6/pull/1175
* https://github.com/vantage6/vantage6/blob/2d7312df44379966ac7c610e0f9aa6559509f305/vantage6-common/vantage6/common/docker/addons.py#L66-L311

bartvanb:
[...]

Issue #1105, opened 2024-03-17
[Bug] Error messages and failure to pull latest image when using ghcr


lsago:
**Describe the bug**
When using GitHub's container registry to fetch images, node fails to guess which version of the registry API is using and errors are displayed in the logs.

**To Reproduce**
Send a task with an algorithm hosted on GHCR. I believe Azure Container Registry also has the same issue.

**Expected behavior**
No errors, and latest image (for the same tag) being pulled – consistent with behavior for algorithms hosted on dockerhub.

**Screenshots**
```
2024-03-17 03:27:23 - task (244)     - INFO     - Retrieving latest image: 'ghcr.io/lsago/v6-public-some-example-here:latest'
2024-03-17 03:27:24 - task (244)     - ERROR    - Could not determine version of the registry! ghcr.io
2024-03-17 03:27:24 - task (244)     - ERROR    - Is this a Harbor registry?
2024-03-17 03:27:24 - task (244)     - ERROR    - Or is the harbor server offline?
2024-03-17 03:27:24 - task (244)     - WARNING  - Only a local image has been found! ghcr.io/lsago/v6-public-some-example-here:latest
2024-03-17 03:27:24 - task (244)     - DEBUG    - Check if algorithm container is already running
```
**Details**
Vantage6 version 4.3.0

lsago:
* [Dockerhub implements](https://docs.docker.com/registry/) [1.0.1 OCI distribution specification](https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md)
* Code in question: https://github.com/vantage6/vantage6/blob/version/4.3.0/vantage6-common/vantage6/common/docker/addons.py#L166-L177 – I think`api/health` is not part of the spec. 
* I guess ideally we'd try to base behavior on (some version of) https://github.com/opencontainers/distribution-spec/blob/main/spec.md

PR #1189, opened 2024-04-08
Remove pull_if_newer()

lsago:
Rationale: https://github.com/vantage6/vantage6/issues/1188


bartvanb (approved):
I made a small change: utility function to log errors when images are not found. For me, that makes it good to go now

lsago:
> I made a small change: utility function to log errors when images are not found. For me, that makes it good to go now

Heh, I had that same change originally, but decided to just let docker throw the error. But indeed this is more clear. :)

bartvanb:
Fix #1188, #1105. Supersedes #1175

PR #1190, opened 2024-04-09
Feature/support list arguments in UI

### LINKED ISSUE #1119: [Feature Request] Support argument list types
**Problem description**
We currently support arguments of type str, float, int, json, column, organization and organizations.

If someone wants to have a list of something (e.g. list of strings), they should give the argument JSON type so that they can add a list, but it is ugly if someone has to type `[ 'a', 'b' ]` into a UI instead of simply `a` and `b` in different fields.

Also, there should be the option to select multiple columns

**Desired solution**
- [ ] support list of str / float / int by multiple input fields
- [ ] support list of columns by multi select dropdown
- [ ] Possibly requires changes in the algorithm store

**Additional context**
Note that we have already an implementation for multiple organizations

bartvanb:
Fix #1119

### DIFF
--- a/vantage6-algorithm-store/vantage6/algorithm/store/model/common/enums.py
+++ b/vantage6-algorithm-store/vantage6/algorithm/store/model/common/enums.py
@@ -19,14 +19,17 @@ class ArgumentType(str, enum.Enum):
     """Enum for argument types"""
 
     COLUMN = "column"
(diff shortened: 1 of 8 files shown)


bartvanb:
Not sure why codacy failed - the unused import it indicates is removed in the last commit. So I'm ignoring it

Issue #1191, opened 2024-04-09
[Change Request] Change `make` to not push to registry by default


lsago:
**Description**
When I run `make image` I expect it to just build an image – traditional use of make. It's slightly unnerving that it tries to push to the registry too :)

If the argument PUSH=true is passed (for example), then pushing would seem expected.

PR #1192, opened 2024-04-09
Modify `make` to not push to registry by default after build

lsago:
Since make will set command line variables to environment variables in subcommands, I use a slightly oddly named 'PUSH_REG' in case any command modifies behavior on 'PUSH' env var set.

This addresses https://github.com/vantage6/vantage6/issues/1191

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,25 @@
 TAG ?= cotopaxi
 REGISTRY ?= harbor2.vantage6.ai
 PLATFORMS ?= linux/arm64,linux/amd64
+# Example for local development
+# TAG ?= local
+# REGISTRY ?= localhost
+# PLATFORMS ?= linux/amd64
 
 # infrastructure base image version
 BASE ?= 4.0
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,10 +5,26 @@
...
+# Use `make PUSH_REG=true` to push images to registry after building
+PUSH_REG ?= false
+
+
+# We use a conditional (true on any non-empty string) later. To avoid
+# accidents, we don't use user-controlled PUSH_REG directly.
+# See: https://www.gnu.org/software/make/manual/html_node/Conditional-Functions.html
+_condition_push :=
+ifeq ($(PUSH_REG), true)
+	_condition_push := not_emtpy_so_true
```
bartvanb, line 25: empty

bartvanb (approved):
one small spelling thing which would be nice to fix but not really necessary :-)

Makefile
lsago: so true! :)

bartvanb:
I'll merge this into 4.4 as it is not too urgent

PR #1195, opened 2024-04-10
Feature/algorithm visualization

### LINKED ISSUE #1057: [Feature Request] Add structured definition of algorithm output to algorithm store
**Problem description**
In the UI, we would like to know how the output of algorithms will be structured so that the results can be visualized. It makes sense to add this output description to the algorithms when they are submitted to the algorithm store. Then, the UI should collect that information and display the results appropriately

**Desired solution**
- [x] Design structure with which algorithm output can be described
- [x] Add output structure description to algorithm store
- [x] Use output description in the UI to visualize the results. At the start, these may just be a few simple visualization options

bartvanb:
Fix #1057

### DIFF
--- a/vantage6-algorithm-store/vantage6/algorithm/store/model/common/ui_visualization_schemas.py
+++ b/vantage6-algorithm-store/vantage6/algorithm/store/model/common/ui_visualization_schemas.py
@@ -1,0 +1,69 @@
+"""
+This file describes how the JSON schema for different types of UI visualizations
+should be structured. The schema is used to validate input data.
+"""
+
+# import jsonschema
+
+from vantage6.algorithm.store.model.common.enums import VisualizationType
+
+
+# To visualize a table in the UI, the algorithm result should contain a table that
+# is structured as follows:
+# [
+#    { "column1": "value1", "column2": "value2", ... },
+#    { "column1": "value3", "column2": "value4", ... },
+# ]
+table_schema = {
+    "type": "object",
+    "properties": {
+        # indicate where in the results the table can be found. E.g. if the table is
+        # in results['data']['table'], the location should be ['data', 'table']. If it
+        # is not specified, the table is assumed to be the root of the results.
+        "location": {
+            "type": "array",
+            "items": {
+                "type": "string",
+            },
+        },
+        # columns of the table. Specify this if you only want to visualize a subset of
+        # the columns in the table. Example value: ["column1", "column2"]
+        "columns": {
+            "type": "array",
+            "items": {
+                "type": "string",
+            },
+        },
+        # if True, transpose the table.
+        "transpose": {
+            "type": "boolean",
+        },
+    },
+    # "required": ["columns"],
+    "additionalProperties": False,
+}
+
+
+def get_schema_for_visualization(visualization_type: str) -> dict:
+    """
+    Get the schema for a specific visualization type.
+
+    Parameters
+    ----------
+    visualization_type : str
+        Type of the visualization.
+
+    Returns
+    -------
+    dict
+        JSON validation schema for the visualization type.
+
+    Raises
+    ------
+    ValueError
+        If the visualization type is not supported.
+    """
+    if visualization_type == VisualizationType.TABLE.value:
+        return table_schema
+    else:
+        raise ValueError(f"Visualization type '{visualization_type}' is not supported.")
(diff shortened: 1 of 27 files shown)


vantage6-algorithm-store/vantage6/algorithm/store/model/ui_visualization.py
```diff
@@ -0,0 +1,33 @@
...
+    name : str
+        Name of the visualization
+    description: str
+        Description of the visualization
+    type_: str
+        Type of the visualization. Currently available: 'table'
+    schema : dict
+        Schema that describes the visualization, e.g. column names of a table
+    function_id : int
+        Id of the function that the visualization is linked to
```
frankcorneliusmartin, line 22: Also document relation `function` (we also do that in other models)

vantage6-algorithm-store/vantage6/algorithm/store/model/ui_visualization.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+
+    # fields
+    name = Column(String)
+    description = Column(String)
+    type_ = Column("type", String)
```
frankcorneliusmartin, line 28: Are you aware of the `Enum` type from SQLAlchemy? Then it will raise an exception in case its not within the Enum. 

```python
from sqlalchemy import Enum

type_ = Column(Enum(MyEnum), name='type')
```

bartvanb: Yes I am aware of it. However, I chose not to use that here, because once a table has been created in the database with such enum constraints, it is impossible to update the constraints in the database table. The only way to do that would be to remove the table and add it back again (and add the data back again) when constraints are updated.

Since I expect the different available types to be updated relatively frequently at this stage in the algo store, I didn't want to use the enums

vantage6-algorithm-store/vantage6/algorithm/store/model/common/ui_visualization_schemas.py
```diff
@@ -0,0 +1,64 @@
+"""
+This file describes how the JSON schema for different types of UI visualizations
+should be structured. The schema is used to validate input data.
+"""
+
+# import jsonschema
```
frankcorneliusmartin, line 6: unused import. Although it might be good for clarity to mention that the `table_schema` bellow is used for validating.

vantage6-algorithm-store/vantage6/algorithm/store/model/ui_visualization.py
bartvanb: We are actually doing this rather inconsistently - in some we do and in some we don't. I'll add the relationships everywhere in algo store.

vantage6-algorithm-store/vantage6/algorithm/store/resource/schema/input_schema.py
```diff
@@ -47,6 +52,7 @@ class FunctionInputSchema(_NameDescriptionSchema):
     type = fields.String(required=True)
```
frankcorneliusmartin, line 52: You can use the `data_key` argument to use a reservered Python keyword in the marshmallow scema's, which might be a nice thing todo, see the documentation of `Field` (also applies to `String` but the documentation of this argument is not repeated there): https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.Field

vantage6-algorithm-store/vantage6/algorithm/store/resource/schema/output_schema.py
```diff
@@ -105,6 +109,16 @@ class Meta:
...
 
 
+class UIVisualizationOutputSchema(HATEOASModelSchema):
+    """Marshmallow output schema to serialize the UIVisualization model"""
+
+    class Meta:
+        model = UIVisualization
+        exclude = ["type_"]
+
+    type = fields.String(attribute="type_")
```
frankcorneliusmartin, line 119: Same here with the `data_key`

frankcorneliusmartin (approved):

frankcorneliusmartin (approved):

vantage6-algorithm-store/vantage6/algorithm/store/resource/schema/input_schema.py
bartvanb: Nice addition indeed, changed that so that we don't have to use the python keyword

vantage6-algorithm-store/vantage6/algorithm/store/resource/schema/output_schema.py
bartvanb: Indeed, the attribute I was using was doing the opposite but this is nicer

frankcorneliusmartin:
The codacy error is due to an unused import in our `__init__.py`

PR #1197, opened 2024-04-10
Decoding all environment variables in algorithm tools wrapper, so tha…

### LINKED ISSUE #1056: [Feature Request] Decode the PROXY environment variables automatically
**Problem description**
The environment variables: 

- `HTTP_PROXY`
- `HTTPS_PROXY`
- `NO_PROXY`
- `http_proxy`
- `https_proxy`
- `no_proxy`

Are base32 encoded when provided to the node. These should be pro-actively (automatically on kick off) be decoded, as libraries like `requests` depend on these. 

**Desired solution**
I think it would be good to add the decoding to the `wrap` section. Alternatively we could not encode these.

**Additional context**
Now you get the error:
```
info > Reading input file /mnt/data/task-000000111/input
info > Dispatching ...
info > Module 'v6-basic-omop-queries' imported!
info > Connecting to: http://host.docker.internal:5000
error > Error encountered while calling send_http_person_count: HTTPConnectionPool(host='nb2hi4b2f4xxg4lvnfsdumzrgi4a!!!!', port=80): Max retries exceeded with url: http://host.docker.internal:5000/celery (Caused by ProxyError('Unable to connect to proxy', NameResolutionError("<urllib3.connection.HTTPConnection object at 0x7f4fc28ceb90>: Failed to resolve 'nb2hi4b2f4xxg4lvnfsdumzrgi4a!!!!' ([Errno -2] Name or service not known)")))
```

As the `requests` library reads these environment variables. A temporary workaround is to decode them in the algorithm:
```python
os.environ["HTTP_PROXY"] = get_env_var("HTTP_PROXY")
os.environ["HTTPS_PROXY"] = get_env_var("HTTPS_PROXY")
os.environ["NO_PROXY"] = get_env_var("NO_PROXY")
os.environ["http_proxy"] = os.environ["HTTP_PROXY"]
os.environ["https_proxy"] = os.environ["HTTPS_PROXY"]
os.environ["no_proxy"] = os.environ["NO_PROXY"]
```

bartvanb:
…t algorithm developers no longer need to decode the envvars themselves

Fix #1056 - but more generic as this automatically decodes *all* envvars.
I tested this locally by building an algorithm image containing the new algorithm tools

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/wrap.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/wrap.py
@@ -52,8 +52,11 @@ def wrap_algorithm(log_traceback: bool = True) -> None:
         exit(1)
     info(f"wrapper for {module}")
 
+    # Decode environment variables that are encoded by the node.
+    _decode_env_vars()
+
     # read input from the mounted input file.
-    input_file = get_env_var("INPUT_FILE")
+    input_file = os.environ["INPUT_FILE"]
     info(f"Reading input file {input_file}")
     input_data = load_input(input_file)
 
@@ -63,7 +66,7 @@ def wrap_algorithm(log_traceback: bool = True) -> None:
 
     # write output from the method to mounted output file. Which will be
     # transferred back to the server by the node-instance.
-    output_file = get_env_var("OUTPUT_FILE")
+    output_file = os.environ["OUTPUT_FILE"]
     info(f"Writing output to {output_file}")
 
     _write_output(output, output_file)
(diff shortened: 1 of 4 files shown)


vantage6-algorithm-tools/vantage6/algorithm/tools/wrap.py
```diff
@@ -170,3 +173,11 @@ def _write_output(output: Any, output_file: str) -> None:
     with open(output_file, "wb") as fp:
         serialized = serialization.serialize(output)
         fp.write(serialized)
+
+
+def _decode_env_vars() -> None:
+    """
+    Decode environment variables that are encoded by the node
+    """
+    for env_var in os.environ:
```
frankcorneliusmartin, line 182: Not all environment variables that are available in the environment are base32 encoded. Think for example about the `HOME` and `PATH` variable. Doing it like this does not break it as the function `get_env_var` also works for non encoded env vars. Could you add a comment in this function, so that it is clear that we are aware that not all environment vars are encoded.

frankcorneliusmartin (changes requested):

vantage6-algorithm-tools/vantage6/algorithm/tools/wrap.py
bartvanb: Comment added

PR #1205, opened 2024-04-15
Added error classes to use in algorithms and algorithm tools to be mo…

bartvanb:
…re specific why an algorithm error occurs

- I've not included the `UnknownError` in the implementation - that should just use the base class `AlgorithmError`
- I've not used the name `ModuleNotFoundError` because it is a Python built-in. Instead prepended `Algorithm`

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/exceptions.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/exceptions.py
@@ -1,4 +1,126 @@
-class DeserializationException(Exception):
+class AlgorithmError(Exception):
+    """Generic exception raised when an algorithm fails."""
+
+
+# ---------------- Privacy exceptions ----------------
+class PrivacyViolation(AlgorithmError):
+    """Generic exception raised for data privacy concerns."""
+
+
+class PrivacyThresholdViolation(PrivacyViolation):
+    """
+    Raised when privacy threshold is violated.
+
+    Example usage:
+    - The number of rows in the data is too low.
+    - Returning the results of the algorithm would violate privacy.
+    """
+
+
+# ---------------- Data exceptions ----------------
+
+
+class DataError(AlgorithmError):
+    """Generic error raised with data handling."""
+
+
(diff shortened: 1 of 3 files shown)


vantage6-algorithm-tools/vantage6/algorithm/tools/exceptions.py
```diff
@@ -1,4 +1,126 @@
...
+
+class CollectResultsError(ClientError):
+    """Raised when result collection fails."""
+
+
+# ---------------- Input exceptions ----------------
+
+
+class InputError(AlgorithmError):
+    """Generic error raised with input handling."""
```
frankcorneliusmartin, line 90: Is this user input?

vantage6-algorithm-tools/vantage6/algorithm/tools/exceptions.py
```diff
@@ -1,4 +1,126 @@
...
+
+
+class DeserializationError(InputError):
```
frankcorneliusmartin, line 93: If `InputError` is only used for user input, this exception should inherit from `AlgorithmError` as this error can also occur at other places.

frankcorneliusmartin (approved):

vantage6-algorithm-tools/vantage6/algorithm/tools/exceptions.py
bartvanb: can also be input of a parent task. Created a few examples

vantage6-algorithm-tools/vantage6/algorithm/tools/exceptions.py
bartvanb: Not the case as stated above

vantage6-algorithm-tools/vantage6/algorithm/tools/exceptions.py
bartvanb: I created a specific subclass `UserInputError`

PR #1216, opened 2024-04-17
Change/default log level to info

### LINKED ISSUE #692: [Change Request] Default logging level to `INFO`
**Description**
Currently the default logging level is debug at many places. We should evaluate the logging and promote some debug messages to info so we can raise the default log level to `INFO`

bartvanb:
Fix #692

Note that the server and node actually don't have a default log level - it's simply the first option in the choices so many people choose that. We could consider putting INFO as first option


vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -264,7 +264,7 @@ def get(self, id_: int) -> Any:
                     )
                 except Exception as e:
                     self.parent.log.error("Unable to load results")
-                    self.parent.log.debug(e)
+                    self.parent.log.error(e)
```
frankcorneliusmartin, line 267: Seeming this is the output that is printed to the user I would keep this one on `debug`. As the user only wants the error if he can actually debug it?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -304,7 +304,7 @@ def from_task(self, task_id: int) -> list[Any]:
                 ]
             except Exception as e:
                 self.parent.log.error("Unable to load results")
-                self.parent.log.debug(e)
+                self.parent.log.error(e)
```
frankcorneliusmartin, line 307: Same for this one

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -42,12 +42,12 @@ def check_docker_running() -> None:
...
         docker_client = docker.from_env()
         docker_client.ping()
-    except Exception as e:
+    except Exception as exc:
         log.error(
             "Cannot reach the Docker engine! Please make sure Docker " "is running."
         )
-        log.warn("Exiting...")
-        log.debug(e)
+        log.error(exc)
```
frankcorneliusmartin, line 49: `log.exception` ?

vantage6-node/vantage6/node/__init__.py
```diff
@@ -686,10 +687,8 @@ def setup_squid_proxy(self, isolated_network_mgr: NetworkManager) -> Squid:
                 isolated_network_mgr, config, self.ctx.name, volume, custom_squid_image
             )
         except Exception as e:
-            self.log.critical(
-                "Squid proxy failed to initialize. " "Continuing without."
-            )
-            self.log.debug(e, exc_info=True)
+            self.log.critical("Squid proxy failed to initialize. Continuing without.")
+            self.log.info(e, exc_info=True)
```
frankcorneliusmartin, line 691: Why not use `self.log.exception` here?

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -507,8 +507,8 @@ def run(
 
         # Check that this task is not already running
         if self.is_running(run_id):
-            self.log.warn("Task is already being executed, discarding task")
-            self.log.debug(f"run_id={run_id} is discarded")
+            self.log.warning("Task is already being executed, discarding task")
```
frankcorneliusmartin, line 510: Lower to `info` as this is not really anything to worry about.

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -651,7 +651,7 @@ def login_to_registries(self, registries: list = []) -> None:
                 self.log.info(f"Logged in to {registry.get('registry')}")
             except docker.errors.APIError as e:
                 self.log.warn(f"Could not login to {registry.get('registry')}")
-                self.log.debug(e)
+                self.log.warn(e)
```
frankcorneliusmartin, line 654: I think `warn` is deprecated? `self.log.warning`

vantage6-node/vantage6/node/docker/task_manager.py
```diff
@@ -140,8 +140,8 @@ def is_finished(self) -> bool:
             self.container.reload()
         except (docker.errors.NotFound, AttributeError):
             self.log.error("Container not found")
```
frankcorneliusmartin, line 142: I think that this should be a warning as it should never be happening

vantage6-node/vantage6/node/docker/task_manager.py
```diff
@@ -519,7 +519,7 @@ def _setup_environment_vars(
                     f"{self.databases.keys()}. This is likely to "
                     "result in an algorithm crash."
                 )
-                self.log.debug(f"User specified database: {database}")
+                self.log.warning(f"User specified database: {database}")
```
frankcorneliusmartin, line 522: Why warning? I think this should be an info?

vantage6-server/vantage6/server/__init__.py
```diff
@@ -344,10 +344,10 @@ def error_remove_db_session(error: HTTPException):
             sessions.
             """
             if error.code == 404:
-                log.debug(f"404 error for route '{_get_request_path(request)}'")
+                log.debug("404 error for route '%s'", _get_request_path(request))
             else:
-                log.warn("HTTP Exception occured during request")
-                log.debug(traceback.format_exc())
+                log.warning("HTTP Exception occured during request")
+                log.warning(traceback.format_exc())
```
frankcorneliusmartin, line 350: Are we sure we want to warn the traceback? I think this info is only used when actually debugging such events?

vantage6-server/vantage6/server/mail_service.py
```diff
@@ -42,7 +42,7 @@ def _send_async_email(self, app: Flask, msg: Message) -> None:
                 self.mail.send(msg)
             except Exception as e:
                 log.error("Mailserver error!")
-                log.debug(e)
+                log.error(e)
```
frankcorneliusmartin, line 45: Why not use the `log.exception`?

vantage6-server/vantage6/server/resource/health.py
```diff
@@ -65,6 +65,6 @@ def get(self):
             db_ok = True
         except Exception as e:
             log.error("DB not responding")
-            log.debug(e)
+            log.error(e)
```
frankcorneliusmartin, line 68: `exception`?

vantage6-server/vantage6/server/resource/recover.py
frankcorneliusmartin, line 1: I am not sure if we want to elevate both logs to `info`?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: That leads to users not seeing the debug message and having no idea how to debug. I think it is much clearer if the actual error gets highlighted - if they can't solve it themselves they will also often ask a better question.
I did change this to `exception` as well.

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: Also changed to `exception`

vantage6-node/vantage6/node/docker/task_manager.py
bartvanb: If it should never be happening, I think it should be an error when it does

vantage6-node/vantage6/node/docker/task_manager.py
bartvanb: I merged it now with the warning just before it

vantage6-server/vantage6/server/__init__.py
bartvanb: It's good if they stand out. By putting them as debug statements, people deploying the server that are not familiar with the code will not link a debug statement to that warning.

vantage6-server/vantage6/server/resource/recover.py
bartvanb: I cannot see which lines this comment refers to?

frankcorneliusmartin: It to all changes in this file. I commented on the entire file.

bartvanb: Ah right, I couldn't see the file changes in my view so then "both" was a bit difficult to find. Now that I saw, yes, I would put both to info, as information about password resets seems important enough to log as non-debug

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
frankcorneliusmartin: I disagree on this. If we could provide a helpful message I would agree, but this exception is super generic. We also don't show these in the UI client. I think if a user want to debug it should set the log level to debug.

vantage6-server/vantage6/server/__init__.py
frankcorneliusmartin: Why not make the first a `warning` that will stand out and the second one a debug? As you would only want to have it when you are debugging right?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: We agreed to log.exception

vantage6-node/vantage6/node/docker/task_manager.py
bartvanb: we agreed to keep error

vantage6-server/vantage6/server/__init__.py
bartvanb: We agreed that second can go to debug. But also, and in general in future, we should clarify in these cases that the debug belongs to the warning

PR #1226, opened 2024-04-29
Change/expand algorithm tools envvar capabilities

bartvanb:
Changes related to review of crosstab (https://github.com/vantage6/v6-crosstab-py/pull/5)

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/util.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/util.py
@@ -162,5 +162,5 @@ def _convert_envvar_to_int(
         raise EnvironmentVariableError(
             f"Environment variable '{envvar_name}' has value '{envvar_value}' but a "
             "positive integer value is required."
-        ) from exc
+        )
     return envvar_value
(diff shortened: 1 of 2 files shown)


vantage6-algorithm-tools/vantage6/algorithm/tools/util.py
```diff
@@ -42,10 +43,12 @@ def error(msg: str) -> None:
...
 
-# TODO v5+ move this function to wrap.py and no longer expose it to be used by
-# algorithms but as part of _decode_env_vars. It is kept here for backwards
-# compatibility with 4.2/4.3 algorithms
-def get_env_var(var_name: str, default: str | None = None) -> str:
+def get_env_var(
+    var_name: str,
+    default: str | None = None,
+    as_type="str",
+    only_positive: bool = False,
```
frankcorneliusmartin, line 50: I find this a little bit of a awkward place to check wether a value is positive?

bartvanb: I agree, but I do think it's a good check to have. Do you have a suggestion? Would it be better to create a separate function for this?

frankcorneliusmartin: I think indeed a separate function that is called validate or something similar would be more appropriate

frankcorneliusmartin (approved):

Issue #1237, opened 2024-05-06
[Bug] Error on the v6-node when creating a task that uses SQL data sources


hcadavid:
**Describe the bug**

When creating a Task using an SQL data source defined for the target node ([as described on v6's documentation](https://docs.vantage6.ai/en/main/user/pyclient.html)), the following error is reported on the node logs:

```
    raise ImportError("Using URI string without sqlalchemy installed.")
ImportError: Using URI string without sqlalchemy installed.
```

This is caused by the way the 'read_sql' method is used on the [algorithm wrapper](https://github.com/vantage6/vantage6/blob/b4f9676f99d5383a44780689a7dd60c67a73bb85/vantage6-algorithm-tools/vantage6/algorithm/tools/wrappers.py#L267). According to [the documentation](https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html) this method should have a 'connection' parameter, but in the implementation only a URI and the query are given.

```
return pd.read_sql(database_uri, query)
```

**To Reproduce**

1. Create an SQLite database
2. Add it as an SQL database on a node
3. Run a task on that node, using the database, either using the UI or the python client.
```
average_task = client.task.create(
   collaboration=2,
   organizations=[2],
   name="max-task",
   image="hectorcadavid/v6-max-alg",
   description='',
   input_=input_,
   databases=[
      {'label': 'employees', 'query': 'SELECT * FROM employees'}
   ]
)
```

**Screenshots**
![image](https://github.com/vantage6/vantage6/assets/1377051/d5ecf685-0b39-4c33-9237-02bd99913145)
![image](https://github.com/vantage6/vantage6/assets/1377051/3546d376-5344-445f-8a95-8cdea3b809ed)

<img width="1208" alt="image" src="https://github.com/vantage6/vantage6/assets/1377051/321f9a9d-90e0-4017-a42b-a083c3cbabff">

**Details**
 - OS: Ubuntu server 22.04
 - Vantage6 4.4.0

PR #1244, opened 2024-05-07
Fix/sql node database error

hcadavid:
Addresses issue https://github.com/vantage6/vantage6/issues/1237. This implementation should allow the integration of any sql database supported by sqlalchemy by using the URI conventions defined by this toolkit: https://docs.sqlalchemy.org/en/20/core/engines.html


hcadavid:
I created a 'patched' version of a wrapped algorithm (with the proposed updates) for further testing and found out that other changes are needed to have consistency between URI- and file-based SQL databases. I'm closing the PR for the time being.

PR #1248, opened 2024-05-08
Encrypt and decrypt data in the UI

bartvanb:
Also makes some changes to the Python decryption: in the UI it is not possible to directly encrypt bytes so there is another layer of b64 encoding that needs to be removed in Python


frankcorneliusmartin (approved):
Did check docs and backend code. Lets test the UI part

PR #1266, opened 2024-05-15
Removed username field from password recovery form

### LINKED ISSUE #1121: [Feature Request] Make more clear 'username' or 'email' alternative in password recovery
**Problem description**
I very often notice people try to fill both fields, as most people don't read the text and just know they have to "fill something" to continue with password recovery. So they miss that `or`. I missed it too.

**Desired solution**
Either two tabs within the "Lost your password?" card you can click to switch between "email input" or "username input". Or, more simply, since we are not allowing the character '@' as username anymore, just provide a single field like [twitter for example does]( https://twitter.com/account/begin_password_reset). And then figure out if it's a valid username or email address server side.

image

**Additional context**
Other sites (github, gitlab) only accept 'email', I suspect this might be to reduce illegitimate password reset requests – an attacker would need to know the email associated to the github account they want to gain access to (?). But in vantage6's there are no public profiles, and I think a logged in user, if they can see another user listed, they can probably see both username and email just as easily (?). So allowing for username or email seems fine IMHO.

frankcorneliusmartin:
Fixes #1121 

Note that I left in the code that can handle both email and username. I felt like leaving it in in case we had a change or hearth. But let me know if it needs te be removed.

### DIFF
--- a/vantage6-ui/src/app/models/forms/login-form.model.ts
+++ b/vantage6-ui/src/app/models/forms/login-form.model.ts
@@ -15,6 +15,5 @@ export interface MFAResetTokenForm {
 }
 
 export interface LostPasswordForm {
-  email?: string;
-  username?: string;
+  email: string;
 }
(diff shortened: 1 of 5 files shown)


bartvanb (changes requested):
This works but it leaves unused code. The typescript file should be adjusted as well to remove the `username` from the form, and then also the call to the API should be adjusted.

frankcorneliusmartin:
image
This wil remain as this is a server response.

bartvanb (approved):

PR #1268, opened 2024-05-15
Bugfix/unfixed proxy server return status codes

### LINKED ISSUE #1241: [Bug] Proxy server reports status code 200 even when other status code is returned from the server
**Describe the bug**

https://github.com/vantage6/vantage6/blob/b4f9676f99d5383a44780689a7dd60c67a73bb85/vantage6-node/vantage6/node/proxy_server.py#L277-L286

By doing this the `ContainerClients` will not retry the request when a non 2xx code is returned from the server.

frankcorneliusmartin:
Fixes #1241 

I also did some small maintenance

### DIFF
--- a/vantage6-node/vantage6/node/proxy_server.py
+++ b/vantage6-node/vantage6/node/proxy_server.py
@@ -10,10 +10,11 @@ to access other places in the network.
 import requests
 import logging
 
+from time import sleep
 from http import HTTPStatus
 from requests import Response
 
-from flask import Flask, request, jsonify
+from flask import Flask, request
 
 from vantage6.common import bytes_to_base64s, base64s_to_bytes, logger_name
 from vantage6.common.client.node_client import NodeClient
@@ -46,7 +47,7 @@ def get_method(method: str) -> callable:
     """
     method_name: str = method.lower()
 
-    loopup = {
+    method_map = {
         "get": requests.get,
         "post": requests.post,
         "patch": requests.patch,
@@ -54,7 +55,7 @@ def get_method(method: str) -> callable:
         "delete": requests.delete,
     }
 
-    return loopup.get(method_name, requests.get)
+    return method_map.get(method_name, requests.get)
 
 
 def make_proxied_request(endpoint: str) -> Response:
@@ -137,16 +138,17 @@ def make_request(
 
         except Exception:
             log.exception(
-                f"On attempt {i}, the proxy request raised an " f"exception: <{url}>"
+                f"On attempt {i} to reach {url}, the proxy request raised an exception"
             )
+            sleep(1)
 
-    # if all attemps fail, raise an exception to be handled by its parent
+    # if all attempts fail, raise an exception to be handled by its parent
     raise Exception("Proxy request failed")
 
 
 def decrypt_result(run: dict) -> dict:
     """
-    Decrypt the `result` from a run dictonary
+    Decrypt the `result` from a run dictionary
 
     Parameters
     ----------
@@ -213,7 +215,10 @@ def proxy_task():
             "Task proxy request received but proxy server was not "
             "initialized properly."
         )
-        return jsonify({"msg": "Proxy server not initialized properly"}), 500
+        return (
+            {"msg": "Proxy server not initialized properly"},
+            HTTPStatus.INTERNAL_SERVER_ERROR,
+        )
 
     # All requests from algorithms are unencrypted. We encrypt the input
     # field for a specific organization(s) specified by the algorithm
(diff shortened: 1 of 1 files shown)


vantage6-node/vantage6/node/proxy_server.py
```diff
@@ -283,7 +290,7 @@ def encrypt_input(organization: dict) -> dict:
             "msg": "Request failed, see node logs"
         }, HTTPStatus.INTERNAL_SERVER_ERROR
 
-    return response.json(), HTTPStatus.OK
+    return response.content, response.status_code, response.headers.items()
```
bartvanb, line 293: You are now also returning the headers items - why? 

We should also update the docstring as it is not a Response that is returned but a dict and a status code (and maybe also a header dict). If we keep the header dict, we should also adapt it in line 291 and in the docstring, and verify that this response is actually dealt with properly (I think this isn't the case currently because the return is fed into the algorithm client right?)

vantage6-node/vantage6/node/proxy_server.py
```diff
@@ -137,16 +138,17 @@ def make_request(
 
         except Exception:
             log.exception(
-                f"On attempt {i}, the proxy request raised an " f"exception: <{url}>"
+                f"On attempt {i} to reach {url}, the proxy request raised an exception"
             )
+            sleep(1)
 
-    # if all attemps fail, raise an exception to be handled by its parent
+    # if all attempts fail, raise an exception to be handled by its parent
```
bartvanb, line 145: This line is not just activated if all attempts fail, if we go through the except above we also get here. So maybe modify the comment?

vantage6-node/vantage6/node/proxy_server.py
```diff
@@ -326,13 +333,13 @@ def proxy_result() -> Response:
...
 
-    # Attempt to decrypt the results. The enpoint should have returned
+    # Attempt to decrypt the results. The endpoint should have returned
     # a list of results
     results = get_response_json_and_handle_exceptions(response)
     for result in results["data"]:
         result = decrypt_result(result)
 
-    return jsonify(results), HTTPStatus.OK
+    return results, response.status_code, response.headers.items()
```
bartvanb, line 342: again the question why we return a header dict

vantage6-node/vantage6/node/proxy_server.py
```diff
@@ -371,7 +378,7 @@ def proxy_results(id_: int) -> Response:
     result = get_response_json_and_handle_exceptions(response)
     result = decrypt_result(result)
 
-    return result, HTTPStatus.OK
+    return result, response.status_code, response.headers.items()
```
bartvanb, line 381: and again

vantage6-node/vantage6/node/proxy_server.py
frankcorneliusmartin: This is the default Flask structure for returning a response, see https://flask.palletsprojects.com/en/3.0.x/quickstart/#about-responses

`return <body>, <status_code>, <headers>` 

Then on the client side (=algorithm client) the headers are also returned in the header part, so that should not break any existing algorithms. Returning the headers can be useful for pagination, something that the algorithm needs to be able to deal with at some point.

I am not sure why we should change line 291? At line 291 the request failed so we do not have access to headers that need to be forwarded?

The doc string reports that a `requests.Response` is returned. While in fact we are returning a tuple, however flask converts this tuple into a `requests.Response`.. We could change it to tuple and then describe the tuple components?

vantage6-node/vantage6/node/proxy_server.py
frankcorneliusmartin: You *only* get here when you go through the exception above, right? So the statement seems correct to me?

vantage6-node/vantage6/node/proxy_server.py
bartvanb: Yes of course, silly of me. This seems indeed appropriate

vantage6-node/vantage6/node/proxy_server.py
bartvanb: On reading this again, I think I misunderstood the comment. I thought it meant "when all N attempts to send proxy requests fail" while I now think it means "all attempts in this iteration fail". I was confused with the `client.request()` method where we do a retry if a request fails.

Maybe we can change the wording of the comment a bit to something like: "The request failed. Raise an exception".

PS: If I'm not mistaken, there are two routes you can get here: 1) via the `if` in line 122 and 2) via the `except` in line 139. 

PPS: would it actually be an idea to implement a retry in the proxy server? Might lead to less algorithm fails - but maybe too much effort as proxy server will be replaced/rewritten at some point

frankcorneliusmartin: You only get here if all `RETRY` attempts have not reached line 137 ` return`. A retry is exhausted when one of the following events happens

1. a `> 210` status code is returned from the server (the connection with the server was successful)
2. an exception is raised, most likely the client was not able to reach the server

So this statement:
> when all N attempts to send proxy requests fail

Seems true to me. And in that case the comment seems oke to me?

Issue #1269, opened 2024-05-15
[Docs] Document Azure deployment


frankcorneliusmartin:
**Description**
Add Azure deployment to the docs (see internal docs)

* [ ] Sticky sessions
* [ ] Docker compose
* [ ] Upload of config file
* [ ] ...

frankcorneliusmartin:
Alreadt released with v4.5

PR #1270, opened 2024-05-15
Added sticky sessions requirement to the docs for horizontal scaling

### LINKED ISSUE #1224: [Docs] document sticky session requirement
**Description**
For websocket connection the server needs to be configured to enable sticky sessions. We should add this to the server documentation on server deployment, ideally with example for nginx and/or Azure.

frankcorneliusmartin:
Fixes #1224 

I did not document the ARR affinity option for Azure as there was no section for this yet. I opened a separate issue for this: https://github.com/vantage6/vantage6/issues/1269

### DIFF
--- a/docs/server/deploy.rst
+++ b/docs/server/deploy.rst
@@ -16,6 +16,7 @@ There are many deployment options. We simply provide a few examples.
 -  …
 
 .. note::
+
     Because the server uses socketIO to exchange messages with the nodes and
     users, it is not trivial to horizontally scale the server. To prevent that
     socket messages get lost:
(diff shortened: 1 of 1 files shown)


docs/server/deploy.rst
```diff
@@ -18,32 +18,46 @@ There are many deployment options. We simply provide a few examples.
...
 
 .. _deploy-nginx:
 
 NGINX
 """""
 
-A basic setup is shown below. Note that SSL is not configured in this example.
+A basic setup is shown below. This example shows how to configure NGINX where there
+are 3 backend servers that are load balanced. The load balancer is configured to use
+sticky sessions.
```
bartvanb, line 34: Since most people will deploy a single server, maybe it is good to explicitly mention that you can also just have one server.

docs/server/deploy.rst
```diff
@@ -53,7 +67,7 @@ A basic setup is shown below. Note that SSL is not configured in this example.
            proxy_buffering off;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "Upgrade";
-           proxy_pass http://127.0.0.1:5000/socket.io;
+           proxy_pass http://backend/socket.io;
```
bartvanb, line 70: I think for most people the localhost address will be what they need in nginx, and by changing it to backend here, the docs are less easy to understand. 
Maybe we can have two examples: the original one and this updated version. We can also offer them as file downloads (as we already have in the configuration section)

bartvanb (commented):
I think retaining the original (basic) nginx example is important

docs/server/deploy.rst
frankcorneliusmartin: Backend is what is defined in the `upstream` section of this same NGINX config file, that is pretty common to do. It is also not possible to specify the IP here as this is the load balancing example, so we somehow need to tell NGINX that there are multiple instances of `backend`

But I agree on the show both examples and in that case we can specify the localhost IP `127.0.0.1` in this directive block.

docs/server/deploy.rst
frankcorneliusmartin: Fair, I re added the original example and extended the test a little.

bartvanb (commented):
Thanks for the changes! One more question: in the original (simple) example, should we not also put the sticky sessions? (in other words, maybe the original example was misconfigured regarding sticky sessions? :"-) )

frankcorneliusmartin:
> Thanks for the changes! One more question: in the original (simple) example, should we not also put the sticky sessions? (in other words, maybe the original example was misconfigured regarding sticky sessions? :"-) )

Why would you need sticky sessions if you only have one worker?

bartvanb:
Is indeed not needed, sticky sessions only required for multiple server workers

PR #1272, opened 2024-05-15
For other and omop database types the column names algorithm is no lo…

### LINKED ISSUE #1117: [Feature Request] `column_names` for `database_connector` types
**Problem description**
Currently the `v6-basic-py` is send when a task is created to obtain column names. In the case a `database_connection` decorator is used in the algorithm this `column_names` method wont work. This results in a lot of failed tasks in the overview (and an error bar when creating a task)

**Desired solution**
When encountering a database which is of type other (or omop) we should not send this `column_names`.

**Additional context**
This will be unified for all data sources when the chaining feature is implemented.

frankcorneliusmartin:
…nger send

Fixes #1117

### DIFF
--- a/vantage6-ui/src/app/models/api/task.models.ts
+++ b/vantage6-ui/src/app/models/api/task.models.ts
@@ -122,7 +122,6 @@ export interface TaskResult {
 
 export interface TaskDatabase {
   label: string;
-  type?: string;
   query?: string;
   sheet_name?: string;
 }
--- a/vantage6-ui/src/app/models/api/task.models.ts
+++ b/vantage6-ui/src/app/models/api/task.models.ts
@@ -122,6 +122,7 @@ export interface TaskResult {
 
 export interface TaskDatabase {
   label: string;
+  type?: string;
   query?: string;
   sheet_name?: string;
 }
--- a/vantage6-ui/src/app/pages/task/create/steps/database-step/database-step.component.scss
+++ b/vantage6-ui/src/app/pages/task/create/steps/database-step/database-step.component.scss
@@ -1,3 +1,5 @@
+@import "../../../../../styles/config.scss";
+
 .field-description {
     // font-style: italic;
     margin-bottom: $spacing-px-2;
--- a/vantage6-ui/src/app/pages/task/create/steps/database-step/database-step.component.scss
+++ b/vantage6-ui/src/app/pages/task/create/steps/database-step/database-step.component.scss
@@ -1,4 +1,8 @@
-.field-description {
-  // font-style: italic;
-  margin-bottom: $spacing-px-2;
-}
+
+
+.detail-txt {
+    font-size: small;
+    color: grey;
+    display: inline;
+  }
+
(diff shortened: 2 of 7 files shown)


vantage6-ui/src/app/pages/task/create/steps/database-step/database-step.component.scss
```diff
@@ -1,4 +1,8 @@
-.field-description {
-  // font-style: italic;
-  margin-bottom: $spacing-px-2;
-}
```
bartvanb, line 4: why remove the field-description class? It is still being used in the HTML

vantage6-ui/src/app/models/api/task.models.ts
```diff
@@ -122,6 +122,7 @@ export interface TaskResult {
 
 export interface TaskDatabase {
   label: string;
+  type?: string;
```
bartvanb, line 125: Note that `TaskDatabase` is an interface that is part of the `CreateTask` interface. This means that when you include a `type` property, it will be sent to the API when creating the task. 

For what you want to accomplish, using the existing `Database` interface may be better (this has a name (aka label) and type property).

vantage6-ui/src/app/pages/task/create/steps/database-step/database-step.component.scss
frankcorneliusmartin: It wasn't used as this file wasn't imported. So I removed it, but maybe I should have fixed it rather then delete it ;-). I added it back.

vantage6-ui/src/app/models/api/task.models.ts
frankcorneliusmartin: Thanks for the explanation!

vantage6-ui/src/app/pages/task/create/steps/database-step/database-step.component.scss
bartvanb: Haha oops, my bad. It's just a bit of spacing so easy to overlook, but indeed lets keep it!

vantage6-ui/src/app/pages/task/create/task-create.component.ts
```diff
@@ -298,18 +298,21 @@ export class TaskCreateComponent implements OnInit, OnDestroy, AfterViewInit {
...
     // are implemented
-    if (taskDatabase.type == 'other' || taskDatabase.type == 'omop') {
+    const database = databases[0];
+    if (database.type == 'other' || database.type == 'omop') {
         this.isLoadingColumns = false;
         return;
     }
 
+    // TODO modify when choosing database for preprocessing is implemented
+    const taskDatabase = taskDatabases[0];
```
bartvanb, line 314: I'm just realizing now that this is an existing bug... in this function, the columns are retrieved but because of this line, they are only retrieved for the first database. We should create an issue of this and then merge this

bartvanb (approved):
Nice! Approved, but not merged to remind us to submit an issue for an existing bug that I found (see below)

PR #1279, opened 2024-05-21
Use sphinx autosummary to generate function docs so that we no longer…

bartvanb:
… need to create them per function


vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -1,3 +1,5 @@
+""" Client for the algorithm container to communicate with the central server. """
```
frankcorneliusmartin, line 1: *vantage6* server (at least, I thought we decided to not use *central* anymore). I also see that you used the *vantage6* convention in the other doc strings

frankcorneliusmartin (approved):
Just a minor doc change, already approved

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: adapted

PR #1296, opened 2024-05-27
Development debugging environment

lsago:
For people less familiar with the code base, being able to set breakpoints can come in handy when developing new features / trying to fix bugs. 

I suggest we use this PR to start a discussion about to approach it. I've made many arbitrary choices, including making some files part of the repo itself.. which might not the best way to go about it.


bartvanb (commented):
I just got round to having a look at this. I am curious to see how this works in practice - let's plan a small meeting for that as you suggested.

My main question so far is whether we could maybe use the `v6 dev` functionality more. There, we already have templates for configuration files and commands to spin up a server and nodes locally. I think it would be nice integrating that with this feature as I think it will make it easier to maintain both in the future.

bartvanb:
I thought about this again, and also discussed briefly with @frankcorneliusmartin. We have the feeling that probably the implementation can integrate nicely with `v6 dev` but it is quite a hassle to figure that out. The extra maintenance burden of updating the config files etc is probably low. Meanwhile, this is a valuable feature that saves time for developers. 

So @lsago I propose you polish it up a bit and then we can merge it! At a later point in time, we can always make the decision to integrate it more tightly with features like `v6 dev`, if and when we see a need for that.

PR #1297, opened 2024-05-28
Feature/algorithm store policies

### LINKED ISSUE #1026: [Feature Request] Settings management for algorithm store
**Problem description**
Algorithm stores will have a set of policies. It is important to make sure that these settings are visible e.g. in the UI. For example, they can be used to determine if users can see algorithms from a store.

**Desired solution**

- An extendible management system for algorithm store settings
- Policies should be defined in the config file
- The following settings could be implemented at this point:
  - Whether viewing algorithms is open to anyone or only to users of whitelisted servers
  - Is store accessible for any server or otherwise, list of servers that are allowed to connect to the algorithms
  - List of servers where (part of) users may be are allowed to edit algorithms in the store
  - (maybe) Whether `localhost` servers are allowed to be whitelisted (only for dev stores)

**Later?**
  - For now, policies can **only** be defined in the config file. At a later stage, we may want them to be editable in e.g. the UI

bartvanb:
Fix #1026


bartvanb:
Superseded by #1299 where these commits are also included

Issue #1198, opened 2024-04-10
[Feature Request] Indicate in task from which algorithm store the algorithm was run


bartvanb:
**Problem description**
When running a task using the UI, there is no way to find out at a later stage from which algorithm store an image was run. Then, when accessing the results of the task later, it is unclear which store to get the algorithm from. 

This has two disadvantages: 1) multiple requests have to be done and more importantly 2), the algorithm metadata could differ between algorithm stores and we want to show the original metadata.

**Desired solution**
- [x] Add additional database field for tasks to couple them to an algorithm store
- [x] Modify GET requests to get the algorithm store id field as well
  - [x] python client
  - [x] algorithm client
  - [x] UI
- [x] Modify POST /task endpoint
  - [x] python client 
  - [x] UI
  - [x] algorithm client
- [x] when DELETE /algorithm-store is run, ensure that these IDs are cleared
- [x] ~~Patch modifications?~~

PR #1318, opened 2024-06-12
Node policy to set allowed algorithms by algorithm store

### LINKED ISSUE #1293: [Feature Request] Option for node administrator to alllow all algorithms from certain store
**Problem description**
It would be very convenient if a node admin can indicate that they trust all algorithms from a certain algorithm store. Then, the store can indicate if an algorithm (including the sha256) is available in the store.

**Desired solution**
- [x] Store: when algorithm is submitted to store, get the latest sha256
- [x] Store: option to fetch new latest sha256 ~~or to change it manually~~
- [x] When algorithm is submitted to server, pick up the digest from the algorithm store
  - [x] Implement #1198 as part of this
  - [x] Make sure it can also be set in the python client
  - [x] Use OCI v2 to determine digest without pulling the image
    - [x] Adjust docs to document OCI v2 requirement for docker registries
    - [x] authentication at registry for manifest request
  - [x] inform the node which algorithm store is used to pick up the digest when sending the socket event
  - [x] Run new tests for vantage6-common automatically
- [x] Node: implement policy.
  - [x] Should be AND / OR combinable with other algorithm policies
- [x] Adapt `v6 node new` to ask what to set

bartvanb:
Fix #1293, #1198

- node can define policies to allow any algorithm from a whitelisted algorithm store
- algorithm store saves algorithm image digest when it is posted to the store using manifest from OCI v2 API
- task have new field `store_id` to save which store algorithm was taken from (optional - you can still run algorithms outside the store)
- when task is created from a certain algorithm store, it verifies with the store that the algorithm is in there. Also, the digest is stored as part of the image even if the user did not specify it - this way it can be checked later with which algorithm image version the algorithm was run
- **code to check `allowed_algorithms` has been updated**
- UI, python client and algorithm client have been updated where necessary


bartvanb:
We can ignore the codacy issue

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -744,19 +801,6 @@ def post_task(data: dict, socketio: SocketIO, rules: RuleCollection):
...
-            "task_created",
-            {
-                "task_id": task.id,
-                "job_id": task.job_id,
-                "collaboration_id": collaboration_id,
-                "init_org_id": init_org.id,
-            },
-            room=f"collaboration_{collaboration_id}",
-            namespace="/tasks",
-        )
```
bartvanb, line 758: note that this socket event was not listened to anywhere

vantage6-server/vantage6/server/algo_store_communication.py
```diff
@@ -191,12 +192,11 @@ def request_algo_store(
         request to the algorithm store is unsuccessful, a dict with an error message is
         returned instead of the response.
     """
-    # TODO this is not pretty, but it works for now. This should change
-    # when we have a separate auth service
     is_localhost_algo_store = _contains_localhost(algo_store_url)
+    print("params in request_algo_store", params)
```
wbaccinelli, line 196: was this for debugging?

vantage6-server/vantage6/server/model/task.py
```diff
@@ -50,6 +50,8 @@ class Task(Base):
         Organization that created this task
     init_user : :class:`~.model.user.User`
         User that created this task
+    databases : list[:class:`~.model.task_database.TaskDatabase`]
+        List of databases that are part of this task
```
wbaccinelli, line 54: should also `alsgorithm_store` be added to the attributes?

wbaccinelli (approved):
Looks good! I don't have any useful input for the PR 😅

vantage6-server/vantage6/server/algo_store_communication.py
bartvanb: thanks for spotting that

vantage6-server/vantage6/server/model/task.py
bartvanb: Indeed! And there were actually also a couple of others missing so added those as well

vantage6-algorithm-store/vantage6/algorithm/store/resource/__init__.py
```diff
@@ -296,7 +296,7 @@ def decorator(*args, **kwargs):
     return protection_decorator
 
 
-def with_permission_to_view_algorithms(resource: str, operation: Operation) -> callable:
+def with_permission_to_view_algorithms() -> callable:
```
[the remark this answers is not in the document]
bartvanb: Done

vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -182,6 +183,77 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
...
+    Returns
+    -------
+    list[str]
+        List of allowed algorithm stores
+    """
+    info("Below you can add algorithm stores that are allowed to run on your node.")
+    info(
+        "You can use regular expressions to match multiple algorithms, or you can "
+        "use strings to provide one algorithm at a time."
+    )
```
frankcorneliusmartin, line 236: ... match multiple **algorithm stores** ...
and
... one algorithm **store** at the time.

vantage6-algorithm-store/vantage6/algorithm/store/resource/algorithm.py
```diff
@@ -93,10 +98,56 @@ def permissions(permissions: PermissionManager) -> None:
...
+
+        This method is run in a separate thread to prevent the request from taking too
+        long. Do NOT call this method from the main thread!
+
+        Parameters
+        ----------
+        image : str
+            Image url
+        config : dict
+            Configuration dict
```
frankcorneliusmartin, line 116: `config` is not an parameter

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -302,3 +300,89 @@ def delete_volume_if_exists(client: docker.DockerClient, volume_name: Volume) ->
...
+    if (
+        response.status_code == HTTPStatus.UNAUTHORIZED
+        and registry_user
+        and registry_password
+    ):
+        response = requests.get(
+            manifest_endpoint,
+            auth=HTTPBasicAuth(registry_user, registry_password),
+            timeout=60,
+        )
```
frankcorneliusmartin, line 378: I am a little worried about this as the basic auth is not specified by OCI. I found this source that tells me that bearer is also used by Github Container Registry and Azure Containers:

https://github.com/richlander/container-registry-api?tab=readme-ov-file#authorization

Which registries did you check?

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -346,7 +330,65 @@ def is_docker_image_allowed(self, docker_image_name: str, task_info: dict) -> bo
...
+            for algorithm in allowed_algorithms:
+                if not self._is_regex_pattern(algorithm):
+                    # check if string matches exactly
+                    if algorithm == docker_image_name:
+                        algorithm_whitelisted = True
+                    # if allowed algorithm is "some/image", but the docker image
+                    # includes the hash (e.g. "some/image@sha256:..."), we also
+                    # want to allow it
+                    elif docker_image_name.startswith(f"{algorithm}@sha256:"):
+                        algorithm_whitelisted = True
```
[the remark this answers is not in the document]
frankcorneliusmartin: And how about specific tags from an image?

vantage6-server/vantage6/server/model/task.py
```diff
@@ -29,28 +29,39 @@ class Task(Base):
...
         Organization that created this task
     init_user : :class:`~.model.user.User`
         User that created this task
-
+    databases : list[:class:`~.model.task_database.TaskDatabase`]
+        List of databases that are part of this task
+    study : :class:`~.model.study.Study`
+        Study that this task belongs to
+    algorithm_store : :class:`~.model.algorithm_store.AlgorithmStore`
+        Algorithm store that this task belongs to
```
frankcorneliusmartin, line 64: I would rather say 
 
```suggestion
        Algorithm store that this task uses
```

vantage6-algorithm-store/vantage6/algorithm/store/resource/algorithm.py
```diff
@@ -93,10 +98,56 @@ def permissions(permissions: PermissionManager) -> None:
...
+            if reg["registry"] == registry:
+                registry_user = reg.get("username")
+                registry_password = reg.get("password")
+                break
+        response = get_manifest(
+            image_name, registry, image, tag, registry_user, registry_password
+        )
+
+        # get the digest from the response headers
+        return image_wo_tag, response.headers["Docker-Content-Digest"]
```
frankcorneliusmartin, line 144: What are the consequences if this header is not present? This is an optional header as specified here: https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#legacy-docker-support-http-headers

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -539,11 +553,11 @@ def post(self):
 
         tags: ["Task"]
         """
-        return self.post_task(request.get_json(), self.socketio, self.r)
+        return self.post_task(request.get_json(), self.socketio, self.r, self.config)
```
frankcorneliusmartin, line 556: This endpoint will return a 500 now when someone posts without a JSON body. (I am aware that this was probably me, and this already existed)

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -676,6 +692,46 @@ def post_task(data: dict, socketio: SocketIO, rules: RuleCollection):
...
+            if (
+                not store.is_for_all_collaborations()
+                and store.collaboration_id != collaboration_id
+            ):
+                return {
+                    "msg": (
+                        "The algorithm store is not part of the collaboration "
+                        "to which the task is posted."
+                    )
+                }, HTTPStatus.BAD_REQUEST
```
frankcorneliusmartin, line 714: I don't think `BAD_REQUEST` is the best code? I think FORBIDDEN would be more appropiate?

vantage6-server/vantage6/server/resource/task.py
frankcorneliusmartin: I am a little surpised. But indeed :') There is another socket event `new_task`..

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -925,6 +969,60 @@ def _check_input_encryption(
...
+                f"Could not retrieve algorithm from store! {response.get('msg')}"
+            )
+        try:
+            algorithm = response.json()["data"][0]
+        except (KeyError, IndexError) as e:
+            raise Exception("Algorithm not found in store!") from e
+
+        image = algorithm["image"]
+        digest = algorithm["digest"]
+        if image and not digest:
```
frankcorneliusmartin, line 1017: Will it ever get here? Is `algorithm["digist"]` always present? And in case it does not have a digist it returns `""`?

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -925,6 +969,60 @@ def _check_input_encryption(
...
+
+        Raises
+        ------
+        Exception
+            If the algorithm cannot be retrieved from the store.
+        """
+        # get the algorithm from the store
+        response, status_code = request_algo_store(
+            algo_store_url=store.url,
+            server_url=get_server_url(config, request.args.get("server_url")),
```
frankcorneliusmartin, line 1001: It seems to me that and `config.get("server_url", request.args.get("server_url"))` can return `None` in `get_server_url`? I think both are not garanteed as one is dependent on the user and the other on the node administrator?

Maybe check that one of them is present and in case not, thow an error code back.

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -925,6 +969,60 @@ def _check_input_encryption(
...
+            params={"image": image},
+        )
+        if status_code != HTTPStatus.OK:
+            raise Exception(
+                f"Could not retrieve algorithm from store! {response.get('msg')}"
+            )
+        try:
+            algorithm = response.json()["data"][0]
+        except (KeyError, IndexError) as e:
+            raise Exception("Algorithm not found in store!") from e
```
frankcorneliusmartin, line 1013: Maybe add an additional `except` to catch all? `.json()` can raise `requests.exceptions.JSONDecodeError`

frankcorneliusmartin (changes requested):
Some concerns regarding the standardization of OCI and some other minor changes/suggestions.

Looks good.

vantage6/vantage6/cli/configuration_wizard.py
bartvanb: Good spot, important fix!

vantage6-algorithm-store/vantage6/algorithm/store/resource/algorithm.py
bartvanb: deleted it

vantage6-node/vantage6/node/docker/docker_manager.py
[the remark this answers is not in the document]
bartvanb: I added this TODO:
        # TODO also allow for "some/image:tag"? In that case, be sure to
        # prevent that if the allowed algorithm expression is "some_image",
        # that "some_image:443.com/whatever" is NOT allowed, i.e. the check
        # 'docker_image_name.startswith(f"{algorithm}:")' is not enough

Since it's not trivial and also wasn't possible up to now and no-one asked for it, I suggest to leave it. Note btw that the extra sha check *IS* necessary because the server will now add the sha256 to the requested algorithm image in the task if the algorithm store provides it to the server.

vantage6-server/vantage6/server/resource/task.py
bartvanb: I think it was probably me because I created the `post_task` function so that I could call it from the GET `/ui/column` endpoint I think. But anyway, to fix it I put a try/except around the get_json()

vantage6-server/vantage6/server/resource/task.py
bartvanb: I have no strong preference so I changed to FORBIDDEN as you suggested

vantage6-server/vantage6/server/resource/task.py
bartvanb: You're right, in the current implementation I expect it would fail making the request to a `None` endpoint, however now I added an exception to make the user get a clear error

vantage6-server/vantage6/server/resource/task.py
bartvanb: Algorithm digest will always be present in the store when this PR is merged, however, in current versions (4.3-4.5) it is not. So to keep stores that are updated to 4.6+ backwards compatible, this check is here. 
I'll add a TODO to remove this in 5.0

vantage6-server/vantage6/server/resource/task.py
bartvanb: I just changed the current `KeyError, IndexError` to generic `Exception` as that is most fail-proof

vantage6-algorithm-store/vantage6/algorithm/store/resource/algorithm.py
bartvanb: Very good question. The consequence would have been that the digest was not determined and therefore algorithm could not be submitted.
I have therefore now created an alternative method `__calculate_digest` which calculates the digest from the manifest.

docs/algorithm_store/yaml/algorithm_store_config.yaml
```diff
@@ -75,6 +75,13 @@ policies:
   # localhost servers to connect.
   allow_localhost: false
 
+# Credentials used to login to private Docker registries. These credentials are used
+# to e.g. find the hashes of the algorithm.
```
lsago, line 79: As per below, if docker registry is public for an image (i.e. anybody can pull it), they should also be able to pull the manifest of that image and get the digest of it.

lsago (commented):
I've only reviewed this partially, as I'm not familiar with new algo store code base.

docs/algorithm_store/yaml/algorithm_store_config.yaml
bartvanb: I don't understand what the issue is here. I guess I'm missing what you mean by 'as per below', can you clarify?

docs/node/node_config.yaml
```diff
@@ -105,10 +105,28 @@ encryption:
...
     - ^myalgorithm\.ai/some-algorithm
+
+  # It is also possible to allow all algorithms from particular algorithm stores. Set
+  # these stores here. They may be strings or regular expressions. If you don't specify
+  # this, algorithms from any store are allowed (unless other policies prevent this).
+  allowed_algorithm_stores:
+    # allow all algorithms from the vantage6 community store
+    - https://store.cotopaxi.vantage6.ai
+    # allow any store that is a subdomain of vantage6.ai
+    - ^https://*.vantage6.ai
```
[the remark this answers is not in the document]
bartvanb: Oops.. I now added two backslashes before the dots and a $ at the end.

vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -182,6 +183,77 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
...
+        "You can use regular expressions to match multiple algorithm stores, or you can"
+        " use strings to provide one algorithm store at a time."
+    )
+    info("Examples:")
+    info(
+        "https://store.cotopaxi.vantage6.ai    Allow all algorithms from the "
+        "community store"
+    )
+    info(
+        "^https://*.vantage6.ai$               Allow all algorithms from any "
```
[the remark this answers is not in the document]
bartvanb: here the $ was already there but the two \ have been added

vantage6-common/tests/vantage6/common/docker/test_addons.py
```diff
@@ -0,0 +1,136 @@
+from unittest import TestCase
+
+from vantage6.common.docker.addons import parse_image_name
+from docker.errors import InvalidRepository
+
+
+# These tests are adapted from [1]. We do this because parse_image_name use
```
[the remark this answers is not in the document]
bartvanb: Changed

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -307,3 +305,89 @@ def delete_volume_if_exists(client: docker.DockerClient, volume_name: Volume) ->
...
+    """
+    Get the manifest of an image
+
+    This uses the OCI distribution specification which is supported by all major
+    container registries.
+
+    Parameters
+    ----------
+    full_image_url: str
+        The full image url
```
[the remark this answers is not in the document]
bartvanb: Alright it may be clearer to remove this arg indeed so I'll do that, and improve the specification of the other ars

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -346,7 +330,69 @@ def is_docker_image_allowed(self, docker_image_name: str, task_info: dict) -> bo
                 )
                 return False
 
-        return True
+        algorithm_whitelisted = False
+        if allowed_algorithms:
+            if isinstance(allowed_algorithms, str):
+                allowed_algorithms = [allowed_algorithms]
+            for algorithm in allowed_algorithms:
+                if not self._is_regex_pattern(algorithm):
```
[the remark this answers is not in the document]
bartvanb: I think it's an excellent idea to separate them explicitly, as I totally agree that the current way is a bit funky. Can you create an issue for that? Otherwise let me know and I will do it.

For the time being we have to make sure that v4 nodes with regexes in `allowed_algorithms` will still be supported but we could already introduces the `allowed_algorithm_regexes` and then in v5 remove support for regex in `allowed_algorithms`. Or alternatively, we could implement this in v5 directly (should be there early 2025 so not too far away)

vantage6-node/vantage6/node/docker/docker_manager.py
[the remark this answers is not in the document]
bartvanb: I realize I was cutting corners I shouldnt cut when adding the TODO. People may already be having tag-based image filters so we should make sure they will keep working. I'll have a look at how to best implement this.

vantage6-common/vantage6/common/docker/addons.py
bartvanb: I just implemented the workflow of Luis, with a slight adaptation. To make it work on Dockerhub images, I also added the `service` parameter of the `www-authenticate` header, otherwise I still got a 401 there.

Tested this with Luis' image and also with `jhogenboom/average`, an algorithm image that Joshi made available via Dockerhub.

Note that the function does not support username/password-protected images yet - I left a TODO for that in the code, but didnt tackle it now as we don't have a use case at present.

vantage6-node/vantage6/node/docker/docker_manager.py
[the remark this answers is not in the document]
bartvanb: I now redid this implementation to allow images in the following cases:
- exact match (as available before)
- allowed algorithm is `some_image` and evaluated image is `some_image:tag` or `some_image@sha256...`
- if allowed algorithm is `some_image:tag_or_sha`  and evaluated image is `some_image:other_tag_or_sha`, the digest of both is computed and image is allowed if they are the same.

I personally think it's much nicer now but hope you can check it thoroughly as this is important to get right!

docs/node/node_config.yaml
```diff
@@ -105,10 +105,28 @@ encryption:
...
     - ^myalgorithm\.ai/some-algorithm
+
+  # It is also possible to allow all algorithms from particular algorithm stores. Set
+  # these stores here. They may be strings or regular expressions. If you don't specify
+  # this, algorithms from any store are allowed (unless other policies prevent this).
+  allowed_algorithm_stores:
+    # allow all algorithms from the vantage6 community store
+    - https://store.cotopaxi.vantage6.ai
+    # allow any store that is a subdomain of vantage6.ai
+    - ^https://*\.vantage6\.ai$
```
lsago, line 121: Heh these regexes are error prone I think

```suggestion
    - ^https://.*\.vantage6\.ai$
```

Or maybe even more precise:
```suggestion
    - ^https://[a-z]+\.vantage6\.ai$
```

vantage6-node/vantage6/node/docker/docker_manager.py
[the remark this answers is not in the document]
lsago: Nice, I agree, this, at first glace, seems sensible to me as well. However I still want to have a thorough look at this, as I agree this is an important function to get right.

But, I'm submitting this review now so as not to delay switching to using `inspect_distribution()` *in case you consider it worth it*.   .. given we're hoping to ship this sooner rather than later as you said.

vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -182,6 +183,77 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
...
+        "You can use regular expressions to match multiple algorithm stores, or you can"
+        " use strings to provide one algorithm store at a time."
+    )
+    info("Examples:")
+    info(
+        "https://store.cotopaxi.vantage6.ai    Allow all algorithms from the "
+        "community store"
+    )
+    info(
+        "^https://*\.vantage6\.ai$               Allow all algorithms from any "
```
lsago, line 243: I guess here we are also missing a dot or something (e.g. `[a-z]`) before the Kleene star..


Or, something like..
```suggestion
        "^https://[a-z]+\.vantage6\.ai$               Allow all algorithms from any "
```

vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -182,6 +183,77 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
...
+    info("Examples:")
+    # pylint: disable=W1401
+    # flake8: noqa: W605
+    info("^harbor2\.vantage6\.ai/demo/average$    Allow the demo average algorithm")
+    info(
+        "^harbor2\.vantage6\.ai/algorithms/.*   Allow all algorithms from "
+        "harbor2.vantage6.ai/algorithms"
+    )
+    info(
+        "^harbor2\.vantage6\.ai/demo/average:@sha256:82becede...$    Allow a "
```
lsago, line 209: That `:` between `average` and `@sha256` turns it into an invalid reference I think. Probably just a typo :p

docs/node/node_config.yaml
bartvanb: Good catch, I really suck at regex :-) I've added the last version.

docs/algorithm_store/yaml/algorithm_store_config.yaml
bartvanb: Ah gotcha. I removed the section now as it's not implemented

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -340,13 +326,99 @@ def is_docker_image_allowed(self, docker_image_name: str, task_info: dict) -> bo
...
+        if allowed_algorithms:
+            if isinstance(allowed_algorithms, str):
+                allowed_algorithms = [allowed_algorithms]
+            for allowed_algo in allowed_algorithms:
+                if not self._is_regex_pattern(allowed_algo):
+                    evaluated_img_wo_tag, _ = parse_repository_tag(evaluated_img)
+                    allowed_wo_tag, _ = parse_repository_tag(allowed_algo)
+                    if allowed_algo == evaluated_img:
+                        # OK if allowed algorithm and provided algorithm match exactly
+                        algorithm_whitelisted = True
```
frankcorneliusmartin, line 345: We should also break out of the loop? We've found what we are looking for? No need to scan for additional matching patterns? (Same goes for all other `True`s)

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -340,13 +326,99 @@ def is_docker_image_allowed(self, docker_image_name: str, task_info: dict) -> bo
...
                 return False
 
-        return True
+        algorithm_whitelisted = False
+        if allowed_algorithms:
+            if isinstance(allowed_algorithms, str):
+                allowed_algorithms = [allowed_algorithms]
+            for allowed_algo in allowed_algorithms:
+                if not self._is_regex_pattern(allowed_algo):
+                    evaluated_img_wo_tag, _ = parse_repository_tag(evaluated_img)
```
frankcorneliusmartin, line 341: You probably should do this outside of the for-loop.

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -340,13 +326,99 @@ def is_docker_image_allowed(self, docker_image_name: str, task_info: dict) -> bo
...
+                            self.log.warning(
+                                "Could not obtain digest for image %s",
+                                evaluated_img,
+                            )
+                        try:
+                            digest_policy_image = get_digest(allowed_algo)
+                        except:
+                            self.log.warning(
+                                "Could not obtain digest for image %s", allowed_algo
+                            )
```
frankcorneliusmartin, line 370: If either one of the exceptions is reached you should explicitly leave this block. As the verification bellow may be compromised. I dont think relying on `None` to be outputed from an exception block is safe.

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -307,3 +307,164 @@ def delete_volume_if_exists(client: docker.DockerClient, volume_name: Volume) ->
...
+
+    Returns
+    -------
+    tuple[str, str, str]
+        Registry, repository, and tag. Tag is "latest" if not specified in 'image'
+    """
+    registry_repository, tag = parse_repository_tag(image)
+    tag = tag or "latest"
+    registry, repository = resolve_repository_name(registry_repository)
+    return registry, repository, tag
```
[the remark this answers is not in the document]
bartvanb: I've adapted the function to remove the '4.5' tag in such cases and return the sha256. I've also updated the docstring and added the unittest. 

Obviously we have to choose what to return, tag or sha256 (or both). I now made the choice to return sha but if anyone feels strongly about it we can of course change that.

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -1,21 +1,21 @@
-from datetime import datetime
-import logging
 import re
-import docker
-import requests
-import base64
 import json
+import hashlib
+from http import HTTPStatus
```
frankcorneliusmartin, line 4: This from import should move lower to the `from` imports block. Isnt that Black?

frankcorneliusmartin (changes requested):

frankcorneliusmartin (changes requested):
Some final remarks

vantage6-common/vantage6/common/docker/addons.py
bartvanb: There's a pylint rule that imports from the standard library have to come before imports from other dependencies. So in order to not violate that I'll leave this here.

vantage6/vantage6/cli/configuration_wizard.py
bartvanb: I guess you're right, well spotted

vantage6-node/vantage6/node/docker/docker_manager.py
bartvanb: You're right, not expecting a major performance boost but who knows how node configurations may look in the future

vantage6-node/vantage6/node/docker/docker_manager.py
bartvanb: `get_digest()` has now been modified to return `None` if digest is not returned. Is this OK now?

vantage6-node/vantage6/node/docker/docker_manager.py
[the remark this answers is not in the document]
bartvanb: I created this issue just now.

vantage6-node/vantage6/node/docker/docker_manager.py
frankcorneliusmartin: I like it. Not because of performance but for clarity

vantage6-node/vantage6/node/docker/docker_manager.py
frankcorneliusmartin: Yes thats clearer

frankcorneliusmartin (approved):

lsago (commented):
I like the idea of comparing digest (allowed) to digest (requested) in the images, seems like the most solid approach.

I leave something to probably consider before release.

vantage6-node/vantage6/node/docker/docker_manager.py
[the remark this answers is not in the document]
bartvanb: So *I think* we could resolve this by having `parse_repository_tag` return the actual tag rather than the sha256 as 'tag'. A tag is always pointed to 1 image hash, so that wouldn't be an issue. Then `parse_repository_tag` should do the following

```
parse_repository_tag("somename:sometag@sha256:somehash") == ("somename", "sometag")
```

If it does, that would work with the following policies:

- allowed image `somename:sometag`: tags are the same so same image
- allowed image `somename@sha256:somehash`: the hash for `sometag` would be retrieved and if that is the same as `somehash` it will be allowed
- allowed image `somename:sometag@sha256:somehash`: in this case, hash would also be retrieved and compared

So the choice I made previously to return the hash rather than the tag seems to be unfortunate - by returning the tag rather than the hash we could resolve these issues I think

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -341,13 +327,104 @@ def is_docker_image_allowed(self, docker_image_name: str, task_info: dict) -> bo
...
+                    "allow to start tasks."
                 )
                 return False
 
-        return True
+        algorithm_whitelisted = False
+        if allowed_algorithms:
+            if isinstance(allowed_algorithms, str):
+                allowed_algorithms = [allowed_algorithms]
+            evaluated_img_wo_tag, _ = parse_repository_tag(evaluated_img)
```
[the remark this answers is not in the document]
bartvanb: Good spot, I'll replace the use of the library function here with the improved function from the addons

PR #1330, opened 2024-06-14
Feature/visualize line

### LINKED ISSUE #1324: [Feature Request] Add option to visualize line charts
**Problem description**
Kaplan Meier output can be visualized as line

**Desired solution**
Adapt algorithm store and UI to visualize the line

bartvanb:
Fix #1324 

Note that this also contains commits from PR #1328 - that should be reviewed first


frankcorneliusmartin (changes requested):
The x axis does not seemed to be evenly spaced. It seems to just use the x values from the output in a sequence.

Issue #1333, opened 2024-06-14
[Bug] Radom BadNameSpaceError when starting a node


hcadavid:
**Describe the bug**
When starting a node (v6 node start), sometimes it crashes while trying to share its configuration with the server:

````
2024-06-14 12:32:58 - node           - DEBUG    - Sharing node configuration: {'encryption': True, 'allowed_algorithms': 'all', 'database_labels': ['default'], 'database_types': {'db_type_default': 'csv'}, 'database_columns': {'columns_default': ['Name', 'Age', 'Location']}}
Exception in thread Thread-6 (_handle_eio_message):
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.10/site-packages/socketio/client.py", line 705, in _handle_eio_message
    self._handle_event(pkt.namespace, pkt.id, pkt.data)
  File "/usr/local/lib/python3.10/site-packages/socketio/client.py", line 581, in _handle_event
    r = self._trigger_event(data[0], namespace, *data[1:])
  File "/usr/local/lib/python3.10/site-packages/socketio/client.py", line 636, in _trigger_event
    return self.namespace_handlers[namespace].trigger_event(
  File "/usr/local/lib/python3.10/site-packages/socketio/namespace.py", line 18, in trigger_event
    return getattr(self, handler_name)(*args)
  File "/vantage6/vantage6-node/vantage6/node/socket.py", line 46, in on_sync
    self.node_worker_ref.share_node_details()
  File "/vantage6/vantage6-node/vantage6/node/__init__.py", line 1096, in share_node_details
    self.socketIO.emit("node_info_update", config_to_share, namespace="/tasks")
  File "/usr/local/lib/python3.10/site-packages/socketio/client.py", line 393, in emit
    raise exceptions.BadNamespaceError(
socketio.exceptions.BadNamespaceError: /tasks is not a connected namespace.
2024-06-14 12:32:58 - socket         - INFO     - Node <collab-One - orgA> joined room <collaboration_1>
````
Despite the error, the node keeps running (in the UI it is shown as 'active'/green), and the problem is noticed only when you try to create a task (e.g., from the UI), as there are no databases to choose from.

[...]

PR #1334, opened 2024-06-14
Addressing random BadNameSpace error by including the 'always_connect' option to Socket.io

hcadavid:
This is a minor fix that seems to address issue #1333: random socket.io error (socketio.exceptions.BadNamespaceError: /tasks is not a connected namespace) on the v6 nodes while trying to share its parameters with the server during its initialization. It is based on this discussion thread ([Client is getting socketio.exceptions.BadNamespaceError after reconnecting · Issue #920 · miguelgrinberg/python-socketio (github.com)](https://github.com/miguelgrinberg/python-socketio/issues/920)). I tested this with a server running with the 'patched' code within a SURF-RC workspace, and a node running on a separate one (also on SURF). When I removed the 'always_connect' option and restarted the server and the node, the problem occurred again.

The [official documentation](https://python-socketio.readthedocs.io/en/latest/api.html) states:

___**always_connect**... Set to True if you need to emit events from the connect handler and your client is confused when it receives events before the connection acceptance. In any other case use the default of False.___

For me, issue #1333 fits the 'confusion' part.

### DIFF
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -219,6 +219,7 @@ class ServerApp:
                 cors_allowed_origins=cors_settings,
                 logger=debug_mode,
                 engineio_logger=debug_mode,
+                always_connect=True
             )
         except Exception as e:
             log.warning(
@@ -235,6 +236,7 @@ class ServerApp:
                 cors_allowed_origins=cors_settings,
                 logger=debug_mode,
                 engineio_logger=debug_mode,
+                always_connect=True
             )
 
         # FIXME: temporary fix to get socket object into the namespace class
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -219,7 +219,7 @@ class ServerApp:
                 cors_allowed_origins=cors_settings,
                 logger=debug_mode,
                 engineio_logger=debug_mode,
-                always_connect=True
+                always_connect=True,
             )
         except Exception as e:
             log.warning(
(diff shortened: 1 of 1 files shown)


bartvanb (approved):
I think you are absolutely correct in your analysis that this will solve the problem. In the server's `on_connect`, it emits the `sync` event to the node, which in turn causes the node to send a `node_info_update` event. When the `node_info_update` event occurs before the `on_connect` is finished, you get the error that you're facing - nice racing condition.

While I'm approving this as it will be a fix that works, I have the impression that using `always_connect` is not necessarily recommended - there must be a reason for having provisionary connections by default right? I'm thinking an even better solution would be to get rid of the `node_info_update` event entirely, and replace it with a POST `/node/config` API endpoint, so that we prevent the racing condition on the socket.

If you agree, we can still merge this now but let's then also create an issue to create that API endpoint and remove the `always_connect` option again - shouldn't be too much work

hcadavid:
> I think you are absolutely correct in your analysis that this will solve the problem. In the server's `on_connect`, it emits the `sync` event to the node, which in turn causes the node to send a `node_info_update` event. When the `node_info_update` event occurs before the `on_connect` is finished, you get the error that you're facing - nice racing condition.
> 
> While I'm approving this as it will be a fix that works, I have the impression that using `always_connect` is not necessarily recommended - there must be a reason for having provisionary connections by default right? I'm thinking an even better solution would be to get rid of the `node_info_update` event entirely, and replace it with a POST `/node/config` API endpoint, so that we prevent the racing condition on the socket.
> 
> If you agree, we can still merge this now but let's then also create an issue to create that API endpoint and remove the `always_connect` option again - shouldn't be too much work

@bartvanb I agree, there must be a reason for the Socket.io documentation to recommend not using this setting unless necessary. I didn't see any side-effects of this change, but to be on the safe side it'd indeed be a good idea, after merging this fix, to create an issue for replacing this event-based 'node-info-update' with a simpler, good-old API call :).

bartvanb:
I've merged this in 4.6 to ensure we'll test it properly before releasing it

PR #1339, opened 2024-06-19
1180 change request compatibility with eduvpn v3

dsmits:
I'm merging into 4.6 but maybe it needs to be something different. If possible I would like to use this as soon as possible.

The change that I made will break compatibility with eduvpn v2 but hopefully nobody cares.

### DIFF
--- a/vantage6-node/vantage6/node/__init__.py
+++ b/vantage6-node/vantage6/node/__init__.py
@@ -842,10 +842,6 @@ class Node:
         if connect_mode == VPNConnectMode.FIRST_TRY:
             self.log.debug("Using existing config file to connect to VPN")
             next_mode = VPNConnectMode.REFRESH_COMPLETE
-        # elif connect_mode == VPNConnectMode.REFRESH_KEYPAIR:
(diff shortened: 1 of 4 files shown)


vantage6-node/vantage6/node/__init__.py
```diff
@@ -841,11 +841,11 @@ def _connect_vpn(
         do_try = True
         if connect_mode == VPNConnectMode.FIRST_TRY:
             self.log.debug("Using existing config file to connect to VPN")
-            next_mode = VPNConnectMode.REFRESH_KEYPAIR
-        elif connect_mode == VPNConnectMode.REFRESH_KEYPAIR:
-            self.log.debug("Refreshing VPN keypair...")
-            do_try = self.client.refresh_vpn_keypair(ovpn_file=ovpn_file)
             next_mode = VPNConnectMode.REFRESH_COMPLETE
+        # elif connect_mode == VPNConnectMode.REFRESH_KEYPAIR:
```
dsmits, line 845: I removed the refreshing of the keypair because I don't see that functionality in v3

PR #1344, opened 2024-06-20
Feature/policy require pull algorithm image

### LINKED ISSUE #1200: [Feature Request] Give node admin option to not fallback on local image
**Problem description**
(Low prio). A node admin might want to make sure that algorithm images are always at the latest version. And that the local image is not use in lieu of being able to reach it from the registry.

**Desired solution**
A node config option to control that.

**Additional context**
See this thread for context: https://github.com/vantage6/vantage6/pull/1196#discussion_r1559498994

bartvanb:
Fix #1200


vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -526,6 +526,7 @@ def run(
             alpine_image=self.alpine_image,
             proxy=self.proxy,
             device_requests=self.algorithm_device_requests,
+            requires_pull=self._policies.get("require_algorithm_pull", False),
```
frankcorneliusmartin, line 529: Shouldn't we put this into a global?

frankcorneliusmartin (approved):
OK

vantage6-node/vantage6/node/docker/docker_manager.py
bartvanb: You mean the name of the policy? Yeah maybe we should - and we should also do it for the other policy names then.

frankcorneliusmartin: That is also a good idea, but I really ment the `False` value which is a default for this setting.

bartvanb: Ah right. Well now we have both :-)

PR #1387, opened 2024-07-22
Address review allowed images: Return tag instead of sha256 when pars…

bartvanb:
…ing image, improve getting image name without tag, and improving error handling


lsago (commented):
It looks to me like this would solve the issue raised in the other PR.


IMHO, ideally, we'd do the following:
* Break `is_docker_image_allowed()` into multiple functions. So it's easier to maintain/read and so we we can write tests for them easily. I personally wouldn't vote to unit-test all the things, especially when this requires extensive mocking. But for functions like these that are more critical and closer to pure functions, I think it might be a good idea.
* Avoid being in the business of parsing ourselves docker image names, although this does seem hard if we want to provide the flexibility to allow any image of the same repo without regex.

But perhaps it's best to tackle that refactor in a different future PR?

PR #1399, opened 2024-07-24
Feature/add UI and store to vdev

### LINKED ISSUE #1078: [Feature Request] Add algorithm store and UI to `v6 dev` network
**Problem description**
The dev network would be easier to use if an algorithm store and UI were also a part of it

**Desired solution**
- UI can be added to server configuration and it will automatically be started
- Perhaps the community algorithm store can be linked to the dev network by default, otherwise maybe create a simple algorithm store with an average algorithm or so in it

bartvanb:
Fix #1078


vantage6/vantage6/cli/algostore/remove.py
```diff
@@ -0,0 +1,43 @@
...
+        ).ask():
+            info("Algorithm store will not be deleted")
+            exit(0)
+
+    # now remove the folders...
+    remove_file(ctx.config_file, "configuration")
+
+    for handler in itertools.chain(ctx.log.handlers, ctx.log.root.handlers):
+        handler.close()
+    remove_file(ctx.log_file, "log")
```
frankcorneliusmartin, line 43: I think this is a rotating log? Then you should also delete al `log.n` files?

vantage6/vantage6/cli/dev/start.py
```diff
@@ -40,13 +51,61 @@ def start_demo_network(
...
+            force=True,  # required to link localhost store
+        )
+        # note that we do not need to register the user as root of the store: this is
+        # already handled in the store config file and is executed on store startup (and
+        # successful because server is already started up at that point)
+    info("Done!")
+
+    # link the community store also to the server
+    info("Linking community algorithm store to local server...")
+    COMMUNITY_STORE = "https://store.cotopaxi.vantage6.ai"
```
frankcorneliusmartin, line 103: Can we store this somewhere where its easier to maintain?

frankcorneliusmartin (approved):

vantage6/vantage6/cli/algostore/remove.py
bartvanb: I am now just removing the entire directory, and doing the same for `v6 server remove`. This also has the advantage that any other files are also deleted (e.g. for the server RabbitMQ configuration files were not deleted)

vantage6/vantage6/cli/dev/start.py
bartvanb: Definitely, moved it to the globals

PR #1401, opened 2024-07-24
Don't require user to run client.setup_encryption(), unless they have…

### LINKED ISSUE #1302: [Change Request] Dont require setup_encryption in python client
**Description**
Whenever initializing the Python client, we always need to do 

```
client.setup_encryption(None)
# or setup with a private key if actually using encryption
```

This is inconvenient for users, who get an unclear error message if they forgot to do this or didn't know about it

**Proposed solution**

- No longer require it for unencrypted collaboration, but fix in the initialization
- For encrypted collabs, clear error message if it has been forgotten

bartvanb:
… an encrypted collaboration

I also checked if we needed to change error messages in the client in case encryption was not setup when using an encrypted collaboration. This was not the case: encryption setup is only required when creating a task or reading the results. When creating a task, the server will return a clear message that input is not encrypted. When reading results, it gives a clear message that encryption is not setup.

Fix #1302

### DIFF
--- a/vantage6-client/vantage6/client/__init__.py
+++ b/vantage6-client/vantage6/client/__init__.py
@@ -14,7 +14,7 @@ import traceback
 from pathlib import Path
 
 from vantage6.common.globals import APPNAME
-from vantage6.common.encryption import RSACryptor
+from vantage6.common.encryption import DummyCryptor, RSACryptor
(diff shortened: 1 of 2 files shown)


frankcorneliusmartin (approved):
Thanks. This kind-a should have been there from the beginning.

Issue #1218, opened 2024-04-22
[Bug] `allowed_organizations` not working


lsago:
**Describe the bug**
`allowed_organizations` policies seems to not be working

**To Reproduce**
* Set up a node with:
  ```yaml
  policies:
    allowed_organizations:
      - 7
  ```
* Send a task
* Error in node:
  ```
  2024-04-22 20:45:06 - common         - DEBUG    - Making request: GET | https://v6server.example.com:443/api/organization | {'name': 7}
  2024-04-22 20:45:06 - node           - ERROR    - 'str' object has no attribute 'get'
  Traceback (most recent call last):
    File "/vantage6/vantage6-node/vantage6/node/__init__.py", line 989, in run_forever
      self.__start_task(taskresult)
    File "/vantage6/vantage6-node/vantage6/node/__init__.py", line 349, in __start_task
      task_status, vpn_ports = self.__docker.run(
    File "/vantage6/vantage6-node/vantage6/node/docker/docker_manager.py", line 503, in run
      if not self.is_docker_image_allowed(image, task_info):
    File "/vantage6/vantage6-node/vantage6/node/docker/docker_manager.py", line 336, in is_docker_image_allowed
      is_allowed = self.client.check_user_allowed_to_send_task(
    File "/vantage6/vantage6-common/vantage6/common/client/node_client.py", line 361, in check_user_allowed_to_send_task
      if org.get("name") == allowed_org and org.get("id") == init_org_id:
  AttributeError: 'str' object has no attribute 'get'
  ```

**Expected behavior**
Allowing or disallowing the algorithm correctly

**Details**
 - OS: Linux
 - Version: 4.4.0

PR #1407, opened 2024-07-29
Draft: refactor node policy checks

### LINKED ISSUE #1406: [Change Request] Node policy checks refactor
**Description**
Current [`is_docker_image_allowed()` is a bit lengthy](https://github.com/vantage6/vantage6/blob/3b38ac1e738a95cda1d78d90cc34f4f1190e9cdb/vantage6-node/vantage6/node/docker/docker_manager.py#L288-L448). If we were to break this function further we could write tests for it more easily. Since it's a somewhat critical function controlling execution of algorithms, making sure it's well tested and clear would be desirable.

When trying to fix #1218, a good opportunity for a refactor comes up.

lsago:
This addresses #1218 and #1406

In a draft state. If we agree this is a sensible approach for the refactor and settle how far it should go (TODOs), I will tackle those TODOs, review the code with correctness in mind and test it. After I will ask for a regular review.


vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -285,164 +286,374 @@ def create_volume(self, volume_name: str) -> None:
...
+            True if policy allows for execution, False otherwise
         """
-        Checks the docker image name.
+        allowed_stores = self._policies.get(NodePolicy.ALLOWED_ALGORITHM_STORES, False)
+
+        if not allowed_stores:
+            self.log.warn("No allowed stores policy set. All stores are allowed!")
+            return True
+
+        # TODO: What is store ID? Is it an integer? if so int 0 would be False
```
bartvanb, line 532: Yes it's an int but like all other IDs it should be positive (which we should check)

vantage6/vantage6/cli/configuration_manager.py
```diff
@@ -51,6 +55,16 @@ class NodeConfiguration(Configuration):
...
+        Optional("policies"): {
+            Optional("allowed_organizations"): validate_ids_and_names,
+            Optional("allowed_users"): And(
+                validate_ids_and_names,
+                Use(
+                    lambda x: "names" not in x,
+                    error="`allowed_users` does not support `names` key at the moment.",
+                ),
+            ),
+        },
```
bartvanb, line 67: Am I correct in reading this that you require the configuration to be like
```
policies:
    allowed_organizations:
       names: [ x, y, z]
       ids: [1, 2, 3]
```?
If yes, then that would not be backwards compatible, as we currently just provide a list of ids/names. We should then figure out how to handle that - however as it's not used a lot we may be getting away with shortcuts.

**Review by bartvanb — COMMENTED** [review]
I agree that this is the sensible approach, splitting this function makes it much more readable.

One thing I wanted to mention was that #545 is included in the current sprint which will focus on validation of the all configuration options (which we have not really been using recently). Let's communicate on proceeding with this to prevent major merge conflicts - I can also wait with implementing that as there's no rush.

File: vantage6/vantage6/cli/configuration_manager.py (hunk 2 above)
Comment (lsago) [reply vantage6/vantage6/cli/configuration_manager.py:67]: Correct! Or something like:
```yaml
    allowed_organizations:
       names:
          - x
          - y
          - z
       ids:
          - 1
          - 2
          - 3
```
Should have already included an example in documentation for you.. my bad.

I did this because it might not always be trivial to distinguish between if they mean 'ID' or mean a 'name'. Super corner case: an org is named `1_000`. I think if the node admin writes 1_000 in their yaml, it will be interpreted as 1000, hence as ID. Things like that worried me a bit that somebody might be able to fiddle with it enough to circumvent somebody's policy in some scenario... Hence I opted for explicitness.

As for backwards compatibility, I agree, although in this case IIRC, `allowed_organizations` was simply not working.

bartvanb: > As for backwards compatibility, I agree, although in this case IIRC, allowed_organizations was simply not working.

Ah yes, in that case it's fair to ignore backwards compatibility

frankcorneliusmartin:
Closing this now as its been not active for almost a year. Feel free to re-open if you want to work on this.

Issue #241, opened 2022-08-03
[Feature Request] Delete organizations from server


frankcorneliusmartin:
**Problem description**
Organizations cannot be deleted at present. If they are deleted, all related resources (users, roles, tasks, etc) should also be deleted

**Desired solution**
Add possibility to the API to delete an organization

frankcorneliusmartin:
* Should also be possible through the Python client 
* User is prompted with a summary of all to be deleted resources
* UI should be updated

options
(1) endpoint does not delete if certain param is missing
(2) do not prompt in the Python client, but let the UI do some additional calls before actually deleting it

bartvanb:
I want this option for setting up workshop resources so I'm making it now

PR #1417, opened 2024-08-01
Implemented functionality to delete organizations (API, client and UI)

### LINKED ISSUE #1120: [Feature Request] Delete organisation from UI
**Problem description**
I can't delete an organisation from the ui either from the organisation itself or root.

**Desired solution**
An organisation delete button.

bartvanb:
Also added option to filter by multiple IDs on GET /organization endpoint, which is convenient in the UI. And fixed which organizations are visible in UI when creating roles/users

Fix #1120, #241

### DIFF
--- a/vantage6-server/vantage6/server/resource/organization.py
+++ b/vantage6-server/vantage6/server/resource/organization.py
@@ -49,7 +49,7 @@ def setup(api: Api, api_base: str, services: dict) -> None:
         Organization,
         path + "/<int:id>",
         endpoint="organization_with_id",
-        methods=("GET", "PATCH"),
+        methods=("GET", "PATCH", "DELETE"),
         resource_class_kwargs=services,
     )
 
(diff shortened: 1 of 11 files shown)


vantage6-server/vantage6/server/resource/organization.py
```diff
@@ -468,3 +481,134 @@ def patch(self, id):
...
+                    log.warning(
+                        "Deleting %s roles of organization %s",
+                        len(organization.roles),
+                        organization.name,
+                    )
+                    for role in organization.roles:
+                        role.delete()
+
+            else:
+                return {
```
frankcorneliusmartin, line 599: Make this return statement a guard so that the other block can move one level up and make it more readable.

bartvanb: good idea

Issue #1221, opened 2024-04-24
[Change Request] Rename headers in the UI of the Task page


frankcorneliusmartin:
**Description**
Currently the *central* task has the header **status** at all and the one for the children says **subtasks**. Ideally we would be able to annotate **Central** and **Partial**. But I do not think we have the role of the container at this stage. 

For the short term I think it would be good to have to following structure:

# Status
## Root
[ ] [ ]
## Children
[ ] [ ] 

Alternatively. I am not sure if the UI is able to visualize more than 2 levels? For now there is no use case for more levels. We could kind of assume that if there is one root node and there are children we are dealing with a central (orchastrator/aggregator) and Partial containers?

PR #1419, opened 2024-08-02
Change/UI/clarify naming task status

### LINKED ISSUE #1416: [Change Request] Show task name in task progress page
Can the task buttons also show the task name so that I can distinguish between them? The collaboration ID can be omitted as it is already on top of the page

image

bartvanb:
Fix #1416, #1221 

Now it looks like this:
image

@frankcorneliusmartin @harmbuisman is this what you want? I requested your review mainly to ask this - you don't need to look at the code (though of course you can if you like)


frankcorneliusmartin:
Ok by me. Although I am a little worried that names can be very long... Can we add something like a max char limit + eclipse?

Issue #1467, opened 2024-09-10
[Bug] `v6 node new` in combintation with `allowed_algorithms` results in corrupt configuration


frankcorneliusmartin:
**Describe the bug**
When using `v6 node new` and you specify the allowed algorithms you get an additional line in the configuration file. This invalidates the YAML file.

```yaml
policies:
  ? !!python/object/apply:vantage6.common.globals.NodePolicy
  - allowed_algorithms
  : - ^harbor2\.vantage6\.ai/demo/average$
```

Issue #1457, opened 2024-08-27
[Bug] `v6 node new` fails if server URL has trailing slash


lsago:
**Describe the bug**
`v6 node new` fails if server URL has trailing slash
```
$ v6 node new
....
? The base-URL of the server: https://v6server.example.com/
? Enter port to which the server listens: 443
? Path of the api: /api
? Task directory path: ...
...
...
Server not found at https://v6server.example.com/:443/api/token/node
[error] - Could not authenticate with server: Invalid token type. Token must be a <class 'bytes'>
[error] - Please check (1) your API key and (2) if your server is online
[warn ] - If you continue, you should provide your collaboration settings manually.
? Do you want to abort? Yes
```

**Expected behavior**
Being able to handle that, as it's a bit annoying for the user a fail after having provided all the options.

**Details**
version 4.6.1

lsago:
While we are at it, defaulting to port 443 if server urls starts with `https://` would be nice :)

PR #1468, opened 2024-09-10
Bugfixes for `v6 node new`

frankcorneliusmartin:
Fix for #1457 and #1467 

- [x] tested with the integration tests (needed to update the diagnostics runner + minor change in the code here)

### DIFF
--- a/vantage6/vantage6/cli/globals.py
+++ b/vantage6/vantage6/cli/globals.py
@@ -62,4 +62,3 @@ class AlgoStoreGlobals(str, Enum):
 
     DB_URI_ENV_VAR = "VANTAGE6_ALGO_STORE_DB_URI"
     CONFIG_NAME_ENV_VAR = "VANTAGE6_ALGO_STORE_CONFIG_NAME"
-    PORT = "7602"  # 7600 is the default UI port


dsmits (approved):
Seems fine to me. Only I don't entirely understand the shape of the old yaml vs. the newly fixed one.

vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -40,11 +40,21 @@ def node_configuration_questionaire(dirs: dict, instance_name: str) -> dict:
                 "message": "The base-URL of the server:",
                 "default": "http://localhost",
             },
+        ]
+    )
+    # remove trailing slash from server_url if entered by user
+    config["server_url"] = config["server_url"].rstrip("/")
+
+    # set default port to 443 if server_url is https
+    default_port = "443" if config["server_url"].startswith("https") else "5000"
```
bartvanb, line 49: Can we use constants for 5000 and 443? I bet we use them in more places

frankcorneliusmartin: I changed it everywhere. Also I changed `5000` to `7601` (as this was changed in another PR)

frankcorneliusmartin: Done

vantage6-server/vantage6/server/_data/dev/node_a.yaml
```diff
@@ -1,11 +1,11 @@
 application:
   api_key: demo-key-node2
   server_url: http://localhost
-  port: 5000
+  port: 7601
   api_path: '/api'
   task_dir: tasks
   databases:
-    default: 
+    default:
```
bartvanb, line 9: this looks outdated - but seems to have been outdated before. Maybe check if this is used anywhere and otherwise remove it?

bartvanb: Same for file(s) below

bartvanb: Or maybe better to create an issue, then we don't accidentally delete wrong files on the version we'll use in the workshop

vantage6/vantage6/cli/globals.py
```diff
@@ -58,12 +55,11 @@ class ServerGlobals(str, Enum):
...
     CONFIG_NAME_ENV_VAR = "VANTAGE6_CONFIG_NAME"
-    PORT = "5000"
 
 
 class AlgoStoreGlobals(str, Enum):
     """Enum containing algorithm store environment variables"""
 
     DB_URI_ENV_VAR = "VANTAGE6_ALGO_STORE_DB_URI"
     CONFIG_NAME_ENV_VAR = "VANTAGE6_ALGO_STORE_CONFIG_NAME"
-    PORT = "5002"  # 5001 is the default UI port
```
bartvanb, line 69: I think here we can use `Ports.DEV_ALGO_STORE` right? If not, might be better just to remove this enum property and then use the other enum where this is used.

frankcorneliusmartin: That is exactly what I've done?

vantage6-server/vantage6/server/_data/dev/node_a.yaml
frankcorneliusmartin: I created a separate issue for this as you suggested.

vantage6/vantage6/cli/globals.py
frankcorneliusmartin: Thanks, good catch!

PR #1487, opened 2024-09-17
Added logic to delete studies when collaboration is removed

### LINKED ISSUE #1465: [Bug] Studies are not deleted when collaboration is deleted
**Describe the bug**
The log message is correct but the implementation bellow is missing the logic to do so:

https://github.com/vantage6/vantage6/blob/4dd5ace63d49b2d1ab6a332ffc81533159cff1ec/vantage6-server/vantage6/server/resource/collaboration.py#L636-L651

frankcorneliusmartin:
Fixes #1465


bartvanb (approved):
Nice addition. I updated the target branch to 4.7 branch where other changes for workshop are already targeted - feel free to change back if you prefer.

PR #1490, opened 2024-09-19
Small changes to running tasks in the UI - and change to default roles

### LINKED ISSUE #1445: [Feature Request] Keep Task/Node overview for finished tasks
During execution I get a nice overview of processes and nodes on which which task is executed. This is lost after the main task is finished. Can this also be available for finished tasks? 

This overview:
image

### LINKED ISSUE #1472: [Feature Request] Always add show logs option on task
Now it only shows logs when a task failed.
Also, if possible, allow show logs for tasks that are completed but for which the main task is not yet done.

E.g. in the image below, also show logs for the green cells:
image

### LINKED ISSUE #1478: [Bug] UI doesn't refresh the task status on the UI automatically
When starting a new task on Vantage 4.7.0, the UI shows the task's initial state: PENDING. However, the UI doesn't automatically refresh the status when it changes (e.g. to ACTIVE). Knowing the task status then requires reopening the task lists and selecting the task again to see its actual state. E.g., after launching the task the UI shows it as PENDING indefinitely, despite being already completed:

image

bartvanb:
Server:
- add rule 'view events on collaboration scope' to all default roles. Without this, they cannot receive socket events such as algorithm updates and node statuses for other nodes in their collaboration, which is very inconvenient and leads e.g. to tasks hanging in 'pending' in the UI (fix #1478)

UI:
- extra check to prevent racing condition to prevent that task status updates get stuck on read task page (other part fix #1478)
- fix for repeating task with boolean values: set them by default to false
- Prevent that users can submit the task multiple times on create task page if it doesn't respond quick enough
- keep task status (folded) when task is finished, fix #1445
- also show log file on successfully completed algorithm runs, fix #1472

### DIFF
--- a/vantage6-server/vantage6/server/default_roles.py
+++ b/vantage6-server/vantage6/server/default_roles.py
@@ -65,6 +65,7 @@ def get_default_roles(db) -> list[dict]:
         db.Rule.get_by_("run", Scope.COLLABORATION, Operation.VIEW),
         db.Rule.get_by_("port", Scope.ORGANIZATION, Operation.VIEW),
         db.Rule.get_by_("event", Scope.ORGANIZATION, Operation.RECEIVE),
+        db.Rule.get_by_("event", Scope.COLLABORATION, Operation.RECEIVE),
         db.Rule.get_by_("study", Scope.ORGANIZATION, Operation.VIEW),
     ]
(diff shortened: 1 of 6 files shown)


frankcorneliusmartin (approved):
Great that you fixed this so quickly!

PR #1491, opened 2024-09-20
Fix add already whitelisted server to store

### LINKED ISSUE #1460: [Bug] [unconfirmed]: Algo store does not allow server to add configured server in algo store config
**Describe the bug**
After starting an algorithm store instance with root user and policies defined, I can't create the algorithms store at the server.

**To Reproduce**
Given a config like this:
```yaml
uri: "postgresql://<correct_details_here>/somedatabase"
allow_drop_all: False

logging:
  level: DEBUG
  file: "v6algostore.log"
  use_console: True
  backup_count: 5
  max_size: 1024
  format: "%(asctime)s - %(name)-14s - %(levelname)-8s - %(message)s"
  datefmt: "%Y-%m-%d %H:%M:%S"
  loggers:
    - name: urllib3
      level: warning
    - name: sqlalchemy.engine
      level: warning

debug:
  flask: false

policies:
  algorithm_view: whitelisted
  allowed_servers:
    - "https://someserver.example.com/api"
  allow_localhost: false

root_user:
  v6_server_uri: "https://someserver.example.com/api"
  username: "someusername"
```
Then I create the empty database (e.g. `somedatabase`). And then I start the container.

I notice the [table vantage6server](https://github.com/vantage6/vantage6/blob/version/4.7.0/vantage6-algorithm-store/vantage6/algorithm/store/model/vantage6_server.py) populated.
```
$ docker exec -u postgres -it vantage6-database-container psql -d somedatabase -c 'SELECT * FROM "vantage6server"'
 id |                      url
----+-----------------------------------------------
  1 | https://someserver.example.com/api
```

Now if I go to the UI connected to my vantage6 server and I try to add this store 
image
I get the following, an the store is not created at the server.
image

I haven't checked thoroughly, but I guess behind the scenes, when I hit submit ("Add algorithm store to your collaboration"), the server will POST to the algo store and [hit this code path](https://github.com/vantage6/vantage6/blob/c84b613228c866cbb362968b0b90710ffcf75a38/vantage6-algorithm-store/vantage6/algorithm/store/resource/vantage6_server.py#L223-L233). And because the server already exists in that table, it returns that 403.

 
**Expected behavior**
Able to create algostore in server.

**Additional context**
We (Bart and I) tried this and worked at some point, but perhaps it was because it was 4.6.1 and https://github.com/vantage6/vantage6/pull/1393 hadn't been merged yet? Other possibility is that I missed something – I'm not familiar with algo store at all.

Fix for now: I have manually edited the algostore database. First I change the url of the server in table `vantage6server`, then I add the algo store in the server/UI, then I go back to the database and change the permissions with something like `UPDATE "Permission" SET role_id = 1 WHERE user_id = 2;` – not only is this hacky bad, but also not the quickest way of doing it. But documenting so I/we know what I did later on. :)

**Details**
- **4.7.0**

wbaccinelli:
Fixes #1460 

When adding a store to the server, the store tried to whitelist the server. In case the server was already whitelisted, an error was returned and the store could not be added. 
The case in which an already whitelisted server tries to be whitelisted again returns a different response. In this way, the server is not added twice to the db, but the process is not blocked.

### DIFF
--- a/vantage6-algorithm-store/tests_store/resources/test_v6server.py
+++ b/vantage6-algorithm-store/tests_store/resources/test_v6server.py
@@ -135,7 +135,7 @@ class TestVantage6ServerResource(TestResources):
 
         # if we try this again it should fail
         response = self.app.post("/api/vantage6-server", headers=HEADERS, json=body_)
-        self.assertEqual(response.status_code, HTTPStatus.FORBIDDEN)
+        self.assertEqual(response.status_code, HTTPStatus.ALREADY_REPORTED)
 
         # check that whitelisting localhost servers cannot be done with default policy
         policy.delete()
--- a/vantage6-algorithm-store/vantage6/algorithm/store/resource/vantage6_server.py
+++ b/vantage6-algorithm-store/vantage6/algorithm/store/resource/vantage6_server.py
@@ -230,7 +230,7 @@ class Vantage6Servers(AlgorithmStoreResources):
             return {
                 "msg": "This server is already whitelisted in this algorithm store "
                 "instance."
-            }, HTTPStatus.FORBIDDEN
+            }, HTTPStatus.ALREADY_REPORTED
 
         # the user that is whitelisting the server should be able to delete it
         # in the future. Assign the server manager role to the user executing this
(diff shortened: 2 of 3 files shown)


frankcorneliusmartin (approved):
Never new this existed!

PR #1504, opened 2024-09-27
Add additional filter for collaboration when study is used

frankcorneliusmartin:
The filter for studies selected all organizations from that study and displayed **all** nodes from these organizations. I added a filter to only select those that are also part of the collaboration.

I tested it in a local environment.


bartvanb:
Closing this because fixed in #1507

PR #1506, opened 2024-09-27
When adding policies to database, check if double policies are presen…

bartvanb:
…t (which can happen when multiple store instances are restarting simultaneously

Also a bugfix for adding deprecated policies


frankcorneliusmartin (approved):
Ok

PR #1507, opened 2024-09-30
Change/show nodes on study page

bartvanb:
Also includes a fix on the GET /node endpoint

### DIFF
--- a/vantage6-server/vantage6/server/resource/node.py
+++ b/vantage6-server/vantage6/server/resource/node.py
@@ -270,7 +270,10 @@ class Nodes(NodeBase):
                 q.join(db.Organization)
                 .join(db.StudyMember)
                 .join(db.Study)
-                .filter(db.Study.id == study_id)
+                .filter(
+                    db.Study.id == study_id,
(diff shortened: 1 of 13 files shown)


vantage6-server/vantage6/server/resource/node.py
```diff
@@ -270,7 +270,10 @@ def get(self):
                 q.join(db.Organization)
                 .join(db.StudyMember)
                 .join(db.Study)
-                .filter(db.Study.id == study_id)
```
frankcorneliusmartin, line 273: This is already fixed in https://github.com/vantage6/vantage6/pull/1504/files. 

We can also close that one? Up to you. At least we had the same fix.

frankcorneliusmartin (approved):

vantage6-server/vantage6/server/resource/node.py
bartvanb: Haha nice that we solved the same issue. I'll merge this one as it also contains the UI stuff (also, not sure why you added an additional join on db.Collaboration there?)

PR #1522, opened 2024-10-08
Fix for selecting multiple columns from dropdown when creating tasks …

bartvanb:
…with column-list parameters

- Was broken since v4.4.1
- cause was improper initialization of angular forms as string input, then after loading columns the form input type was changed to a dropdown but Angular threw errors in that case

### DIFF
--- a/vantage6-server/vantage6/server/resource/role.py
+++ b/vantage6-server/vantage6/server/resource/role.py
@@ -459,7 +459,7 @@ class Roles(RoleBase):
         # verify that the organization for which we create a role exists
         if not db.Organization.get(organization_id):
             return {
-                "msg": f'organization "{organization_id}" does not ' "exist!"
+                "msg": f'organization "{organization_id}" does not exist!'
             }, HTTPStatus.NOT_FOUND
 
         # check if user is allowed to create this role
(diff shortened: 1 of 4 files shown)


frankcorneliusmartin (approved):
I dont get how this is solved it but the code looks sweet.

frankcorneliusmartin:
Sorry I had to merge the other issue in here to build a custom image for the workshop

bartvanb:
> I dont get how this is solved it but the code looks sweet. 

Just to try to explain this, for future reference because it also boggled my mind:
- In the old situation, the function `shouldShowParameterSimpleInput()` would return `True` when the page is initialized, because the columns are not available yet at that point
- When the user chooses the database, the return value of `shouldShowParameterSimpleInput` changes to `False`
- This change impacts the HTML form to fill out the multiple columns: it was initialized as a string input field (with possibility to add more) and changes to a multi-select dropdown.
- Angular Forms cannot properly deal with such changes after initialization. It generates a console error when the return value of `shouldShowParameterSimpleInput` is changed, then improperly changes the form itself, and finally, the form only registers the *first* column that is chosen from the dropdown.

Explaining this made me realize that this fix is not perfect: probably the issue will still occur if the column names are not successfully retrieved. Then `shouldShowParameterSimpleInput` will return `False` first, and after column names failed to have been retrieved it will return `True`. This will probably lead to other Angular Form errors - it might be better to **hold off on merging this for now**.

PR #1523, opened 2024-10-08
Bugfix/UI/proper display pd tojson tables

### LINKED ISSUE #1511: [Bug] User interface - faulty rendering with multiple line plots
**The bug**
When creating multiple line plots in the UI you can not switch between them without returning back to a different visual e.g. a table or refreshing the page. 

**To reproduce the bug**
1. Place an algorithm with multiple visuals in the algorithm store
_A good example thereof `https://github.com/MaastrichtU-CDS/v6-coxph`; this repository also contains suitable test data._
2. Run and complete the task via the UI
3. Switch between the 'Cumulative Baseline Hazard' and the 'Baseline Survival Function' line plots.
_The line plot now remains the same, while we expect the curve to look significantly different._
4. Switch to the table, then switch back to the graph again.
_You should now notice that the graph does change._

**Details**
User interface (UI) version 4.7.0
Server version 4.7.0
Tested in Mozilla Firefox

bartvanb:
This makes visualization of tables in the UI more stable, as it will also visualize tables in case you have a specifically named index of your pandas data frame in the Python algorithm.

Also, there is a small fix in resetting visualizations in the last commit, which probably fixes #1511 - that was probably an issue that task was still being loaded while user switched visualization, and then when task finishes loading, it changes back but does not reset the visualization dropdown.

Finally, notice that this PR also includes the commits of #1522 - so we can drop that one and review only this one, or do that one first

### DIFF
--- a/vantage6-ui/src/app/helpers/visualization.helper.ts
+++ b/vantage6-ui/src/app/helpers/visualization.helper.ts
@@ -1,17 +1,31 @@
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function parseDefaultPandasFormat(df: { [key: string]: any[] }, columns: string[] | null): { columns: string[]; rows: any } {
+export function parseDefaultPandasFormat(
+  df: { [key: string]: any[] },
+  parameterTxt: string,
+  columns: string[] | null
+): { columns: string[]; rows: any[] } {
   if (!columns) {
     columns = Object.keys(df);
   }
-  const rows = [];
-  for (let i = 0; i < Object.keys(Object.values(df)[0]).length; i++) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const indices = Object.keys(Object.values(df)[0]);
+  const rows: any = [];
+  // if the indices are not simply numbers, add them as first row
+  const indices_are_numbers: boolean = indices[0] !== '0';
+  if (indices_are_numbers) {
+    // add extra column for indices
+    columns = [parameterTxt, ...columns];
+  }
+  indices.forEach((index: any) => {
     const row: any = {};
     for (const column of columns) {
-      row[column] = df[column][i] as string;
+      if (column === parameterTxt) {
+        row[parameterTxt] = index;
+      } else {
+        row[column] = df[column][index] as string;
+      }
     }
     rows.push(row);
-  }
+  });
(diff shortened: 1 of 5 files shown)


vantage6-ui/src/app/helpers/visualization.helper.ts
```diff
@@ -1,17 +1,31 @@
...
   if (!columns) {
     columns = Object.keys(df);
   }
-  const rows = [];
-  for (let i = 0; i < Object.keys(Object.values(df)[0]).length; i++) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const indices = Object.keys(Object.values(df)[0]);
+  const rows: any = [];
+  // if the indices are not simply numbers, add them as first row
+  const indices_are_numbers: boolean = indices[0] !== '0';
```
rian-schobers, line 13: I don't think I quite understand what's happening here, why do you need to check if 'indices_are_numbers'?
Object.keys always returns a string array, right? Maybe a typeof check?

bartvanb: Ah right, this must be very mysterious code if you're not familiar with Pandas dataframes in Python.
The pandas dataframe is an object like a table that has an index, which by default is 0, 1, 2, ... However, you can also set the index names, to e.g. `alpha, bravo, charlie, ...`, whatever you like, as long as the index is unique and has the same length as the dataframe. 
Now, when you export the dataframe to JSON, it will use the indices as the keys to each row. The code in this component reads that JSON object and converts it back to a table. The assumption that I'm making in this code here, is that if the default index 0, 1, 2... was used, it's not really interesting to display 0, 1, 2,... as first column. However, some algorithms use an index like `sum, mean, stddev, ...` to give these measures for different variables. In that case, we *do* want to show sum, mean, stddev in the visualization.

What I'll do, is add part of the text I've written above to the code so that this will be clear(er) for people reading it in the future.
I don't have a strong opinion on an extra typecheck, but as you say I think it's a string array so the code would work as it currently is (right?)

rian-schobers: Thanks for the clarification, then it is approved. The extra type check is indeed not necessary since it always returns a string array.

rian-schobers (approved):

PR #1524, opened 2024-10-08
Change/flexible api path store

### LINKED ISSUE #1519: [Change Request] Make api path settable in the algorithm store
**Description**
The api path is fixed at `/api` in the algorithm store. However, if in e.g. an nginx config, it is mapped onto another route, the HATEAOS paths are incorrect and the UI doesn't work properly.

The solution is to add a configuration option to set this. 

Bonus: maybe we should here also limit the setting of the api path in the server as it is not relevant to most users.

bartvanb:
Fix #1519


vantage6/vantage6/cli/configuration_wizard.py
```diff
@@ -312,21 +309,6 @@ def _get_common_server_config(
         ]
     )
 
-    # TODO v5+ remove api_path. It complicates configuration
-    if include_api_path:
```
frankcorneliusmartin, line 316: So do I get it right that the only way of setting this option is to manually set it? (Without using the wizard)? We also do it in a similar way for the vantage6-server? Why is it too complex here?

vantage6-backend-common/vantage6/backend/common/globals.py
```diff
@@ -4,3 +4,6 @@
 
 # environment variable name for host URI
 HOST_URI_ENV = "HOST_URI_ENV_VAR"
+
+# default API path
+DEFAULT_API_PATH = "/api"
```
frankcorneliusmartin, line 9: I am thinking it might be better to make it a different default than the server? e.g. `/store`  or `/store-api`?

frankcorneliusmartin (changes requested):

vantage6/vantage6/cli/configuration_wizard.py
bartvanb: There are many configuration options, so I think we should only ask users to configure the most important ones when doing `v6 [component] new`. I think configuring the API path is not important for most users - that's why I didn't include it in the wizard.

bartvanb: That's btw also why the comment is there: in v5 I would personally prefer to also remove the option from the server for the interactive part.

vantage6-backend-common/vantage6/backend/common/globals.py
bartvanb: I thought it would be simpler to keep them the same unless you run into issues - the more simple the easier to remember. And I don't think people will run into issues with them being the same: so far the store's API path was fixed at `/api`. 
Do you think people would run into issues and if so when? I think they are mostly avoided by using different ports locally and different URLs / nginx configs when deploying them

frankcorneliusmartin: I was thinking of all the poor people having to deal with NGINX configurations. This would promote using subpaths over subdomains. 

Anyway I don't really mind leaving it like this either. Feel free to merge it as is.

vantage6/vantage6/cli/configuration_wizard.py
frankcorneliusmartin: Maybe we should remove the wizard completely for all server components.

frankcorneliusmartin (approved):

vantage6/vantage6/cli/configuration_wizard.py
bartvanb: Indeed something to consider

PR #1525, opened 2024-10-08
Add option to set UI image in v6 dev network

### LINKED ISSUE #1518: [Feature Request] Set UI image in `v6 dev create-demo-network`
**Problem description**
It would be nice to add an option `--ui-image` much like the options to set `--server-image` etc

bartvanb:
Fix #1518

### DIFF
--- a/vantage6/vantage6/cli/dev/create.py
+++ b/vantage6/vantage6/cli/dev/create.py
@@ -266,6 +266,7 @@ def create_vserver_config(
     port: int,
     server_url: str,
     extra_config_file: Path,
+    ui_image: str | None,
     ui_port: int,
     store_port: int,
 ) -> Path:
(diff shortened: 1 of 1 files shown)


vantage6/vantage6/cli/dev/create.py
```diff
@@ -299,6 +303,10 @@ def create_vserver_config(
     )
 
     extra_config = _read_extra_config_file(extra_config_file)
+    if ui_image is not None:
+        if extra_config:
+            extra_config += "\n"
+        extra_config += f"images:\n  ui: {ui_image}"
```
frankcorneliusmartin, line 309: This wil break if `extra_config_file` already has a `images` section, right?

frankcorneliusmartin (changes requested):

vantage6/vantage6/cli/dev/create.py
bartvanb: Absolutely correct. I chose to ignore that in favour of having time to solve other things, as I don't think people will often run into that issue, and because it was not completely trivial to fix (now the extra config is read as simple text, then we have to parse as YAML and check that, and then convert back to text).

But I agree we should either solve it here or create a separate issue to fix later. I tend to prefer creating separate issue, what do you prefer?

frankcorneliusmartin: I am fine with a seperate issue :)

PR #1528, opened 2024-10-10
Add server_url as parameter to create task. This argument should be p…

### LINKED ISSUE #1454: [Bug] The web-based UI is unable to submit a task when the 'server_url' is set through environment variables on the UI
The configuration of the `server_url` setting required by the web-UI is inconsistent between versions before 4.6 and the newer ones. Previously you could either set this through a SERVER_URL environment variable on the web UI, or by adding the 'server_url' entry on the server configuration file. Starting version 4.6, this variable is only set when given through the 'server_url' entry on the server config file (otherwise you get `Server URL is not set in the configuration nor in the request arguments. Please provide it as 'server_url' in the request` when trying to submit a task). So it seems that the SERVER_URL environment variable on the web UI is now ignored.

This will cause servers and UIs working properly with older versions of v6 to fail after upgrading them to newer versions.

bartvanb:
…assed on if a linked algorithm store is provided. The idea was that this parameter was passed along with it but that part was wrongly implemented

Now it is properly accepted by the server, automatically sent by the UI and optional in the client (with a clear error that should help user which was already there but could not be acted upon)

Fix #1454

### DIFF
--- a/vantage6-server/vantage6/server/resource/task.py
+++ b/vantage6-server/vantage6/server/resource/task.py
@@ -723,7 +723,10 @@ class Tasks(TaskBase):
                 # get the algorithm from the algorithm store
                 try:
                     image, digest = Tasks._get_image_and_hash_from_store(
-                        store, image, config
+                        store=store,
+                        image=image,
+                        config=config,
+                        server_url_from_request=data.get("server_url"),
                     )
                 except Exception as e:
                     log.exception("Error while getting image from store: %s", e)
@@ -984,7 +987,10 @@ class Tasks(TaskBase):
 
     @staticmethod
     def _get_image_and_hash_from_store(
-        store: db.AlgorithmStore, image: str, config: dict
+        store: db.AlgorithmStore,
+        image: str,
+        config: dict,
+        server_url_from_request: str | None = None,
     ) -> tuple[str, str]:
         """
         Determine the image and hash from the algorithm store.
@@ -997,6 +1003,8 @@ class Tasks(TaskBase):
             URL of the docker image to be used.
         config : dict
             Configuration dictionary.
+        server_url_from_request : str, optional
+            Server URL from the request, by default None
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


vantage6-server/vantage6/server/resource/task.py
```diff
@@ -723,7 +723,10 @@ def post_task(data: dict, socketio: SocketIO, rules: RuleCollection, config: dic
                 # get the algorithm from the algorithm store
                 try:
                     image, digest = Tasks._get_image_and_hash_from_store(
-                        store, image, config
+                        store=store,
+                        image=image,
+                        config=config,
+                        server_url_from_request=data.get("server_url"),
```
frankcorneliusmartin, line 729: Should this not only be used when the `server_url` setting is not provided in the server configuration file? This is quite sensitive for user input mistakes isnt it?

bartvanb: I think we can go both ways here. The reason it is set up like this now is:
- `server_url` is optional for all requests, unless the server admin has not provided it in the config file. I don't think many people will ever provide the argument via the Python client, but if they do, I think it should overwrite the config. Otherwise, if the server admin made a mistake, users can never override that.
- The UI passes on the `server_url` in all algorithm store requests. In my view, that url is more reliable than what a server admin put in the config file, because if the server URL set in the UI was wrong, the user would not even be able to log in to that UI.

frankcorneliusmartin: I see, in the UI this indeed makes more sense (admin config vs build-in mechanism). Just to be sure, this does not allow the user to create an alternative server and exploit this in any way?

If the answer is no, feel free to merge it.

frankcorneliusmartin (approved):

vantage6-server/vantage6/server/resource/task.py
bartvanb: No that's not possible: if this request is done at server A, and the user inputs a server_url of server B, the algorithm store will try to verify the token of server A at server B, which will fail and lead to an error and no action taken.
(And the reverse case of doing this at server B and putting in server A's URL will of course lead to the same error)

Issue #1172, opened 2024-03-25
[Feature Request] Enable directory mounts for local read/write


yannick-vinkesteijn:
**Problem description**
Currently Vantage6 only supports file/db mounts, however our train/inference pipelines often need to store multiple checkpoint and result files with content that can't leave the server. Because each node has their own type of after processing of inference results, we use csv files and can't use a more flexible setup like a database.

**Desired solution**
Enable directory mounts with read/write access to support complex inference/train pipelines and local checkpoint saves.

harmbuisman:
We have a similar request where we can use a mounted folder to write to the file system of the machine that runs a node (see bottom left). One use case for this is for example to make and store local predicitions using a federated prediction model. 

![image](https://github.com/vantage6/vantage6/assets/7702207/f21404da-b91e-4bf1-b496-b1df1c226bd7)

frankcorneliusmartin:
@yannick-vinkesteijn 

* You need a bilateral way to exchange data between the vantage6-node and the EPD?
* The files that are provided are always in CSV format?
* Which format is provided from the vantage6 node to the EPD?
* Does the hospitals upload the data using the FTP protocol or something similar? 
* Could it be possible to deliver the files through a HTTP service? (e.g. to a POST `/files` endpoint?)

Sorry for asking so many questions, but I want to understand the problem so we can provided an appropriate solution.

yannick-vinkesteijn:
@frankcorneliusmartin sorry my notifications did not work.
- You need a bilateral way to exchange data between the vantage6-node and the EPD?
    _Yes because we read the files from the local server and write output files back to the server_
- The files that are provided are always in CSV format?
    _Some hospitals also give us jsons or excel files but these are exceptions_
- Which format is provided from the vantage6 node to the EPD?
    _Currently we return csv files to the server that are then further processed by the hospital. It depends per hospital if they have an automated export to EPD pipeline or do this by hand._
- Does the hospitals upload the data using the FTP protocol or something similar?
    _It depends per hospital how they do this. Some use a shared disk that has some scheduler attached others move the file manually each time we generate output. Because of the difference in automation and setup we just create a csv and are not responsible for the export to EPD part of the pipeline._
- Could it be possible to deliver the files through a HTTP service? (e.g. to a POST /files endpoint?)
[...]

PR #1539, opened 2024-10-18
Added functionality to mount folders (as databases)

mellesies:
Folders can be used for permanent storage between runs (e.g. caching). This is particularly useful when using duckdb. 

Adding functionality to support read-only mounts should be trivial.

### DIFF
--- a/docs/node/node_config.yaml
+++ b/docs/node/node_config.yaml
@@ -90,6 +90,13 @@ databases:
       cdm_schema: public
       results_schema: results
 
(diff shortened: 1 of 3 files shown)


frankcorneliusmartin:
Does it fix Fixes #1172 completely? I think we should leave that issue open for a bit longer as we probably should add decorators in the algorithm tools AND because we need to port this feature to version 5 (k8s).

frankcorneliusmartin (approved):

PR #1552, opened 2024-10-22
Added enum for the `AuthStatus` and set the node by default on offline

### LINKED ISSUE #1461: [Bug] Node status is `None` at the server when node has never been online
**Describe the bug**
Should be `offline`. / `never-seen`

frankcorneliusmartin:
Fixes #1461

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
@@ -1,10 +1,11 @@
 import json
 import logging
-import pandas as pd
 
 from typing import Any
 from importlib import import_module
 from copy import deepcopy
+
+import pandas as pd
 
 from vantage6.common.globals import AuthStatus
 from vantage6.algorithm.tools.wrappers import load_data
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
@@ -1,11 +1,12 @@
 import json
 import logging
+import pandas as pd
+
 from typing import Any
 from importlib import import_module
 from copy import deepcopy
 
-import pandas as pd
-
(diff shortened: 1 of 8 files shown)


wbaccinelli (approved):
Looks good, the code is also cleaner with the Enum.

vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
```diff
@@ -1,11 +1,12 @@
 import json
 import logging
+import pandas as pd
```
bartvanb, line 3: Doesn't linting complain about putting this import here? As in https://github.com/PyCQA/flake8-import-order: "In general stdlib comes first, then 3rd party, then local packages, and that each group is individually alphabetized, however this depends on the style used. Flake8-Import-Order supports a number of [styles](https://github.com/PyCQA/flake8-import-order#styles) and is extensible allowing for [custom styles](https://github.com/PyCQA/flake8-import-order#extending-styles)."

vantage6-server/vantage6/server/websockets.py
```diff
@@ -343,8 +344,8 @@ def on_ping(self) -> None:
         ping and sets them as online.
         """
         auth = db.Authenticatable.get(session.auth_id)
-        auth.status = "online"
-        auth.last_seen = dt.datetime.utcnow()
+        auth.status = AuthStatus.ONLINE.value
+        auth.last_seen = dt.datetime.now(dt.timezone.utc)
```
bartvanb, line 348: Nice that you caught this as well!

bartvanb (approved):
Looks great! One small style comment

vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
frankcorneliusmartin: No my linter does not complain. But I keep forgetting we agreed on this. Ill solve it.

PR #1555, opened 2024-10-22
Bugfix/client-update-functions-require-all-fields-to-be-set

### LINKED ISSUE #1515: [Bug] `client.organization.update` expects all fields to be set.
**Describe the bug**
When you run `client.organization.update(id, name="blabla")` the server will reject it. Because the other fields have been supplied by `None`values.

I suggest we also check the other `.update` methods on this.

frankcorneliusmartin:
Fixes #1515

### DIFF
--- a/vantage6-client/vantage6/client/__init__.py
+++ b/vantage6-client/vantage6/client/__init__.py
@@ -920,10 +920,7 @@ class UserClient(ClientBase):
             dict
                 Containing the meta-data of the updated node
             """
-            data = {
-                "name": name,
-                "clear_ip": clear_ip
-            }
+            data = {"name": name, "clear_ip": clear_ip}
             data = self._clean_update_data(data)
             return self.parent.request(
                 f"node/{id_}",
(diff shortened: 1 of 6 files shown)


vantage6-client/vantage6/client/__init__.py
```diff
@@ -1602,10 +1607,16 @@ def update(
             dict
                 Containing the updated role data
             """
+            data = {
+                "name": name,
+                "description": description,
+                "rules": rules,
+            }
+            data = {k: v for k, v in data.items() if v is not None}
```
bartvanb, line 1615: Nitpicking: would it be an idea to put this line in a separate function `_clean_update_data()` (since you are using it in several places and it is quite trivial but might be easier to read that way)

bartvanb (approved):
One small suggestion which you should feel free to ignore if you disagree

vantage6-client/vantage6/client/__init__.py
frankcorneliusmartin: I agree, I was being lazy I guess. Ill pick this up before Thursday.

Issue #1489, opened 2024-09-19
[Feature Request] Show study in the task view


frankcorneliusmartin:
**Desired solution**
I would like to see the study in the task (detail) view. Maybe also in the list view?

PR #1561, opened 2024-10-28
#1361 fixed task table filter and added study to list and read

### LINKED ISSUE #1361: [Bug] Filters in the UI are cleared when going to next page of table
**Describe the bug**
When going to a different page in the table, the search filters are cleared

**To Reproduce**
1. Go to table of tasks (Analyze -> Task)
2. add a name in the filter that has >10 matches
3. Go to next page
4. Name filter is removed

**Expected behavior**
Filters should be preserved when navigating pages

rian-schobers:
Fixed #1361 task list table bugs:
- Search input not working properly.
- When search input is active subtasks are shown, they're not filtered out again after emptying the input field.
- Pagination to first page not working properly.

Also fixed: 
- #1489  => Added study name to task read page and task list.

### DIFF
--- a/vantage6-ui/src/app/app.module.ts
+++ b/vantage6-ui/src/app/app.module.ts
@@ -82,6 +82,7 @@ import { RoleReadComponent } from './pages/admin/role/read/role-read.component';
 import { RoleCreateComponent } from './pages/admin/role/create/role-create.component';
 import { RoleFormComponent } from './components/forms/role-form/role-form.component';
 import { OrderByPipe } from './pipes/order-by.pipe';
+import { HighlightedTextPipe } from './pipes/highlighted-text.pipe';
 import { SetupMfaComponent } from './pages/auth/setup-mfa/setup-mfa.component';
 import { MfaCodeComponent } from './pages/auth/mfa-code/mfa-code.component';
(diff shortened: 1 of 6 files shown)


bartvanb (commented):
I think the code looks good. One comment though: I think I'd personally prefer to not show the study in the task list view, because it is not used that often, so the field will often be empty. Plus it will increase load time because studies have to be obtained. @frankcorneliusmartin do you agree?

rian-schobers:
> > I think the code looks good. One comment though: I think I'd personally prefer to not show the study in the task list view, because it is not used that often, so the field will often be empty. Plus it will increase load time because studies have to be obtained. @frankcorneliusmartin do you agree?
> 
> I agree with you, but I also see that people want to filter on study.. which also makes sense. But we cant filter on it if we dont collect it (or can we?). And in that case we just as well display it.

The way filtering works needs to be changed before you can filter on multiple columns. So I think for now it's best to remove it from the list view and look at this issue when/if filtering on multiple columns is supported. @bartvanb & @frankcorneliusmartin , Agree?

bartvanb:
I'm fine with either option

bartvanb (approved):

PR #1574, opened 2024-11-06
Prevent warning about port 443 as insecure whitelisted port because c…

bartvanb:
…omparison between str and int failed - enum is now int-baseclassed which is more correct

I checked the code to see where this Enum was used but I don't think this change will lead to issues there (it will sooner solve bugs we didn't find yet)


frankcorneliusmartin (approved):
Lgtm. Nice that it also works withs ints!

PR #1577, opened 2024-11-07
Updated docs dependencies. Main reason is to fix issues with sphinx a…

bartvanb:
…utosummary generation

One more thing I'd like to slip into 4.8. This will prevent the endless loop we sometimes got stuck in when running `make devdocs` on a different branch (and anyway we hadn't updated these in a over 2 years so we should)


frankcorneliusmartin (approved):
Great job, that was very annoying

Issue #1617, opened 2024-12-03
[Bug] Documentation warning is not rendered in web pages


SanderBeekhuis:
**Describe the bug**

A warning in the documentation sources is not rendered in the web pages.   

**To Reproduce**
1. Go to https://docs.vantage6.ai/en/main/algorithms/classic_tutorial.html
2.  The following warning in the documentation sources is not rendered in the web pages.   
https://github.com/vantage6/vantage6/blob/83ec80a881bfe03f78c43e45771e69f086a6cc8a/docs/algorithms/classic_tutorial.rst?plain=1#L4-L9

**Expected behavior**
The warning is rendered in the web page

**Screenshots**
![image](https://github.com/user-attachments/assets/f53abce4-0796-4b89-bd99-58b7f88a0b9d)

**Details**
 - OS: Windows
 - Version: N/A
- Docker deamon version: N/A

**Additional context**
Warning text: 
 ```
.. warning::

    This classic tutorial was written for vantage6 version 2.x. The commands
    below have not been updated and therefore might not work anymore. We are
    leaving this here for reference, as it includes some useful information
    about concepts that may not be included elsewhere in this documentation

```

bartvanb:
Not sure what was wrong but this has been fixed by #1619.

Maybe relevant to mention: there is a lesson from our workshop from October 2024 that teaches how to create your own vantage6 algorithm [here](https://workshop.vantage6.ai/chap7_algo_dev.html). That is much more up-to-date

SanderBeekhuis:
Thanks for the quick response Bart, the link is helpful.

PR #1619, opened 2024-12-03
Move warning outdated classic tutorial below header so that it will n…

bartvanb:
…ot be invisible


bartvanb:
This probably fixes https://github.com/vantage6/vantage6/issues/1617 but I'm not sure as I cannot reproduce that issue locally. Regardless, this PR simply moves the warning message below the header which is better anyway. If it does not resolve the issue, we should look better what may be causing this

PR #1627, opened 2024-12-10
Implemented v6 node restart command

### LINKED ISSUE #1367: [Feature Request] Add vnode restart command
**Problem description**
I often do vnode stop and vnode start to get a fresh node. It would be helpful to have a vnode restart option that does both. Also give it the --attach flag option such that I can attach directly.

Very nice would also be a vnode restart --all --attach which would reboot all nodes and reattach all consoles currently attached. Not sure if this is possible

bartvanb:
Fix #1367

### DIFF
--- a/vantage6/tests_cli/test_node_cli.py
+++ b/vantage6/tests_cli/test_node_cli.py
@@ -9,7 +9,6 @@ from io import BytesIO, StringIO
 from click.testing import CliRunner
 from docker.errors import APIError
 
-from vantage6.cli.node.restart import cli_node_restart
 from vantage6.common.globals import Ports
 from vantage6.cli.globals import APPNAME
 from vantage6.common import STRING_ENCODING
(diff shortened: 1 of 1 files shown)


bartvanb:
Checked codacy issues, they are OK

frankcorneliusmartin:
> Checked codacy issues, they are OK

You can mark them in Codacy as OK, then the green checkmark will also appear here.

vantage6/vantage6/cli/dev/start.py
```diff
@@ -67,7 +67,7 @@ def start_demo_network(
     cmd = ["v6", "algorithm-store", "start", "--name", f"{ctx.name}_store", "--user"]
     if store_image:
         cmd.extend(["--image", store_image])
-    subprocess.run(cmd)
+    subprocess.run(cmd, check=True)
```
frankcorneliusmartin, line 70: Shouldnt we catch the exception in this case?

frankcorneliusmartin: [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) exception

frankcorneliusmartin (approved):

vantage6/vantage6/cli/dev/start.py
bartvanb: Yes, but let's discuss how. Do we want the execution of the CLI commands to stop if one of the components fails to start? If so, then raise clear error and exit, otherwise print error and continue

PR #1637, opened 2024-12-19
Fixed setting the default value for `session_restrict_to_same_image`

frankcorneliusmartin:


vantage6-server/tests_server/test_resources.py
```diff
@@ -99,6 +99,16 @@ def test_collaboration(self):
...
+        # Create a new collaboration
+        col_details = {
+            "name": "New Collaboration",
+            "organization_ids": [org.id],
+            "encrypted": True,
+            "session_restrict_to_same_image": True,
+        }
+
+        self.app.post("/api/collaboration", json=col_details, headers=headers)
+
```
bartvanb, line 111: assert statement is missing

vantage6-server/vantage6/server/resource/collaboration.py
```diff
@@ -351,13 +353,15 @@ def post(self):
...
+
+        data = request.get_json(silent=True)
+
+        try:
+            data = collaboration_input_schema.load(data)
+        except ValidationError as e:
             return {
                 "msg": "Request body is incorrect",
-                "errors": errors,
+                "errors": e.messages,
```
bartvanb, line 364: The code block you are changing here is more or less identical in many endpoints. What is your reason for changing it?
If we change it, I think we should change it everywhere (or preferably, write some utility function and use that, but I think we didn't because the main function has to return stuff if there are errors)

frankcorneliusmartin: The reason is that default values do not get loaded when you use the `.validate` function. Thus I used to `.load` method to both load (inc. default values) and validate the user input.

Not sure we want to do that everywhere? Else I suggest we open a separate issue for this.

bartvanb: If it doesn't make a difference if no defaults are present, than I'd prefer to change it everywhere, because it keeps the code uniform and prevents us from making mistakes in the future. Shouldn't be too much work also. 
I'm fine with doing it in a separate issue.

frankcorneliusmartin: I included it in this PR

bartvanb (approved):

PR #1642, opened 2024-12-23
Feature/#1598 download result

rian-schobers:
downloadresult

### DIFF
--- a/vantage6-ui/src/app/pages/analyze/task/read/task-read.component.ts
+++ b/vantage6-ui/src/app/pages/analyze/task/read/task-read.component.ts
@@ -129,13 +129,13 @@ export class TaskReadComponent implements OnInit, OnDestroy {
     if (!this.task || sync_tasks) {
       this.task = await this.getMainTask();
       this.childTasks = await this.getChildTasks();
-      if(this.task.study)
-        this.study = await this.studyService.getStudy(this.task.study.id.toString())
+      if (this.task.study) this.study = await this.studyService.getStudy(this.task.study.id.toString());
     }
     try {
       if (this.task.algorithm_store) {
         const store = await this.algorithmStoreService.getAlgorithmStore(this.task.algorithm_store?.id.toString());
         this.algorithm = await this.algorithmService.getAlgorithmByUrl(this.task.image, store);
+        // Please note, the function cannot be set if the input cannot be decoded. This will result in NO visualization and information about the function.
         this.function = this.algorithm?.functions.find((_) => _.name === this.task?.input?.method) || null;
         if (!this.selectedVisualization) {
           // by checking in if statement whether visualization was already set, we prevent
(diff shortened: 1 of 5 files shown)


vantage6-ui/src/app/services/task.service.ts
```diff
@@ -123,31 +112,47 @@ export class TaskService {
...
-    if (this.chosenCollaborationService.isEncrypted()) {
-      decryptedResult = this.encryptionService.decryptData(taskResult.result);
-    } else {
-      // if not decrypted, just decode it
-      decryptedResult = atob(taskResult.result);
+    let decryptedResult = taskResult.result;
+    try {
+      decryptedResult = isEncrypted ? this.encryptionService.decryptData(taskResult.result) : atob(taskResult.result);
+    } catch (error) {
+      console.error(this.translateService.instant(errorTranslationCode) + ': ', error);
```
bartvanb, line 122: can you get rid of this console error as well as the other ones in this file?

vantage6-ui/src/app/pages/analyze/task/read/task-read.component.ts
```diff
@@ -299,9 +299,20 @@ export class TaskReadComponent implements OnInit, OnDestroy {
...
+    const textInput = JSON.stringify(run.input);
+    this.fileService.downloadTxtFile(textInput, filename);
+  }
+
   downloadResult(result: TaskResult): void {
     const filename = `vantage6_result_${result.id}.txt`;
-    const textResult = JSON.stringify(result.decoded_result);
+    let textResult = '';
+    if (result.decoded_result === undefined) {
+      textResult = JSON.stringify(result.result);
```
bartvanb, line 312: This should be changed to `textResult = result.result || '';` - the result will already be a string if it can not be decoded/decrypted. Maybe we should change the type of it in the UI interface of the result (it was initially an object because it could be overwritten with the decoded result, but that is now in a separate field)

bartvanb: The main reason for changing this is, by the way, that you now get quotes around the result when you download it, that should not be there.

vantage6-ui/src/app/pages/analyze/task/read/task-read.component.ts
```diff
@@ -299,9 +299,20 @@ export class TaskReadComponent implements OnInit, OnDestroy {
     }
   }
 
+  downloadInput(run: TaskRun): void {
+    const filename = `vantage6_input_${run.id}.txt`;
+    const textInput = JSON.stringify(run.input);
```
bartvanb, line 304: See comment just below here on how to change

vantage6-ui/src/app/services/task.service.ts
```diff
@@ -123,31 +112,47 @@ export class TaskService {
...
-    if (this.chosenCollaborationService.isEncrypted()) {
-      decryptedInput = this.encryptionService.decryptData(taskRunInput);
-    } else {
-      decryptedInput = atob(taskRunInput);
+    let decryptedInput = '';
+    const isEncrypted = this.chosenCollaborationService.isEncrypted();
+    try {
+      decryptedInput = isEncrypted ? this.encryptionService.decryptData(taskRunInput) : atob(taskRunInput);
+    } catch (error) {
+      console.error(this.translateService.instant('task.alert-failed-read-input') + ': ', error);
```
bartvanb, line 144: also remove console errors in this file

bartvanb (changes requested):
A few minor changes

PR #1643, opened 2024-12-23
Extend help text in v6 dev create for Linux users

bartvanb:


vantage6/vantage6/cli/dev/create.py
```diff
@@ -485,8 +485,9 @@ def demo_network(
     "--server-url",
     type=str,
     default="http://host.docker.internal",
-    help="Server URL to point to. If you are using Docker Desktop, "
-    "the default http://host.docker.internal should not be changed.",
+    help="Server URL to point to. If you are using Docker Desktop, the default "
+    "http://host.docker.internal should not be changed. If you are using Linux without"
+    " Docker Desktop, you should set this to http://172.17.0.1",
```
frankcorneliusmartin, line 490: This is not 100% guaranteed. But its the Docker default, so I am kinda OK with it.

frankcorneliusmartin (approved):

vantage6/vantage6/cli/dev/create.py
bartvanb: Yeah I know, but I didn't want to make the message more complex to take the 0.0001% of people into account that do this. (if they are crazy enough to do that they hopefully also know what they're doing)

PR #1653, opened 2025-01-14
Add debugging and testing possibilities

lsago:
This set of changes includes:

* Add v6-dev-profile to quickly stand up a pre-defined network of server and nodes
* Allow debugging of node and server containers
* Use of relative paths in config files to support localized testing
* Allow debugging of algorithms (running in container)

These changes should (hopefully) aid development of vantage6 itself as well as vantage6 algorithms. It's possible to set breakpoints while running the "real deal" (containers). By allowing all files to reside in one directory (no ~/.config/vantage6, etc) and other changes we open the door for more integration tests that are closer to a production environment (no mocking & and code running in containers).

### DIFF
--- a/dev/.gitignore
+++ b/dev/.gitignore
@@ -1,5 +1,5 @@
 # virtual environments
-venv/
+venv*
 .venv/
 
 # debugpy package for use within container


bartvanb:
Hi @lsago thanks! I'll be sure to have a look when you finish and will involve others if necessary. I think it makes sense that I'll have a look once you say it's finished. I see it's quite big but you have some docs included so I should be able to figure it out.

Issue #1589, opened 2024-11-22
[Feature Request] Option for algorithm developer to give comments when they submit an algorithm


bartvanb:
**Problem description**
It would be nice if an algorithm developer has a free text field to give comments when they submit an algorithm to the store. Therein, they could e.g. detail what is different compared to the previous version of this algorithm

**Desired solution**
- Free text field for details when submitting algorithm
- Reviewer gets to read these things on the UI pages where review can be submitted

Issue #1517, opened 2024-10-03
[Feature Request] Add option to describe parameters with markup style


bartvanb:
**Problem description**
Sometimes descriptions of parameters of functions are rather long. In the UI they are just displayed in one paragraph in bold, which is not always desirable.

**Desired solution**
Allow the user that submits the algorithm to the store to specify e.g. markdown-formatted descriptions of the parameters

PR #1661, opened 2025-01-15
 Developer algorithm comment

wbaccinelli:
Added an exstra field in the algorithm to allow the developer to add free text comments at the submission time.

#1589


vantage6-ui/src/environments/environment.development.ts
```diff
@@ -4,7 +4,7 @@ const env: EnvironmentConfig = {
   production: false,
   // server_url: 'https://cotopaxi.vantage6.ai',
   // api_path: ''
-  server_url: 'http://localhost:7601',
```
bartvanb, line 7: Can you please change this back? 7601 is the default port now for a v6 server

vantage6-ui/src/app/components/algorithm/display-algorithm/display-algorithm.component.html
```diff
@@ -62,6 +62,14 @@
         <b>{{ "algorithm.invalidated-at" | translate }}</b>
         <span>{{ printDate(algorithm?.invalidated_at) }}</span>
       </div>
+      <div class="data-list__item data-list__item--big">
+        <b>{{ "algorithm.description" | translate }}</b>
+        <span>{{ algorithm?.description || "-" }}</span>
+      </div>
```
bartvanb, line 68: The description was already there so it can be deleted.

I think also the submission comments don't need to be included here. I think they need a prominent place in algorithm review, but once reviewed these comments shouldn't always be visible.
Maybe you could add an *ngIf to only show it as long as algorithm is not yet approved (or invalidated etc)?

vantage6-ui/src/app/components/forms/algorithm-form/algorithm-form.component.html
```diff
@@ -55,6 +55,10 @@
             </mat-option>
           </mat-select>
         </mat-form-field>
+        <mat-form-field subscriptSizing="dynamic">
+          <mat-label>{{ "algorithm.submission-comments" | translate }}</mat-label>
+          <input matInput type="text" formControlName="submission_comments" />
```
bartvanb, line 60: I feel like it would be nice if this input field was a bit bigger (ideally resizable) so that developers don't feel that they have to make their submission all fit into a single line. Sometimes, you may have changed several things at once compared to last version so then it would be nice to make a bullet point list.

Maybe then it would also make sense to put this on the very top of the form?

vantage6-ui/src/app/pages/store/algorithms-in-review/review-read/review-read.component.html
```diff
@@ -42,6 +42,10 @@
             <span>-</span>
           </div>
         </div>
+        <div class="data-list__item">
+          <b>{{ "algorithm.submission-comments" | translate }}</b>
+          <span>{{ algorithm?.submission_comments || "-" }}</span>
+        </div>
```
bartvanb, line 48: Here, I would also like to have the submission comments featured more prominently. Maybe put them in a separate mat-card on top? Or give them otherwise some more emphasis rather than just being one of the data fields.

bartvanb (changes requested):
That was done quickly @wbaccinelli! It also functions just like I imagined it. My only comments are around styling: I feel that the developer's comments should be featured more prominently on certain places, like when you are a reviewer looking at the algorithm to review, and in other places they may be left out. See my comments below for details, and if you disagree, please reach out!

vantage6-ui/src/environments/environment.development.ts
wbaccinelli: Of course, this should have not been committed.

vantage6-ui/src/app/components/algorithm/display-algorithm/display-algorithm.component.html
wbaccinelli: Yes, this must have been a mistake 😅.
I was also doubting whether to display this always or not. I will make it conditional to the status.

vantage6-ui/src/app/components/forms/algorithm-form/algorithm-form.component.html
wbaccinelli: It makes a lot of sense!

wbaccinelli:
@bartvanb I implemented the changes. One thing I want to point out is that, in order to make the comments readable, I made them readable as markdown. To enable the markdown format, I needed to install [ngx-markdown](https://www.npmjs.com/package/ngx-markdown), and that's why the package.json and package-lock.json have been changed.

bartvanb (approved):
Great work! Now it looks exactly as I wanted it. Also great that you already added Markdown support, we need that later for e.g. #https://github.com/vantage6/vantage6/issues/1517


I have one final (tiny) request: on the page where the reviewer views the algorithm:
image

Could you change `Submission comments` to something like `Comments from algorithm developer` (possibly including the (user)name of the developer)? Just to make it a bit more verbose where the comments are coming from

Issue #1093, opened 2024-03-14
[Feature Request] Flag in algorithm store which functions can be run by users


bartvanb:
**Problem description**
Sometimes it doesn't make sense to run particular algorithm functions as a user - it only makes sense to e.g. run the orchestrator/aggregator function. 

**Desired solution**
For these cases, it would be nice to set a flag to prevent a user from running a function that they shouldn't run

PR #1690, opened 2025-01-20
#1093 - Hide function flag

bbrouweriknl:
Hide function in create task UI to prevent them from running it directly

### DIFF
--- a/vantage6-algorithm-store/vantage6/algorithm/store/__init__.py
+++ b/vantage6-algorithm-store/vantage6/algorithm/store/__init__.py
@@ -135,7 +135,7 @@ class AlgorithmStoreApp:
                 algorithm.status = ReviewStatus.APPROVED.value
                 algorithm.submitted_at = datetime.datetime.now(datetime.timezone.utc)
                 algorithm.approved_at = datetime.datetime.now(datetime.timezone.utc)
-                function.save()
+                algorithm.save()
 
         if self.ctx.config.get("dev", {}).get("disable_review", False):
             self.setup_disable_review()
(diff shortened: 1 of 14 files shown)


bartvanb (changes requested):
It works perfectly, great first contribution! My only comment is on the naming. This has to do with that vantage6 is not simply a frontend-backend app so the flag 'hidden' does not apply to the non-UI scenarios (which we also support)

vantage6-algorithm-store/vantage6/algorithm/store/__init__.py
```diff
@@ -135,8 +135,13 @@ def __init__(self, ctx: AlgorithmStoreContext) -> None:
                 algorithm.status = ReviewStatus.APPROVED.value
                 algorithm.submitted_at = datetime.datetime.now(datetime.timezone.utc)
                 algorithm.approved_at = datetime.datetime.now(datetime.timezone.utc)
-                algorithm.save()
```
bartvanb, line 138: this line shouldn't be removed

vantage6-algorithm-store/vantage6/algorithm/store/__init__.py
```diff
@@ -135,8 +135,13 @@ def __init__(self, ctx: AlgorithmStoreContext) -> None:
...
-
+        
+        # TODO v5+ remove this - for backwards compatibility we are setting the standalone property in all functions created before v4.11 to true
```
bartvanb: Nitpicking: we're releasing this in 4.10, not 4.11

vantage6-ui/src/assets/localizations/en.json
```diff
@@ -232,8 +232,8 @@
     "invalidated-at": "Invalidated at",
     "function": {
       "type": "Type of function",
-      "hidden": "Hide function",
-      "hidden_tooltip": "Hide this function in the create task UI to prevent a user from running it directly.",
+      "standalone": "Standalone function",
+      "standalone_tooltip": "Whether this function produces useful results when running it by itself. Disabling this option will hide the function in the create task UI.",
```
bartvanb, line 236: 'in the create task UI' -> 'when creating tasks in this UI'

bartvanb (commented):
Some very minor comments. Feel free to merge once addressed

PR #1694, opened 2025-01-22
Catch errors when refreshing node client's token so that the thread t…

bartvanb:
…o refresh the token doesn't crash

Note that this was introduced in v3.8 when refresh tokens got a limited validity (https://github.com/vantage6/vantage6/commit/bd0edc010ed54df8e9ed68bf42c39deb2fd2a0b0)

I want to release this in 4.9.1 together with a few other fixes.

### DIFF
--- a/vantage6-common/vantage6/common/client/node_client.py
+++ b/vantage6-common/vantage6/common/client/node_client.py
@@ -80,7 +80,12 @@ class NodeClient(ClientBase):
                 "exp"
             ]
             time_until_expiry = expiry_time - time.time()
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
What happens when the token expires? Will it go on indefinetely or do we get an exception/exit somewhere else?

bartvanb:
> What happens when the token expires? Will it go on indefinetely or do we get an exception/exit somewhere else?

Yes it would go on indefinitely. I'll add an extra check to exit as it doesn't make sense to have a dead node

Issue #1495, opened 2024-09-23
[Change Request] Version 5 Sessions overview issue


frankcorneliusmartin:
**Description**
As the original pull request got to big for my taste I separated a lot off issues into new issues. This issue is to collect everything that needs to be done before we can release v5.

#1469
#1470 
#1471
#1476 
#1477 
#1481 
#1482 
#1483 
#1484 
#1485

PR #1696, opened 2025-01-22
Prevent that node tries to start task when there is no valid token

bartvanb:


frankcorneliusmartin (approved):

frankcorneliusmartin:
FYI in v5 not all containers get a token #1495

PR #1699, opened 2025-01-23
Add quickstart section to the docs 

bartvanb:
This should help users to run v6 dev network easily

### DIFF
--- a/docs/introduction/quickstart.rst
+++ b/docs/introduction/quickstart.rst
@@ -33,15 +33,24 @@ Then, install the vantage6 command line interface (CLI) by running:
 Start a local vantage6 network
 ------------------------------
 
-You can easily set up a local vantage6 network by running the following command:
+Before starting a local vantage6 network, make sure Docker is running. Then, in the
+Python environment where you installed the vantage6 CLI, you can easily set up a local
+vantage6 network by running the following command:
 
 .. code-block:: bash
 
     v6 dev create-demo-network
 
-This will start an interactive dialog that will help you to easily configure your server
-and nodes. Note that you can also do ``v6 dev create-demo-network --help`` to view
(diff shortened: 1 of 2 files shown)


docs/introduction/quickstart.rst
```diff
@@ -0,0 +1,73 @@
...
+Then, install the vantage6 command line interface (CLI) by running:
+
+.. code-block:: bash
+
+    pip install vantage6
+
+Start a local vantage6 network
+------------------------------
+
+You can easily set up a local vantage6 network by running the following command:
```
wbaccinelli, line 36: Should we explicitly mention that the Docker must be running? I'm saying this just because by chance my Docker was on pause while trying and I got an error.

docs/introduction/quickstart.rst
```diff
@@ -0,0 +1,73 @@
...
+
+.. code-block:: bash
+
+    v6 dev create-demo-network
+
+This will start an interactive dialog that will help you to easily configure your server
+and nodes. Note that you can also do ``v6 dev create-demo-network --help`` to view
+which custom options are available.
```
wbaccinelli, line 44: I see that the only thing that is asked is the name to be assigned to the demo network. This could be rephrased to something like: "This will start an interactive dialog that will ask you to assign a name to the network. Note that defaults settings will be used, but custom options can be used. You can run ``v6 dev create-demo-network --help`` to view which custom options are available."

wbaccinelli (approved):
Nice to have these instructions in the documentation, very useful! I left just a couple of minor comments.

docs/introduction/quickstart.rst
bartvanb: Good suggestion!

PR #1706, opened 2025-01-29
Exit a node at startup time if a file-based database doesn't exist

bartvanb:
Also improved log message

### DIFF
--- a/docker/node-and-server.Dockerfile
+++ b/docker/node-and-server.Dockerfile
@@ -53,8 +53,6 @@ RUN CFLAGS="-I/usr/local/opt/openssl/include" \
 
 RUN chmod +x /vantage6/vantage6-server/server.sh
 
-RUN pip install jinja2==3.1.5
-
 # expose the proxy server port
 ARG port=80
 EXPOSE ${port}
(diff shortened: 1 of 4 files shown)


docker/node-and-server.Dockerfile
```diff
@@ -53,6 +53,8 @@ RUN CFLAGS="-I/usr/local/opt/openssl/include" \
 
 RUN chmod +x /vantage6/vantage6-server/server.sh
 
+RUN pip install jinja2==3.1.5
```
frankcorneliusmartin, line 56: Any reason why to do this here? And not in setup.py / requirements.txt?

vantage6/vantage6/cli/node/start.py
```diff
@@ -243,6 +243,14 @@ def cli_node_start(
             # we catch all exceptions here.
             file_based = False
 
+        if db_type in ["folder", "csv", "parquet", "excel"] and not file_based:
```
frankcorneliusmartin, line 246: Could we rename `file_based` to `db_file_exists`. For me it was hard to read (I probably named it so). But now the `force_db_mount` option will not work as it should skip any check. See line 254.

frankcorneliusmartin (changes requested):

docker/node-and-server.Dockerfile
bartvanb: Oops this was something not meant to be committed - I'll remove again

vantage6/vantage6/cli/node/start.py
bartvanb: Will do so!

PR #1719, opened 2025-02-07
Catch Ctrl+C interrupts in CLI properly

bartvanb:


frankcorneliusmartin (approved):
LGTM. However one could consider writing a function to handle the exception to the `exit`. But I am OK to merge it as it is.

PR #1728, opened 2025-02-14
Implement `/role` and `/role/rules` endpoints in algorithm store

riccardosamperna:
This PR implements the missing API endpoints for managing the roles in the algorithm store. A refactoring of some of the components of the role management for the server has also been picked up to reduce code duplication. The corresponding UI elements have been also been implemented.


bartvanb (commented):
You have indeed been very thorough, I love it! It does feel a bit like opening Pandora's box as now I feel we should update all API endpoints this way - and I'm convinced that in that case, we will be refactoring even further as functions like `_get_organization_id` will be used in more endpoints so they should move to a higher level.

But to me it is obvious that this is a significant improvement. We should however also update the other endpoints in the near future to keep things consistent, which will be quite a bit of work. I suggest you work on it now for the `/role` endpoints and that we discuss tomorrow how we tackle the other endpoints.

vantage6-backend-common/vantage6/backend/common/resource/input_schema.py
bartvanb, line 1: If we are already refactoring, I was already getting a bit annoyed with the size of the `input_schema` modules, so I'd like to suggest a folder `input_schema` which contains a `role.py` that contains the role schema's and the base schema's could go in the __init__ or so

vantage6-server/vantage6/server/permission.py
```diff
@@ -345,27 +346,21 @@ def register_rule(
...
 
         Parameters
         ----------
         rules: list[:class:`~vantage6.server.model.rule.Rule`]
             List of rules that user is checked to have
-
-        Returns
-        -------
-        dict | None
-            Dict with a message which rule is missing, else None
```
bartvanb, line 360: Could you add a `Raises` section to the docstring?

vantage6-server/vantage6/server/resource/role.py
```diff
@@ -139,17 +157,115 @@ def permissions(permissions: PermissionManager) -> None:
...
+            query = query.filter(
+                or_(
+                    db.Role.organization_id == auth_org.id,
+                    db.Role.id.in_(own_role_ids),
+                    db.Role.organization_id.is_(None),
+                )
+            )
+        else:
+            query = query.filter(db.Role.id.in_(own_role_ids))
+        return query
```
bartvanb, line 265: I think we can probably generalize this even further so that it can also be used with other GET / endpoints. However as discussed, let's do that in v5

vantage6-server/vantage6/server/resource/role.py
```diff
@@ -520,29 +508,15 @@ def get(self, id):
...
-            or (
-                role.name in [role for role in DefaultRole]
-                and self.r.has_at_least_scope(Scope.ORGANIZATION, P.VIEW)
-            )
-        ):
-            return {
-                "msg": "You do not have permission to view this."
-            }, HTTPStatus.UNAUTHORIZED
+        if not self.has_permission_to_view(role):
+            raise UnauthorizedError("You do not have permission to view this.")
```
bartvanb, line 514: I know you just copied this, but I'd suggest changing the message to "to view this role"

vantage6-server/vantage6/server/resource/role.py
```diff
@@ -602,60 +576,21 @@ def patch(self, id):
...
-                if not rule:
-                    return {
-                        "msg": f"rule with id={rule_id} not found!"
-                    }, HTTPStatus.NOT_FOUND
-                rules.append(rule)
-            denied = self.permissions.check_user_rules(rules)
-            if denied:
-                return denied, HTTPStatus.UNAUTHORIZED
-            role.rules = rules
+            raise BadRequestError("Organization id cannot be changed.")
```
bartvanb, line 582: I'd prefer an error similar to the old one: "Cannot change role's organization" - I'd mention role to be more specific and not mention ID because the user doesn't care about the underlying IDs (at least not if they're using the UI)

bartvanb (commented):
Nice work! I only had time to look at the backend code now, will review the frontend next week.

vantage6-server/vantage6/server/resource/role.py
```diff
@@ -139,17 +146,158 @@ def permissions(permissions: PermissionManager) -> None:
...
+    return None
 
 
 class RoleBase(ServicesResources):
     def __init__(self, socketio, mail, api, permissions, config):
         super().__init__(socketio, mail, api, permissions, config)
-        self.r: RuleCollection = getattr(self.permissions, module_name)
+        self.rule_collection: RuleCollection = getattr(self.permissions, module_name)
+
+    def _get_role(self, role_id):
```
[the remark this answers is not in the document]
frankcorneliusmartin: So I think we have two discussions here:

(1) Do we need to document all 'private' functions and (2) The API documentation in the docstrings are very long.

I think for (1) I could see that we are not as extensive, however I do like a one-liner which explains the function a little more than just the function signature. Leaving out an extensive description, arguments, return types, etc. is fine for me. I do like to have typing for the arguments though.

About (2), the reason why the API docs are so close to the code is so we would not forget to update the docs when we make changes. However I think you also said something about FastAPI doing this differently anyway @bartvanb? I am not super attached to having the documentation here, but I am also not bothered by it either.

vantage6-ui/src/app/app-routing.module.ts
```diff
@@ -472,6 +473,14 @@ const routes: Routes = [
...
         }
       },
+      {
+        path: routerConfig.storeRoleCreate,
+        component: StoreRoleCreateComponent,
+        canActivate: [authenticationGuard(), chosenStoreGuard()],
+        data: {
+          crumbs: [['role-list.title', routePaths.storeRoles], ['role-create.title']]
+        }
+      },
```
bartvanb, line 483: I'd expect another path to be added here to edit roles? Or is that something that still has to be implemented?

vantage6-ui/src/app/components/forms/store-role-form/store-role-form.component.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.store = this.chosenStoreService.store$.value;
+    this.isLoading = false;
+  }
+
+  form = this.fb.nonNullable.group({
+    name: ['', [Validators.required]],
+    description: '',
+  });
+
+  selectedRules: number[] = [];
```
bartvanb, line 42: It's common practice (I believe) in Angular to declare all class variables at the top of the class. And also, we do that in all other classes so that's another reason to move these two variables

vantage6-ui/src/app/app-routing.module.ts
bartvanb: Ah never mind, I saw that this is implemented in the StoreRoleRead Component

bartvanb (commented):
I just also checked the UI part, that looks pretty good. 

While testing it, I only ran into the following error:
image
That was after assigning this role to users, could you ensure that in that case the `delete_dependents` flag is properly passed along?

It did remind me of one more thing to do: we aim to have the full functionality of the API also covered in the Python client. Could you also add the functionality to create/edit/delete store roles there? This should be pretty straightforward

bartvanb (approved):
Everything was addressed nicely.
I noticed one tiny thing in the type hinting using `Dict` that I replaced with `dict` in a separate command. Using `Dict` is the 'old' way of doing this (Python <3.10 I think) - note that we only use `List` in this file to avoid a conflict with the `list()` function if we would use `list` as a type

Issue #1729, opened 2025-02-17
[Bug] Pre-releases overwrite `latest` tag of the `harbor2.vantage6.ai/infrastructure/algorithm-base` image


frankcorneliusmartin:
**Describe the bug**
This should not be the case as then a pre-released algorithm base is the default.

bartvanb:
@frankcorneliusmartin our current procedure is that we *always* push latest tag whenever we push new images. It's just that with other images, like the node image, we don't run into it because we always pull the `cotopaxi` tag.

So if we change this, I think we should change it for all images. My proposal is to always push `latest` tag for any release that is not a prerelease.

PR #1744, opened 2025-02-19
Only add latest tag to infra images for final releases, not pre-releases

bartvanb:
#1729


frankcorneliusmartin (approved):
Ok, so no pre releases for the algorithm base? Seems reasonable as you first need an infra before developing algorithms? Or do we want people to be able to develop on a pre release?

bartvanb:
> Ok, so no pre releases for the algorithm base? Seems reasonable as you first need an infra before developing algorithms? Or do we want people to be able to develop on a pre release?

No, I think there *will be* an `algorithm-base:5.0` image (and similar for other images). However, this will no longer also get the `latest` tag as well. Can you explain why you thought otherwise?

PR #1758, opened 2025-02-25
feat: Add common Helm chart and update node/store charts to use it

frankcorneliusmartin:
- Created a new common Helm chart with reusable helper templates
- Updated node and store charts to use the common chart's helper functions
- Added Chart.lock files for node and store charts
- Updated .gitignore to exclude Helm chart packages
- Added helm charts to the pipeline

In case `integration/k8s-sessions` has been merged. This can directly be merged into `release/5.0.0`.


frankcorneliusmartin:
I still need to run the pipeline in order to test it.

.github/workflows/release.yml
```diff
@@ -91,6 +91,29 @@ jobs:
...
+
+          # Update store chart and its dependency
+          sed -i "s/appVersion: \"0.0.0\"/appVersion: \"${version}\"/" charts/store/Chart.yaml
+          sed -i "s/version: 0.0.0/version: ${version}/" charts/store/Chart.yaml
+          sed -i "/name: common/,/repository:/ s/version: .*/version: \"${version}\"/" charts/store/Chart.yaml
+
+          # Update server chart and its dependency
+          sed -i "s/appVersion: \"0.0.0\"/appVersion: \"${version}\"/" charts/server/Chart.yaml
+          sed -i "s/version: 0.0.0/version: ${version}/" charts/server/Chart.yaml
+          sed -i "/name: common/,/repository:/ s/version: .*/version: \"${version}\"/" charts/server/Chart.yaml
```
bartvanb, line 115: As we discussed, could you move these changes into a bash script and then call that via make? That makes it easier to test it locally and keeps this pipeline cleaner.

.github/workflows/release.yml
```diff
@@ -129,6 +152,38 @@ jobs:
...
+          USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          PASSWORD: ${{ secrets.DOCKER_TOKEN }}
+        run: docker login harbor2.vantage6.ai -u $USERNAME -p $PASSWORD
+
+      - name: Build Helm charts
+        run: |
+          helm package charts/common
+          helm package charts/node
+          helm package charts/store
+          helm package charts/server
```
bartvanb, line 178: I was wondering if the `helm` command is available, so I checked and [this link I clicked](https://medium.com/@yassinejedidi10/deploying-your-app-with-helm-kubernetes-and-github-actions-a-complete-ci-cd-pipeline-8b57f4f5b563) seems to suggest

```
    - name: Install Helm
      uses: azure/setup-helm@v1
```

There also is a github repo to release helm charts: https://github.com/helm/chart-releaser-action. 
Not sure what makes sense in our case

.github/workflows/release.yml
bartvanb: Also, could you make it a for loop, as the commands are almost identical?

charts/node/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: common
+  repository: file://../common
+  version: 0.1.0
+digest: sha256:636a65e9846bdff17cc4e65b0849061f783759a37aa51fb85ff6fd8ba5e68467
+generated: "2025-02-25T13:15:43.797510864+01:00"
```
bartvanb, line 6: this file looks very specific, should this be included?

charts/node/templates/_helpers.tpl
```diff
@@ -11,16 +11,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-{{- if .Values.fullnameOverride }}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
-{{- if contains $name .Release.Name }}
-{{- .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
-{{- end }}
-{{- end }}
```
bartvanb, line 23: nice that this is moved to common functionality

charts/store/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: common
+  repository: file://../common
+  version: 0.1.0
+digest: sha256:636a65e9846bdff17cc4e65b0849061f783759a37aa51fb85ff6fd8ba5e68467
+generated: "2025-02-25T13:16:42.947314479+01:00"
```
bartvanb, line 6: Again not sure if this needs to be here?

charts/server/Chart.yaml
```diff
@@ -15,10 +15,15 @@ type: application
...
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "1.16.0"
+appVersion: "0.0.0"
+
+dependencies:
+  - name: common
+    version: "0.0.0"
+    repository: file://../common
```
bartvanb, line 29: Just wondering: for node and store, you have changed 3 files, for the server just the 1. Why are there no changes here for Chart.lock and _helpers.tpl?

bartvanb (changes requested):
Nice improvements!

charts/node/Chart.lock
frankcorneliusmartin: This is what you get when using `helm dependency update` which is required to be able to use the common package in the other charts. I could check wether we could completely remove this lock file (as it is a local Chart anyway) but it is typical to include this lock file in your repo.

.github/workflows/release.yml
frankcorneliusmartin: Done.

charts/server/Chart.yaml
frankcorneliusmartin: Thanks, I completely skipped the server... Now the server also has 3.

PR #1762, opened 2025-02-26
Fix/integration/k8s sessions/ports and db errors

hcadavid:
- Fixes DB error caused by the SQL statements created by sqlalchemy when the 'encryption' setting is inserted on the nodeconfig table as a boolean value, and then a non-boolean value is inserted. Why Postgress SQL is having this dynamic-type behavior is yet to be explored.
- Node proxy port allocation process adjusted to avoid error caused by a platform-dependent error code condition.
- Using the FQDN of the server service for the node configuration
- Replacing error-prone file-system shortcuts on devspace.yml with ENV variables


vantage6-node/vantage6/node/__init__.py
```diff
@@ -179,22 +182,36 @@ def __proxy_server_worker(self) -> None:
...
+                )
                 self.log.exception(e)
+                # After a non-os related exception there shouldn't be more retries
+                raise
+
+        if not port_assigned:
+            self.log.error(
+                f"Unable to assing a port for the node proxy after {try_number} attempts"
+            )
+            raise OSError(
```
frankcorneliusmartin, line 212: Maybe we should let the node die here with an `exit(1)`? @bartvanb what do you think? A node without proxy is not able to run master tasks...

vantage6-node/vantage6/node/__init__.py
```diff
@@ -179,22 +182,36 @@ def __proxy_server_worker(self) -> None:
...
 
             except OSError as e:
                 self.log.info("Error during attempt %s", try_number)
                 self.log.info("%s: %s", type(e), e)
 
-                if e.errno == 48:
-                    proxy_port = random.randint(2048, 16384)
-                    self.log.warning("Retrying with a different port: %s", proxy_port)
-                    os.environ["PROXY_SERVER_PORT"] = str(proxy_port)
+                # TODO use error codes to refine exception handling in this method
```
frankcorneliusmartin, line 191: Remove the commented out stuff.

devspace.yaml
frankcorneliusmartin, line 1: Cool, did not know of these default vars.

frankcorneliusmartin (changes requested):
I dont understand why we need to change to port 80? I would prefer to keep the server listening on 7601

hcadavid:
> I dont understand why we need to change to port 80? I would prefer to keep the server listening on 7601

This is because the K8S service for the V6-Server is configured to run at port 80. This could be changed, but I'm not sure if there are other dependencies to this particular port.


https://github.com/vantage6/vantage6/blob/5492b808304c657c2a8d4c44d4e2bcaaf3e3723b/charts/server/templates/server/server-backend-service.yml#L1-L18

bartvanb (approved):

PR #1764, opened 2025-02-26
Fixing race condition when initializing server and store postgresSQL instances

hcadavid:
I observed that sometimes either the V6-Server database or the Store database is not initialized. According to the logs, this initialization is skipped due to an existing datafile. After that, there were errors of the type 'vantage6' / 'store' database doesn't exist.

I think this is because both PSQL services were using the same data file. I created separate ones for each one and haven't seen these errors back.

### DIFF
--- a/charts/server/values.yaml
+++ b/charts/server/values.yaml
@@ -53,7 +53,7 @@ database:
   name: vantage6
 
   #hostpath of the database mount
-  volumePath: /mnt/data
+  volumePath: /mnt/data_server
 
 ui:
 
--- a/charts/store/values.yaml
+++ b/charts/store/values.yaml
@@ -33,4 +33,4 @@ database:
   name: store
 
   #hostpath of the database mount
-  volumePath: /mnt/data
+  volumePath: /mnt/data_store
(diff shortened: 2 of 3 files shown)


frankcorneliusmartin (approved):
Hi Hector good catch. I approve it. Would it make sense to make this a variable in `values.yml`?

However, I got a message from a more Kubernetes experts that having a DB in your k8s cluster is not good practice. Maybe for development it is oke, but we should consider external dbs for this purpose. (or figure out what a good practice is)

hcadavid:
> Hi Hector good catch. I approve it. Would it make sense to make this a variable in `values.yml`?
> 

Yes, definitely... for a dev environment, it would be good to be able to set shorter 'retry' time

> However, I got a message from a more Kubernetes experts that having a DB in your k8s cluster is not good practice. Maybe for development it is oke, but we should consider external dbs for this purpose. (or figure out what a good practice is)

That's right, this makes sense only for development purposes. K8S is not intended for 'scaling' RDMBS infrastructure, and a PersistentVolume is not the best place to keep (reliably) DB's datafiles. In a proper setup, the RDMS should be external, running on its own infrastructure. But this can be done already, right?

PR #1765, opened 2025-02-27
Feature/#1351 Live logging

riccardosamperna:
This PR implements a logging mechanism using Docker native `logs(stream=True)` for a running container. This should allow to easily see what is happening when an algorithm is running. 

Logs in the UI seem to be available also for killed tasks.

### DIFF
--- a/vantage6-node/vantage6/node/docker/docker_manager.py
+++ b/vantage6-node/vantage6/node/docker/docker_manager.py
@@ -650,6 +650,7 @@ class DockerManager(DockerBaseManager):
                 NodePolicy.REQUIRE_ALGORITHM_PULL, DEFAULT_REQUIRE_ALGO_IMAGE_PULL
             ),
             socketIO=socketIO,
+            collaboration_id=self.client.collaboration_id,
         )
 
         # attempt to kick of the task. If it fails do to unknown reasons we try
--- a/vantage6-node/vantage6/node/docker/docker_manager.py
+++ b/vantage6-node/vantage6/node/docker/docker_manager.py
@@ -649,6 +649,7 @@ class DockerManager(DockerBaseManager):
             requires_pull=self._policies.get(
                 NodePolicy.REQUIRE_ALGORITHM_PULL, DEFAULT_REQUIRE_ALGO_IMAGE_PULL
             ),
+            socketIO=socketIO,
         )
 
         # attempt to kick of the task. If it fails do to unknown reasons we try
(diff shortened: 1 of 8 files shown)


vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -659,6 +661,8 @@ def run(
                     databases_to_use=databases_to_use,
                 )
 
+                task.stream_logs(socketIO, run_id, task_info["id"])
```
bartvanb, line 664: We might miss the first few log messages by starting it after the task is fully initialized - maybe that can be improved

bartvanb (commented):
I like doing this over the socket channel - it's easy and straightfoward. Especially because we could also share the log messages over socket to the UI, where the logs can then be shown without a need to do API polling. 

As you realized, this way of doing it is not entirely compatible with v5, but essentially the only place where you're using the Docker API is in `self.container.logs(stream=True)` and I'm sure that can be transferred to k8s easily

vantage6-server/vantage6/server/websockets.py
```diff
@@ -409,6 +409,41 @@ def __alert_node_status(self, online: bool, node: Authenticatable) -> None:
...
+            return
+
+        run_id = data.get("run_id")
+        task_id = data.get("task_id")
+        log_message = data.get("log")
+
+        # log the message in the server logs
+        self.log.info(f"Log from run_id={run_id}, task_id={task_id}: {log_message}")
+        for room in session.rooms:
+            emit("algorithm_log", data, room=room)
```
bartvanb, line 445: This message should not be emitted in all rooms: some rooms may belong to a different collaboration and sending this logging info to that room could be considered a data leak.

In this case, we want to emit the log messages to a room where users can listen to them (via the UI). There are two types of rooms users have access to: `collaboration_<id>` and `collaboration_<id>_organization_<id>`. I've briefly looked into this and it seems the latter is hardly used: that only gives users updates when a node from their own organization goes online/offline. All default roles give users also access to the `collaboration_<id>` rooms in which case the latter room is useless. I think it therefore makes sense to remove this room - I'll create an issue.

So the proper room to send this message in is the room `collaboration_<collab_id>`. That room contains only users that are allowed to see events happening in that collaboration so you ensure only the proper users can see this.

bartvanb (commented):
Looks pretty neat and clean. Just a comment about the socket rooms

PR #1767, opened 2025-02-27
Temporary disabled the settable host alias and bind address

frankcorneliusmartin:
Sorry @hcadavid I disabled these for now.


hcadavid (approved):
Approved.
If this makes something fail on MacOS, it's better to remove it and focus first on finding a configuration that works, locally-only, on WLS, Linux and MacOS.

bartvanb (approved):

PR #1771, opened 2025-03-03
feat: Enhance server configuration with more flexible options

frankcorneliusmartin:
- Add configurable server description and IP
- Introduce optional external database and RabbitMQ URI configurations
- Improve logging configuration with more granular settings
- Add support for dynamic or static JWT secret generation
- Update server backend configmap to handle new configuration options

### DIFF
--- a/charts/server/templates/server/server-backend-configmap.yml
+++ b/charts/server/templates/server/server-backend-configmap.yml
@@ -7,9 +7,9 @@ data:
   config.yaml: |
     algorithm_stores: []
     {{- if .Values.server.allow_drop_all }}
-      allow_drop_all: {{ .Values.server.allow_drop_all }}
+    allow_drop_all: {{ .Values.server.allow_drop_all }}
     {{- else }}
-      allow_drop_all: False
+    allow_drop_all: False
     {{- end }}
     api_path: {{ .Values.server.apiPath }}
     description: {{ .Values.server.description }}
@@ -19,12 +19,12 @@ data:
     {{- end }}
     logging:
       backup_count: {{ .Values.server.logging.backup_count }}
-      datefmt: {{ .Values.server.logging.datefmt }}
+      datefmt: {{ .Values.server.logging.datefmt | quote }}
       file: {{ .Values.server.logging.file }}
-      format: {{ .Values.server.logging.format }}
+      format: {{ .Values.server.logging.format | quote }}
       level: {{ .Values.server.logging.level }}
       loggers:
-      {{- range .Values.logging.loggers }}
+      {{- range .Values.server.logging.loggers }}
       - level: {{ .level }}
         name: {{ .name }}
       {{- end }}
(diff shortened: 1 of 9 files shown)


charts/server/values.yaml
```diff
@@ -13,14 +18,50 @@ server:
   # The number of replicas of the server
   replications: 1
 
+  # In production it is recommended to use an external database (outside of k8s).
+  # By default, a postgres database is created inside the k8s cluster.
+  # If you want to use an external database, you can set the database_uri here.
+  # database_uri: postgresql://user:password@localhost:5432/mydatabase
```
bartvanb, line 24: Just to check: this replaces whitelisting? Awesome...

charts/server/values.yaml
```diff
@@ -13,14 +18,50 @@ server:
...
+      name: engineio.server
+    - level: warning
+      name: sqlalchemy.engine
+    - level: warning
+      name: requests_oauthlib.oauth2_session
+
+  # In case you want to use a static secret key for the JWT token. Setting this is not
+  # recommended for production environments.
+  jwt:
+    secret: "super-secret-key-that-you-shouldnt-use-in-production"
```
bartvanb, line 60: If we're going to put this in this values.yaml, we should make sure to handle in a to-be-implemented CLI that this value is not used by default. If this is used by default, I'm sure that people *are* going to use this in production

charts/server/values.yaml
```diff
@@ -13,14 +18,50 @@ server:
...
+
+  dev:
+    host_uri: host.docker.internal
+    drop_all: 'True'
```
bartvanb, line 64: Why is this `'True'` and in line 39 there is `true`?

charts/server/values.yaml
frankcorneliusmartin: I will comment this out and add it to the `devspace.yaml` that way it cannot be accidentally be used in production.

charts/server/values.yaml
frankcorneliusmartin: Ehm sorry no. I get the confusion. External means not within k8s. And this is the server instance, so no whitelisting needed. Ill update the comment to avoid this confusion.

charts/server/templates/server/server-backend-configmap.yml
```diff
@@ -6,39 +6,49 @@ metadata:
...
+    {{- if .Values.server.allow_drop_all }}
+    allow_drop_all: {{ .Values.server.allow_drop_all }}
+    {{- else }}
+    allow_drop_all: False
+    {{- end }}
     api_path: {{ .Values.server.apiPath }}
-    description: Vantage6 server version 5
-    ip: 0.0.0.0
+    description: {{ .Values.server.description }}
+    ip: {{ .Values.server.ip }}
```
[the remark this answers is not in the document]
frankcorneliusmartin: Yes this is really only relevant when not running in a container, which we no longer support anyways. Ill revert it to 0.0.0.0.

charts/server/values.yaml
frankcorneliusmartin: This is how it was in the original code, so I did not want to change it. Same goes for the store template. 

I did a little digging and the code for both the server and store reads:

```python
ctx = ServerContext.from_external_config_file(config, system_folders)
allow_drop_all = ctx.config["allow_drop_all"]
Database().connect(uri=ctx.get_database_uri(), allow_drop_all=allow_drop_all)
```
After this no conversion is made. So the translation should happen in the creation of the server context. I checked this: 

```python
>>> ctx = ServerContext("testing", system_folders=False)
>>> ctx.config["allow_drop_all"]
'False'
>>> bool(ctx.config["allow_drop_all"])
```

So this is a bug that is also in 4.10. We should make a separate issue for this and address it at all places.

frankcorneliusmartin: Reported here: https://github.com/vantage6/vantage6/issues/1774

bartvanb: Added that issue to v4.10 todo list

charts/server/values.yaml
bartvanb: I see that you uncommented it but I don't see it added to `devspace.yaml`

frankcorneliusmartin: It was already there!

PR #1781, opened 2025-03-04
Using API path in setting auth path for JWT

bartvanb:
Not entirely sure about this, I guess need to check some more what this parameter is about

### DIFF
--- a/vantage6-server/vantage6/server/__init__.py
+++ b/vantage6-server/vantage6/server/__init__.py
@@ -103,7 +103,7 @@ class ServerApp:
             static_folder=Path(__file__).parent / "static",
         )
         self.debug: dict = self.ctx.config.get("debug", {})
-        self.configure_flask()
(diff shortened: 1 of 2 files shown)


frankcorneliusmartin (approved):
Was there a bug for this? It seems like this never caused any issues, any particular reason you start fixing it now? The code seems OK to me.

vantage6-server/vantage6/server/__init__.py
```diff
@@ -273,14 +273,23 @@ def setup_socket_connection(self) -> SocketIO:
 
         return socketio
 
-    def configure_flask(self) -> None:
-        """Configure the Flask settings of the vantage6 server."""
+    def configure_flask(self, api_path) -> None:
```
frankcorneliusmartin, line 276: type missing for `api_path`

bartvanb:
I just stumbled across this error so I fixed it. Not entirely sure about the consequences (they must have been limited) but it's obvious it was wrong

PR #1807, opened 2025-03-12
[Bug] "TypeError: can't subtract offset-naive and offset-aware datetimes"

riccardosamperna:
When using the UI to login, the server would throw the following error:
`TypeError: can't subtract offset-naive and offset-aware datetimes`
This PR fixes the problem by parsing the `self.last_login_attempt` to UTC.


vantage6-server/vantage6/server/model/user.py
```diff
@@ -181,9 +181,11 @@ def is_blocked(
         int | None
             How many minutes user is still blocked for
         """
+        from vantage6.server.resource import parse_datetime
```
bartvanb, line 184: could you move this import to the top of the file? After that feel free to merge

bartvanb (approved):

vantage6-server/vantage6/server/model/user.py
bartvanb: Ah right, in the `model` folder we shouldn't import from the `resource` folder. Maybe we can put the parse_datetime function somewhere else, it doesn't really only relate to resources anyway. I guess `vantage6/server/utils.py` makes sense

bartvanb (approved):

PR #1808, opened 2025-03-12
Feature/#1680 settable k8s task namespace

bartvanb:
Main feats: 
- settable task namespace
- extraction tasks produce a parquet file

What needs to be changed, is that logs from extraction tasks are currently not saved to the server: there is still a bug in that so it's disabled

### DIFF
--- a/vantage6-common/vantage6/common/globals.py
+++ b/vantage6-common/vantage6/common/globals.py
@@ -22,6 +22,8 @@ DEFAULT_UI_IMAGE = f"infrastructure/ui:{MAIN_VERSION_NAME}"
 
 DEFAULT_ALGO_STORE_IMAGE = f"infrastructure/algorithm-store:{MAIN_VERSION_NAME}"
 
+DEFAULT_ALPINE_IMAGE = f"{DEFAULT_DOCKER_REGISTRY}/infrastructure/alpine:latest"
+
 #
 #   COMMON GLOBALS
 #
--- a/vantage6-common/vantage6/common/globals.py
+++ b/vantage6-common/vantage6/common/globals.py
@@ -22,7 +22,7 @@ DEFAULT_UI_IMAGE = f"infrastructure/ui:{MAIN_VERSION_NAME}"
 
 DEFAULT_ALGO_STORE_IMAGE = f"infrastructure/algorithm-store:{MAIN_VERSION_NAME}"
 
-DEFAULT_ALPINE_IMAGE = f"{DEFAULT_DOCKER_REGISTRY}/infrastructure/alpine:latest"
+DEFAULT_ALPINE_IMAGE = "infrastructure/alpine:latest"
 
 #
 #   COMMON GLOBALS
--- a/vantage6-node/vantage6/node/__init__.py
+++ b/vantage6-node/vantage6/node/__init__.py
@@ -222,7 +222,6 @@ class Node:
         task_results = self.client.run.list(
             state=TaskStatusQueryOptions.OPEN.value, include_task=True
         )
-        self.log.debug("task_results: %s", task_results)
 
         # add the tasks to the queue
         self.__add_tasks_to_queue(task_results)
@@ -442,7 +441,8 @@ class Node:
                     data={
                         "result": results.data,
                         # TODO (HC) check if only logs[0] are enough
-                        "log": results.logs[0],
+                        # TODO include logs
+                        # "log": results.logs[0],
                         "status": results.status,
                         "finished_at": datetime.datetime.now().isoformat(),
                     },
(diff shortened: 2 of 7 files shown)


vantage6-common/vantage6/common/globals.py
```diff
@@ -22,6 +22,8 @@
 
 DEFAULT_ALGO_STORE_IMAGE = f"infrastructure/algorithm-store:{MAIN_VERSION_NAME}"
 
+DEFAULT_ALPINE_IMAGE = f"{DEFAULT_DOCKER_REGISTRY}/infrastructure/alpine:latest"
```
frankcorneliusmartin, line 25: Why did you include the `DEFAULT_DOCKER_REGISTRY`? We dont seem to do that for the `DEFAULT_ALGO_STORE_IMAGE`, better to be consistent?

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -91,6 +93,79 @@ def __init__(self, ctx: NodeContext, client: NodeClient):
...
+
+        Returns
+        -------
+        bool
+            True if the namespace exists or was created and is writeable, False
+            otherwise
+        """
+        # TODO consider using hierarchical namespaces, it may be easier to create
+        # those in a large cluster environment
+        # https://kubernetes.io/blog/2020/08/14/introducing-hierarchical-namespaces/
```
frankcorneliusmartin, line 112: Is there already an issue in our tracker for this?

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -313,7 +390,10 @@ def run(
             spec=k8s_client.V1JobSpec(
                 template=k8s_client.V1PodTemplateSpec(
                     metadata=k8s_client.V1ObjectMeta(
-                        labels={"app": run_io.container_name}
+                        labels={
+                            "app": run_io.container_name,
+                            "node_name": self.ctx.name,
```
frankcorneliusmartin, line 395: There is no guarantee this name is unique. There could be nodes from the same server or different servers having the samen name / id. Why not simply use a generated uuid or use the existing API key? 

(You are using this name to identify which job containers belong to this node right?)

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -988,10 +1075,10 @@ def process_next_completed_job(self) -> Result:
         completed_job = False
         while not completed_job:
 
-            # Get all jobs from the vantage6-node namespace
-            jobs = self.batch_api.list_namespaced_job("vantage6-node")
+            # Get all jobs from the task namespace for this node
+            jobs = self.batch_api.list_namespaced_job(self.task_namespace)
             if not jobs.items:
-                time.sleep(1)
+                time.sleep(10)
```
frankcorneliusmartin, line 1081: Dont you think this is a bit excessive? Or is this under the assumption that a container always takes at least 10 seconds? Maybe lower it to 5?

vantage6-common/vantage6/common/globals.py
bartvanb: Changed to be consistent

vantage6-node/vantage6/node/__init__.py
```diff
@@ -434,7 +441,8 @@ def __poll_task_results(self) -> None:
                     data={
                         "result": results.data,
                         # TODO (HC) check if only logs[0] are enough
-                        "log": results.logs[0],
+                        # TODO include logs
+                        # "log": results.logs[0],
```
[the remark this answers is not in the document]
bartvanb: This was not working, I just added commit to get it working

vantage6-node/vantage6/node/k8s/container_manager.py
bartvanb: https://github.com/vantage6/vantage6/issues/1815 just created

vantage6-node/vantage6/node/k8s/container_manager.py
bartvanb: Agree. This label is not really used yet (at least it doesn't work). I'll work now on adapting this label. I think it's best to use the API key to generate an identifier (so that API key is not visible anywhere in plain text).

bartvanb: I removed the node_name label again. The `run_io.container_name` is already unique (it bases on run_id) so the extra label wasn't necessary here.

Instead, I added a node identifier label to the *jobs*. Jobs did not have labels and by attaching this label to it, we allow the node to figure out when a job finishes that it was their job

frankcorneliusmartin: Perfect.

PR #1814, opened 2025-03-13
Feature/#253 Improved node diagnostic tools

riccardosamperna:
This PR implements a node diagnostic strategy making use of [Prometheus](https://prometheus.io). Prometheus is used in a non-standard way because the current vantage6 setup doesn't allow for a pull mechanism from nodes, the server doesn't know the IP addresses of the nodes. Nodes are responsible for updating the server about resource usage similar to what is done for logging and task updates. 

A custom prometheus exporter running alongside the vantage6 server consumes the messages received by the nodes and updates Prometheus metrics.

This solution was the best compromise considering the current setup of vantage6.

### DIFF
--- a/vantage6/vantage6/cli/prometheus/prometheus.yml
+++ b/vantage6/vantage6/cli/prometheus/prometheus.yml
@@ -1,4 +1,5 @@
 global:
-  scrape_interval: 15s
+  # How frequently to scrape targets by default.
+  scrape_interval: 1m
 
 scrape_configs: []
--- a/docs/node/node_config.yaml
+++ b/docs/node/node_config.yaml
@@ -289,3 +289,7 @@ task_dir: C:\Users\<your-user>\AppData\Local\vantage6\node\mydir
 # for other organizations in your collaboration to understand why a task
 # is not completed. Obviously, no sensitive data is shared. Default true
 share_config: true
+
+# Interval (in seconds) at which the node sends system metadata to the server.
+# This should align with the Prometheus scrape_interval to avoid stale data.
+report_interval_seconds: 45
--- a/vantage6-node/vantage6/node/__init__.py
+++ b/vantage6-node/vantage6/node/__init__.py
@@ -210,13 +210,15 @@ class Node:
         """
         Periodically send system metadata to the server.
         """
+        report_interval = self.config.get("report_interval_seconds", 45)
+
         while True:
             try:
                 metadata = self.__gather_system_metadata()
                 self.socketIO.emit("node_metrics_update", metadata, namespace="/tasks")
             except Exception:
                 self.log.exception("Metadata thread had an exception")
-            time.sleep(PING_INTERVAL_SECONDS)
+            time.sleep(report_interval)
 
     def __gather_system_metadata(self) -> dict:
         """


vantage6/vantage6/cli/prometheus/monitoring_manager.py
```diff
@@ -0,0 +1,117 @@
...
+
+            info(f"Prometheus configuration updated with target: {server_address}")
+
+        except Exception as e:
+            error(f"Failed to update Prometheus configuration: {e}")
+            raise
+
+    def _get_container(self):
+        """
+        Check if a Prometheus container is already running
```
frankcorneliusmartin, line 112: This docstring does not seem to be correct.

docs/server/yaml/server_config.yaml
```diff
@@ -205,3 +205,12 @@ dev:
...
   host_uri: http://host.docker.internal
+
+# Settings for the prometheus monitoring
+prometheus:
+  # Set to true to start Prometheus server when starting the server
+  start_with_server: false
+  # Set to true to enable the prometheus monitoring
+  enabled: true
+  # Port at which the prometheus exporter will be available on your local machine
+  exporter_port: 9100
```
frankcorneliusmartin, line 216: Can you be a bit more elaborate on the settings?

* `start_with_server` is clear
* `enabled` if is already starts with the server, what is enabled in this case the mechanism that the node are able to report to the server? (I also have not found a reference in the code to it? Did I miss it?)
* which service eposes the `exporter_port` and whom consumes it?

vantage6-server/vantage6/server/__init__.py
```diff
@@ -184,6 +185,8 @@ def __init__(self, ctx: ServerContext) -> None:
         t = Thread(target=self.__node_status_worker, daemon=True)
         t.start()
 
+        start_prometheus_exporter(port=self.ctx.config.get("prometheus_port", 8000))
```
frankcorneliusmartin, line 188: In vantage6 we typically use the 76xx range for ports of our services:

7600 --> UI
7601 --> Server API
7602 --> Store API 

(76 comes from V+A+N+T+A+G+E+6 if you index the letters 1-26)

I suggest to
* use `7603` (--> Server prometheus exporter API).
* Add this default port in the `globals.py`

vantage6/vantage6/cli/globals.py
```diff
@@ -42,6 +42,8 @@
 # Address of community algorithm store
 COMMUNITY_STORE = "https://store.cotopaxi.vantage6.ai/api"
 
+PROMETHEUS_CONTAINER_NAME = f"{APPNAME}-prometheus"
```
frankcorneliusmartin, line 45: To be consistent could you instead put this in the ServerContext?: https://github.com/vantage6/vantage6/blob/main/vantage6/vantage6/cli/context/server.py

The naming convention in this `ctx` prevents naming conflicts with other servers on the same machine.

vantage6/vantage6/cli/prometheus/prometheus.yml
```diff
@@ -0,0 +1,4 @@
+global:
+  scrape_interval: 15s
```
frankcorneliusmartin, line 2: Can we make this settable?

frankcorneliusmartin: This is the prometheus instance scraping the server interval?

frankcorneliusmartin (changes requested):
* Could you also add documentation in the /docs?
* I am OK with having only a comment as docstring for 'private' functions

vantage6/vantage6/cli/prometheus/monitoring_manager.py
```diff
@@ -0,0 +1,117 @@
+import docker
+from pathlib import Path
+from vantage6.cli.globals import PROMETHEUS_CONTAINER_NAME
+from vantage6.common import info, error
+import yaml
```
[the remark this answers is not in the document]
frankcorneliusmartin: I am not sure wether `Black` also does this import order thing?

vantage6/vantage6/cli/prometheus/monitoring_manager.py
```diff
@@ -0,0 +1,117 @@
...
+        self._prepare_config()
+
+        volumes = {
+            str(self.config_file): {
+                "bind": "/etc/prometheus/prometheus.yml",
+                "mode": "ro",
+            },
+            str(self.data_dir): {"bind": "/prometheus", "mode": "rw"},
+        }
+        ports = {"9090/tcp": 9090}
```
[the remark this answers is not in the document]
frankcorneliusmartin: Can you create an issue for that? Then we can close this.

vantage6/vantage6/cli/prometheus/monitoring_manager.py
frankcorneliusmartin: Well its a bit of an explicit check. If I read the code I would say; if the prometheus container exists get the details. Or this function should return a boolean. Is that clearer?

vantage6/vantage6/cli/globals.py
frankcorneliusmartin: I think you made an error in the commit. You've used `self.ctx.docker_container_name` but I think it should be `self.ctx.prometheus_container_name`.

frankcorneliusmartin (approved):

PR #1817, opened 2025-03-13
Enable logging for the node to the console

frankcorneliusmartin:
* Also includes a change to make the server db mounts settable (so we avoid conflicts)
* Review this first, and merge. Then the checks ... should ... pass: https://github.com/vantage6/vantage6/pull/1818


devspace.yaml
```diff
@@ -21,6 +21,13 @@ vars:
   NUMBER_OF_NODES:
     question: "How many nodes do you want to create for your development environment?"
     default: 2
+  SERVER_DATABASE_MOUNT_PATH:
+    question: "Where is your server database located on the host machine?"
+    default: ${PWD}/dev/data/db_pv_server
+  STORE_DATABASE_MOUNT_PATH:
+    question: "Where is your store database located on the host machine?"
+    default: ${PWD}/dev/data/db_pv_store
```
bartvanb, line 29: I just wanted to note that for running WSL, you will have to provide the WSL path, which will still not be persisted after WSL restarts. Maybe we can resolve this in a CLI command (although maybe it makes more sense to have regular users running a local dev env to do it in their regular windows env)

vantage6/vantage6/cli/dev/create.py
```diff
@@ -10,7 +10,7 @@
 from vantage6.common.globals import APPNAME, InstanceType, Ports
 from vantage6.common import info, error, generate_apikey
 
-import vantage6.cli.dev.data as data_dir
+# import vantage6.cli.dev.data as data_dir
```
bartvanb, line 13: This is not really a solution for a no-longer-existing dir :-) But since this command is not working anyway in v5, I guess it's fine (although not sure why you would comment it out)

vantage6/vantage6/cli/server/attach.py
```diff
@@ -1,63 +1,24 @@
...
+        "devspace",
+        "logs",
+        "--follow",
+        "--label-selector",
+        "app=vantage6-server",
+        "--label-selector",
+        "component=vantage6-server",
+    ]
+    process = Popen(command, stdout=None, stderr=None)
+    process.wait()
```
bartvanb, line 24: I'm missing the store in this PR. Also, it would be nicer to further refactor this in a common function so that we can just do `open_logs(my_label_selector)`

vantage6/vantage6/cli/server/attach.py
```diff
@@ -1,63 +1,24 @@
...
+    info("Attaching to server logs...")
+
+    command = [
+        "devspace",
+        "logs",
+        "--follow",
+        "--label-selector",
+        "app=vantage6-server",
+        "--label-selector",
+        "component=vantage6-server",
```
bartvanb, line 21: You can combine these label selectors to a single one like this: `app=vantage6-server,component=vantage6-server`

bartvanb (changes requested):
Some small changes.

Also, I'm thinking that we could also create CLI commands to get the logs of the databases, UI etc. However, I don't think they're useful so maybe let's not do that now.

devspace.yaml
frankcorneliusmartin: I added this comment to the v6 CLI command issue to remind us that this should be fixed by the CLI.

vantage6/vantage6/cli/dev/create.py
frankcorneliusmartin: So this has been removed in the other PR. So this is no longer relevant.

vantage6/vantage6/cli/server/attach.py
frankcorneliusmartin: Good suggestion, updated it to use a common function.

vantage6/vantage6/cli/server/attach.py
frankcorneliusmartin: Perfect, thanks.

bartvanb (commented):
Code looks good to me but unit tests are failing for the attach commands now. Can you adapt those?

bartvanb (approved):

PR #1818, opened 2025-03-13
Change/remove dev commands

frankcorneliusmartin:
* These keep breaking the unittests 
* Integration tests need to be fixed with a new environment


bartvanb (approved):
LGTM. One question: I assume you deleted the `dev` commands code because the new version will be very dissimilar, and you kept the `test integration-test` because that will likely be more similar?

frankcorneliusmartin:
> LGTM. One question: I assume you deleted the `dev` commands code because the new version will be very dissimilar, and you kept the `test integration-test` because that will likely be more similar?

Yes indeed.

PR #1819, opened 2025-03-14
Fix duplicated node logging in v5 dev env

bartvanb:
Logs were duplicated as logs were written in several threads and probably due to logger singletons this led to multiple copies of each log statement


frankcorneliusmartin (approved):
I comment on these changes in another PR: https://github.com/vantage6/vantage6/pull/1820#issue-2920151741

Issue #809, opened 2023-09-05
[Change Request] Method name should not be part of the algorithm `input` field


bartvanb:
**Description**
The `input` field is for data that will be encrypted. It would be nice to move the algorithm method name out of this field so that we can show it easily to the user without needing to decrypt it.

bartvanb:
I'm moving this to v5.0 because that is getting closer, and it is easier to do it there because we do not need to take backwards compatibility into account

PR #1820, opened 2025-03-14
Add option to set session step type (data extract/preprocess/compute) for algorithm functions

bartvanb:
Note that to aid in development, the commits from the PR on duplicate logging is also in this PR


frankcorneliusmartin (changes requested):
Now we make the distinction between the enum in the store containing `Federated` and `Central`. However I feel we can derive these of the `AlgorithmStepType` if we extend it a little:

`DATA_EXTRACTION`, `PREPROCESSING`, `FEDERATED` (=compute) `CENTRAL` (=compute) and `POST_PROCESSING`. Or do you need the *two types* for the UI? In that case these types might be derived from this `vantage6-common/vantage6/common/enum.py` enum.

bartvanb:
> Now we make the distinction between the enum in the store containing Federated and Central. However I feel we can derive these of the AlgorithmStepType if we extend it a little:
DATA_EXTRACTION, PREPROCESSING, FEDERATED (=compute) CENTRAL (=compute) and POST_PROCESSING. Or do you need the two types for the UI? In that case these types might be derived from this vantage6-common/vantage6/common/enum.py enum.

I think this is a good idea. There is however one catch: what if the user submits a task from an algorithm that is not in the store, how do we know if it's a federated or central task? We could ask for this extra input via the python client, or we could introduce also a generic 'compute' type where this is treated as unknown.
I feel most for asking for extra input: most people will use the UI anyway and then it's just derived from there. Note that if you provide an algorithm store, you also don't have to provide this extra info as it will be in the store.

bartvanb:
I'm picking the merging up in a separate branch, change/merge-federated-central-type-with-step-type. It's more work than anticipated and it makes sense to pick it up together with #809 because we can then use task input to extract the type of function from the store at the server.

Issue #1686, opened 2025-01-16
[Feature Request] [k8s] Document the new deployment 


frankcorneliusmartin:
**Problem description**
Now that we changed to our prefered way of deploying the infra in k8s we need to update our documentation accordingly. ~~Also the development environment has been changed quite a bit so we should also update those sections in the documentation.~~

Collected items (also from comments below):
- [ ] Update requirements for vantage6
- [x] Update v6 dev -> v6 sandbox
- [ ] Mention that on WSL deletes stuff on reboot. Potentially delete the message from devspace.yaml where it is mentioned in several places [comment](https://github.com/vantage6/vantage6/pull/1828#discussion_r2009805297)
- [ ] https://github.com/vantage6/vantage6/pull/1884#discussion_r2057898805: Add doc on how to find host IP address and refer to it in the devspace.yaml
- [ ] Check if any other docs require updates

frankcorneliusmartin:
https://github.com/vantage6/vantage6/pull/1828#discussion_r2009805297

bartvanb:
I'm moving the documentation of the dev env to a separate subissue of the dev env overview issue - that seems more structured. I'll cross out references to it here

CunliangGeng:
https://github.com/vantage6/vantage6/pull/1884#discussion_r2057898805: Add doc on how to find host IP address and refer to it in the `devspace.yaml`.

bartvanb:
All these points seem sufficiently handled now, the only thing left is to document the ingress/load balancing but there is a separate issue for that

PR #1828, opened 2025-03-19
Feature/basic sessions algo in local dev store

frankcorneliusmartin:


dev/load-basic-fixtures.py
```diff
@@ -149,4 +156,4 @@ def create_node(index, collaboration, organization):
 
 print("=> Creating nodes")
 for i in range(number_of_nodes):
-    create_node(i, col_1, organizations[i])
+    create_node(i, col_1, organizations[i], task_namespace)
```
bartvanb, line 159: Now that I look at this anyway, can we do `i+1` so that we create `node_1` as first node instead of `node_0`? Same for organization and user (and maybe I'm forgetting more?)

devspace.yaml
```diff
@@ -21,14 +21,16 @@ vars:
   NUMBER_OF_NODES:
     question: "How many nodes do you want to create for your development environment?"
     default: 2
+  TASK_NAMESPACE:
+    question: "What is the task namespace for your development environment?"
+    default: vantage6-tasks
```
bartvanb, line 26: I really wonder if we need to bother people with asking this - for local development environment, people will think this is OK right? Maybe we can catch this in the CLI instead? Then people can still change it with a flag but they don't have to answer the question necessarily

bartvanb (changes requested):
Small changes, also due to developments in other PRs

frankcorneliusmartin:
> Pretty difficult to review the PR for me, it is way out of my known context.

I can imagine. Lets get v5 running on your laptop soonish.

dev/connect-store.py
```diff
@@ -17,48 +17,96 @@
...
+    code_url="https://github.com/vantage6-ai/v6-session-basics",
+    partitioning="horizontal",
+    functions=[
+        {
+            "name": "read_csv",
+            "display_name": "Read CSV file",
+            "standalone": True,
+            "description": "",
+            "ui_visualizations": [],
+            "type": "federated",
```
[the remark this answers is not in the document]
frankcorneliusmartin: Done

devspace.yaml
frankcorneliusmartin: Hmm fair, they can now still overwrite this using the `--var TASK_NAMESPACE [namespace]` flag when starting any `devspace` command.

dev/load-basic-fixtures.py
frankcorneliusmartin: I used @riccardosamperna suggestion.

bartvanb: Agreed, that's better

devspace.yaml
```diff
@@ -152,17 +150,14 @@ deployments:
...
-        # WSL
         #   volumePath: /run/desktop/mnt/host/wsl/[SOME_PATH]
-        # MacOS
-        #   volumePath: ???
-        # Linux
-        #  users can probably use the same path as in the example below. Or any other
-        #  path that is accessible by the Docker daemon.
+        #
+        # This creates a mount to the `/mnt/wsl/[SOME_PATH]` directory. Note that this
+        # directory is emptied when the WSL2 instance is restarted.
```
bartvanb, line 160: Maybe just a note that this should be available in the docs, and then we can remove this from the devspace.yaml (I think it's there in multiple places)

bartvanb (approved):

devspace.yaml
frankcorneliusmartin: I added this comment to the v5 docs issue: https://github.com/vantage6/vantage6/issues/1686

PR #1830, opened 2025-03-19
Tmp/test unit test fails

bartvanb:
This is just a PR to test why unit tests fail remotely but not locally - do not merge


frankcorneliusmartin:
You can also open these things as draft? Or are you then missing something?

bartvanb:
Yeah unit tests were not running when I was opening them as draft

PR #1855, opened 2025-03-28
Feature/#1363 Optimize `wait_for_results()` by implementing task status endpoint

riccardosamperna:
This PR implements a task status endpoint that can be queried instead of retrieving the whole task. This should hopefully optimize the `wait_for_results()` reducing some load on the server. There is a lot of refactoring that should be done in the rest of the `Task` endpoint but I didn't pick it up in this issue.

### DIFF
--- a/vantage6-common/vantage6/common/client/client_base.py
+++ b/vantage6-common/vantage6/common/client/client_base.py
@@ -15,6 +15,56 @@ from vantage6.common.task_status import has_task_finished
 module_name = __name__.split(".")[1]
 
 
+@staticmethod
+def _log_completion(task_id: int, start_time: float, log_animation: bool) -> None:
+    """
+    Log the completion message for a task.
+
+    Parameters
+    ----------
+    task_id : int
+        ID of the completed task.
+    start_time : float
+        The time when the task started.
+    log_animation : bool
+        Whether to log the message as an animation or a regular log.
+    """
+    elapsed_time = int(time.time() - start_time)
+    message = f"Task {task_id} completed in {elapsed_time} seconds."
+
+    if log_animation:
+        print(f"\r{message}                     ")
+    else:
+        logging.info(message)
+
+
+@staticmethod
+def _log_progress(
+    task_id: int, start_time: float, log_animation: bool, animation_frame: str
+) -> None:
+    """
+    Log the progress message for a task.
+
+    Parameters
+    ----------
+    task_id : int
+        ID of the task in progress.
+    start_time : float
+        The time when the task started.
+    log_animation : bool
+        Whether to log the message as an animation or a regular log.
+    animation_frame : str
+        The current frame of the animation.
+    """
+    elapsed_time = int(time.time() - start_time)
+    message = f"{animation_frame} Waiting for task {task_id}... ({elapsed_time}s)"
+
+    if log_animation:
+        print(f"\r{message}", end="")
+    else:
+        logging.info(message)
+
+
 class ClientBase(object):
     """Common interface to the central server.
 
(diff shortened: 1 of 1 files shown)


vantage6-common/vantage6/common/task_status.py
```diff
@@ -1,4 +1,9 @@
 from enum import Enum
+import itertools
+import sys
+import time
+
+MAX_INTERVAL = 300
```
bartvanb, line 6: Current practice is that we put all constants in the file `globals.py` so that we know where to find constants. However I expect this constant will only be used in this file so I'm fine with either. I think it's actually better practice to define it here. @frankcorneliusmartin are you OK with changing that practice?

vantage6-common/vantage6/common/task_status.py
```diff
@@ -66,3 +71,39 @@ def has_task_finished(status: TaskStatus) -> bool:
...
+            break
+
+        frame = next(animation)
+        sys.stdout.write(
+            f"\r{frame} Waiting for task {task_id} ({int(time.time()-t)}s)"
+        )
+        sys.stdout.flush()
+
+        time.sleep(interval)
+        interval = min(interval * 1.5, MAX_INTERVAL)
```
bartvanb, line 109: Could you make the `1.5` a constant variable as well, like MAX_INTERVAL?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -142,13 +142,7 @@ def wait_for_results(self, task_id: int, interval: float = 1) -> list:
...
             List of task results.
         """
-        status = self.task.get(task_id).get("status")
-        while not has_task_finished(status):
-            info(f"Waiting for results of task {task_id}...")
-            time.sleep(interval)
-            status = self.task.get(task_id).get("status")
-        info("Done!")
-
+        wait_for_task_completion(self.request, task_id, interval)
```
bartvanb, line 145: This function uses the itertools animation to print findings, instead of a new line every time as in the current implementation. Have you tried running an algorithm with this new function and do the logs look OK then? How does this work with the live logging feature we released in 4.10.2? Is that updated every time the animation changes between '/|\-' signs? That might put a bit more strain on the socket than necessary

(Note that to test this, you need to build a new algorithm docker image including this change to the AlgorithmClient, Frank or me can help with that)

vantage6-server/vantage6/server/resource/task.py
```diff
@@ -1219,3 +1226,87 @@ def _delete_subtasks(task: db.Task) -> None:
             Task._delete_subtasks(child_task)
             log.info(f" Removing child task id={child_task.id}")
             child_task.delete()
+
+
+class TaskStatusEndpoint(TaskBase):
+    """Resource for /api/task/<id>/status"""
+
+    @only_for(("user", "node", "container"))
```
bartvanb, line 1234: I don't think a node will ever access this endpoint. At least in the present implementation it doesn't. So we could consider not allowing the node to access this endpoint. On the other hand, this endpoint shows merely a subset of what the get /task endpoint shows and the node *is* allowed to see that, so there is also no real reason to forbid it to see this.

bartvanb (commented):
Looks pretty good, a few minor comments. Most importantly, I have a question regarding how the algorithm logs work now.

vantage6-common/vantage6/common/task_status.py
frankcorneliusmartin: I agree with this, however we already have a lot of constants in the `globals.py` which are used only in single files (I think). How strickt are we going to be?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: I think by removing the animation entirely, you are not just removing it in the algorithm logs, but also when users call `client.wait_for_results` via the user client. In that case the logs don't go over the socket and the animation IS desirable (we introduced it because we used to have many lines in jupyter notebooks saying 'waiting for results.... waiting for results....' etc.
Maybe the easiest solution is to extend `wait_for_task_completion` with an arg `log_animation` (default True)? Then we support both use cases

vantage6-server/vantage6/server/resource/task.py
bartvanb: Let's remove the node access. We have no use case for it so I guess better to only allow what is necessary

vantage6-common/vantage6/common/task_status.py
```diff
@@ -66,3 +72,31 @@ def has_task_finished(status: TaskStatus) -> bool:
...
+        response = request_func(f"task/{task_id}/status")
+        status = response.get("status")
+
+        if has_task_finished(status):
+            logging.info(f"Task {task_id} completed in {int(time.time() - t)} seconds.")
+            break
+
+        logging.info(f"Waiting for task {task_id}... ({int(time.time() - t)}s)")
+        time.sleep(interval)
+        interval = min(interval * INTERVAL_MULTIPLIER, MAX_INTERVAL)
```
bartvanb, line 102: I think this function is only used inside the `AlgortihmClient` and the `UserClient` (aka `Client`). Therefore, I suggest putting this function in the BaseClient

bartvanb (commented):
I ended up completely redoing the review as I didn't remember too much any more. I resolved some earlier comments, answered the others, and added one more comment.

vantage6-common/vantage6/common/client/client_base.py
```diff
@@ -581,6 +583,89 @@ def __check_algorithm_store_valid(self, is_for_algorithm_store: bool) -> bool:
...
+
+            if has_task_finished(status):
+                self._log_completion(task_id, start_time, log_animation)
+                break
+
+            self._log_progress(task_id, start_time, log_animation, next(animation))
+            time.sleep(interval)
+            interval = min(interval * INTERVAL_MULTIPLIER, MAX_INTERVAL)
+
+    def _log_completion(task_id: int, start_time: float, log_animation: bool) -> None:
```
bartvanb, line 623: you're missing a `self` here and in `_log_progress`. Best I think is to make this a `@staticmethod`

bartvanb (approved):
Feel free to merge after resolving the codacy issues (i.e. add @staticmethod twice)

PR #1856, opened 2025-03-31
Move task status queries to models and use expressions

bartvanb:
I haven't tested this code yet so we should wait to merge until it is tested


bartvanb:
Closing this as it is now part of #1861

PR #1861, opened 2025-04-01
UI: dataframe preprocessing, and several bugfixes and other changes

### LINKED ISSUE #1533: It would be nice if you could move this function to the model and then filter on `db.Task.waiting()` or so
It would be nice if you could move this function to the model and then filter on `db.Task.waiting()` or so

_Originally posted by @bartvanb in https://github.com/vantage6/vantage6/pull/1247#discussion_r1726894478_

bartvanb:
backend:
- bugfix preprocessing tasks (filter on task waiting status did not go well with sqlalchemy 2.0 changes)
  - included a refactor that fixes #1533 

fronend:
- added option to do preprocessing tasks
- retrieve columns from dataframes when creating tasks
- when creating compute tasks, only show compute functions
- Added pagination for table with available dataframe columns 
- added option to delete dataframes via UI
- Cleaned up old 'preprocessing' and 'filter' steps in the generic create task component


frankcorneliusmartin (approved):
Also a nice refactor

PR #1873, opened 2025-04-02
Adapt path of socket.io to prevent socket 404 errors. 

bartvanb:
Reported by Narasimha Raghavan


frankcorneliusmartin:
I've seen it work also without the `/` so it might be something specific to his setup. I do not mind merging this though.

PR #1884, opened 2025-04-18
Ensure dev env runs for different k8s

### LINKED ISSUE #1882: [Bug] Ensure dev env runs for microk8s and docker desktop k8s
Current devspace and helm settings assume that containers can talk to host machine through `host.docker.internal` (the `host_uri`), which works well for docker k8s. But it does not work for microk8s even changing `host_uri` to actual ip address of the host machine. The reason is explained in a following example: 

v6 server container port 7601 is forwarded to host machine on port 7601, by default it's bound to ip `localhost`.
By setting `host_uri` to actual host ip, e.g. `192.168.74.1`, pod container cannot connect to  `192.168.74.1:7601` because the ip address is not bound to the port `7601`. So we also need to bind the port to this address in devspace settings.

Test updated dev env for following combinations:

mac

- [x] microk8s
- [x] docker k8s

ubuntu
- [x] microk8s

CunliangGeng:
Fix #1882.

### DIFF
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -205,6 +205,7 @@ dev:
     imageSelector: ${SERVER_IMAGE}
     ports:
       - port: 7601:7601
+        bindAddress: localhost
       - port: 7601:7601
         bindAddress: ${PORTS_BIND_ADDRESS}
 
@@ -214,6 +215,7 @@ dev:
     imageSelector: ${STORE_IMAGE}
     ports:
       - port: 7602:7602
+        bindAddress: localhost
       - port: 7602:7602
         bindAddress: ${PORTS_BIND_ADDRESS}
 
@@ -236,6 +238,7 @@ dev:
     command: ["/bin/sh", "/vantage6/vantage6-server/dev_server.sh"]
     ports:
     - port: 7601:7601
+      bindAddress: localhost
     - port: 7601:7601
       bindAddress: ${PORTS_BIND_ADDRESS}
     logs:
@@ -245,6 +248,7 @@ dev:
     imageSelector: ${STORE_IMAGE}
     ports:
       - port: 7602:7602
+        bindAddress: localhost
       - port: 7602:7602
         bindAddress: ${PORTS_BIND_ADDRESS}
 
@@ -272,7 +276,7 @@ dev:
       disableDownload: true
     ports:
       - port: 7600:4200
-        # bindAddress: ${PORTS_BIND_ADDRESS}
+        bindAddress: localhost
 
   # Where as we could hot-reload the server code using wsgi, we cannot do this with the
   # node code. We leverage the `restartHelper` and `onUpload.restartContainer` to
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -13,8 +13,13 @@ vars:
   TASK_DIRECTORY:
     question: "Where is your task directory located on the host machine?"
     default: ${PWD}/.tasks
+  HOST_URI:
+    question: "What is the ip address of your host machine?"
+    default: host.docker.internal
   #set to 0.0.0.0 to make the dev environment (ui, api) accessible from outside
-  # PORTS_BIND_ADDRESS: "127.0.0.1"
+  PORTS_BIND_ADDRESS:
+    question: "What is the bind ip address of your ports? [Bind to 'localhost' if using docker, otherwise the suggested address]"
+    default: ${HOST_URI}
   NUMBER_OF_NODES:
     question: "How many nodes do you want to create for your development environment?"
     default: 3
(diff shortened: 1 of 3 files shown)


devspace.yaml
frankcorneliusmartin, line 1: Now you define all ports twice, could you remove all duplicate ports? (Or did I miss something)

devspace.yaml
```diff
@@ -196,15 +205,17 @@ dev:
     imageSelector: ${SERVER_IMAGE}
     ports:
       - port: 7601:7601
-        # bindAddress: ${PORTS_BIND_ADDRESS}
+      - port: 7601:7601
```
frankcorneliusmartin, line 208:
```suggestion
```

devspace.yaml
```diff
@@ -196,15 +205,17 @@ dev:
...
+        bindAddress: ${PORTS_BIND_ADDRESS}
 
   # We use this to enable the server and the store to be coupled. We need a port to talk
   # to as the ingress is not deployed yet.
   vantage6-store-init:
     imageSelector: ${STORE_IMAGE}
     ports:
       - port: 7602:7602
-        # bindAddress: ${PORTS_BIND_ADDRESS}
+      - port: 7602:7602
```
frankcorneliusmartin, line 217:
```suggestion
```

devspace.yaml
```diff
@@ -225,15 +236,17 @@ dev:
     command: ["/bin/sh", "/vantage6/vantage6-server/dev_server.sh"]
     ports:
     - port: 7601:7601
-        # bindAddress: ${PORTS_BIND_ADDRESS}
+    - port: 7601:7601
```
frankcorneliusmartin, line 239:
```suggestion
```

devspace.yaml
```diff
@@ -225,15 +236,17 @@ dev:
...
+      bindAddress: ${PORTS_BIND_ADDRESS}
     logs:
       enabled: true
 
   vantage6-store:
     imageSelector: ${STORE_IMAGE}
     ports:
       - port: 7602:7602
-        # bindAddress: ${PORTS_BIND_ADDRESS}
+      - port: 7602:7602
```
frankcorneliusmartin, line 248:
```suggestion
```

frankcorneliusmartin (changes requested):

devspace.yaml
```diff
@@ -13,8 +13,13 @@ vars:
...
     question: "Where is your task directory located on the host machine?"
     default: ${PWD}/.tasks
+  HOST_URI:
+    question: "What is the ip address of your host machine?"
+    default: host.docker.internal
   #set to 0.0.0.0 to make the dev environment (ui, api) accessible from outside
-  # PORTS_BIND_ADDRESS: "127.0.0.1"
+  PORTS_BIND_ADDRESS:
+    question: "What is the bind ip address of your ports? [Bind to 'localhost' if using docker, otherwise the suggested address]"
+    default: ${HOST_URI}
```
bartvanb, line 22: I see we're adding more and more questions to the devspace setup. We should limit this as much as possible to avoid bothering the users.
Should we create an issue to see how to limit this as much as possible?

devspace.yaml
```diff
@@ -13,8 +13,13 @@ vars:
   TASK_DIRECTORY:
     question: "Where is your task directory located on the host machine?"
     default: ${PWD}/.tasks
+  HOST_URI:
+    question: "What is the ip address of your host machine?"
```
bartvanb, line 17: Most users will not know this. We should help them with suggestions what it should normally be in which OSs etc

CunliangGeng: There should be an reference to doc on helping figuring out the host ip.

devspace.yaml
CunliangGeng: I agree

devspace.yaml
CunliangGeng: It's something like below:

> If you are using docker k8s, then use `host.docker.internal` for host ip. Otherwise, you can find the host ip in following ways:
> 1. MacOS: login the Linux VM that is running the k8s, then use command `ip route | awk '/default/ {print $3}'` to get the ip
> 2. Linux: use command `ip route | awk '/default/ {print $9}'` or `hostname -I | awk '{print $1}'` to get the ip
> 3. Windows + WSL: ~~no idea yet, I guess same way as Linux~~

I think it's too long to put it in the question.

bartvanb: Agree! Then link to docs is better

CunliangGeng: Added this comment to the issue https://github.com/vantage6/vantage6/issues/1686#issuecomment-2827534938

devspace.yaml
CunliangGeng: I have made all bind addresses explicit for all port forwardings.

frankcorneliusmartin: Ah yes, now I understand. Thank you.

frankcorneliusmartin (approved):

PR #1887, opened 2025-04-22
1558 feature request community store without login

dsmits:
Of course it's not perfect but I think this contains the essential components.
New page can be found at `UI_URL/#/communitystore`.


dsmits:
Looks like there are some fake changed files because i just applied some formatting. I will clean it up

vantage6-ui/src/app/components/algorithm/algorithm-table/algorithm-table.component.html
```diff
@@ -0,0 +1,76 @@
+<table mat-table
```
bartvanb, line 1: Rather than visualizing the algorithms in a new way, I'd prefer using the existing component to be consistent to the user and to reduce the amount of code we need to maintain.
Maybe you can use this code snippet: `<app-display-algorithms [algorithms]="algorithms" [routeOnClick]="routePaths.algorithm"></app-display-algorithms>`. You would need this component to request the algorithm list but than the HTML should be minimal.
(Note that you will also need to request another route to view an algorithms details though)

bartvanb: I now looked a bit better at this full PR and what I think can be done to simplify is
- replace the `<app-algorithm-table>` in the community store component with a `<app-display-algorithms>`
- remove this component
- create a new route to visualize an algorithm's details (the one I'm mentioning above `routePaths.algorithm`, I think points to a route that's only accessible when authenticated).

Of course you are welcome to improve the `app-display-algorithms` component but I'd prefer not to have two similar components. Do you agree?

bartvanb (changes requested):
I think this is a supernice addition that will help potential users see what they can do with vantage6.

I have a few comments/questions:
- Could you add a link somewhere in the documentation that users can see this at  https://portal.cotopaxi.vantage6.ai/communitystore? That should help them find this page after this is released.
- The vantage6 logo is too big for the bar it's in, at least in my browser
- Note that the endpoint to get algorithms, by default, only returns approved algorithms. There are some flags, e.g. `under_review=1`, to get other statuses. But what would we prefer? Maybe only showing approved algorithms makes sense in this place

vantage6-ui/src/app/components/algorithm/algorithm-table/algorithm-table.component.html
dsmits: I do agree that reusing components is preferable. I guess I would also prefer the table in the private algorithm store setting, so maybe I could change the component to app-display-algorithms and create a new issue to improve it?

bartvanb: Hm maybe we should discuss this in a slightly bigger group, e.g. in standup tuesday? Then we avoid doing the work before we have consensus.
Personally I feel that I don't necessarily prefer the way it is now but there was an idea to add an image for algorithms at some point. What I want to avoid is having too much text so when doing a table we should limit that.

dsmits: Ok I realize I have another reason to create a new component rather than using the existing algorithm component. The route `routePaths.algorithm` points to the component `AlgorithmListReadOnlyComponent`. This component does many calls to the vantage6 server (e.g. in the [initData function](https://github.com/vantage6/vantage6/blob/a017666a4a244074f26aa53704f57a1d9cc1152c/vantage6-ui/src/app/pages/analyze/algorithm/read-only/algorithm-read-only.component.ts#L54)). My idea of the community store is that it should be independent of any vantage6 server. Moreover, the calls to the vantage6 server require authentication while I simply mean to access the publicly available information from the algorithm store.

I still agree that having only one component is still more desirable. However, the AlgorithmListReadOnlyComponent will require additional refactoring to make it work for the community store UI.

bartvanb: I think the `AlgorithmReadOnlyComponent` is mostly a component built around the `DisplayAlgorithmComponent`. I'm wondering if maybe we can use that component directly - otherwise we probably have to built a component _similar_ to the algorithmReadOnlyComponent. These components built around the DisplayAlgorithmComponent are pretty small though so that's not necessarily a big problem to me.

dsmits:
Ok so I had a talk with @bartvanb and we agreed on the following:
I will reuse the existing component, but I'm going to refactor `algorithm-read-only.component` so that v6 server related stuff such as collaboration and algorithm store id's are optional.

vantage6-ui/src/app/layouts/community-store/community-store.component.html
```diff
@@ -1,14 +1,9 @@
...
-    <mat-toolbar-row>
-        <a routerLink="{{routePaths.communityStore}}">
-            <img src="/assets/images/logo.png" class="logo" alt="vantage6"/>
-        </a>
-    </mat-toolbar-row>
+<mat-toolbar color="primary">
+    <a routerLink="{{routePaths.communityStore}}">
+        <img src="/assets/images/logo.png" class="logo" alt="vantage6"/>
+    </a>
+    <span class="store-header">Community Store</span>
```
bartvanb: Can you put this text in `en.json` and then use the translation as elsewhere?

docs/introduction/communitystore.rst
```diff
@@ -0,0 +1,5 @@
+Available algorithms
+=====================
+The vantage6 community has developed a number of algorithms for a variety of usecases. These can
```
bartvanb, line 3: usecases -> use cases

vantage6-ui/src/app/app-routing.module.ts
```diff
@@ -1,69 +1,141 @@
...
-import { StoreRoleListComponent } from './pages/store/role/list/store-role-list.component';
-import { StoreRoleReadComponent } from './pages/store/role/read/store-role-read.component';
-import { AlgorithmInReviewListComponent } from './pages/store/algorithms-in-review/algorithm-in-review-list/algorithm-in-review-list.component';
-import { AlgorithmAssignReviewComponent } from './pages/store/algorithms-in-review/algorithm-assign-review/algorithm-assign-review.component';
-import { ReviewReadComponent } from './pages/store/algorithms-in-review/review-read/review-read.component';
-import { ReviewSubmitComponent } from './pages/store/algorithms-in-review/review-submit/review-submit.component';
-import { MyPendingAlgorithmsComponent } from './pages/store/algorithms-in-review/my-pending-algorithms/my-pending-algorithms.component';
-import { OldAlgorithmListComponent } from './pages/store/algorithm/old-list/old-algorithm-list.component';
-import { StoreRoleCreateComponent } from './pages/store/role/create/store-role-create.component';
+import {NgModule} from '@angular/core';
```
bartvanb, line 1: Not sure why this is (auto)formatted differently. Could you undo it?

bartvanb: If it's difficult for you, I can also open the file, save and commit to undo this.

vantage6-ui/src/app/layouts/community-store/community-store.component.html
```diff
@@ -0,0 +1,14 @@
+<mat-toolbar color="primary">
+    <a routerLink="{{routePaths.communityStore}}">
```
bartvanb, line 2: Preferred way of interpolating this is `[routerLink]="routePaths.communityStore"`. Could you update (also in line 6)?

vantage6-ui/src/app/pages/store/algorithm/list-public/algorithm-list-public.component.ts
```diff
@@ -0,0 +1,53 @@
...
+import {
+  DisplayAlgorithmsComponent
+} from '../../../../components/algorithm/display-algorithms/display-algorithms.component';
+import {MatCard, MatCardContent} from '@angular/material/card';
+import {MatProgressSpinner} from '@angular/material/progress-spinner';
+import {TranslateModule} from '@ngx-translate/core';
+import {SafeHtmlPipe} from "../../../../pipes/safe-html.pipe";
+
+@Component({
+  selector: 'app-algorithm-list',
```
bartvanb, line 17: could you change this to `app-algorithm-list-public` (to be consistent with other components)? Also update elsewhere if you're using it of course

vantage6-ui/src/app/pages/store/algorithm/list-public/algorithm-list-public.component.ts
```diff
@@ -0,0 +1,53 @@
...
+  algorithms: Algorithm[] = [];
+
+  constructor(
+    private algorithmService: AlgorithmService
+  ) {
+  }
+
+  async ngOnInit() {
+    this.algorithms = await this.algorithmService.getAlgorithmsForCommunityStore();
+    console.log("algorithms", this.algorithms);
```
bartvanb, line 46: remove console log

vantage6-ui/src/app/components/algorithm/algorithm-table/algorithm-table.component.html
bartvanb, line 1: If I'm not mistaken, this component can now be removed. Right?

vantage6-ui/src/app/pages/store/algorithm/list-public/algorithm-list-public.component.html
```diff
@@ -0,0 +1,20 @@
+<mat-card>
+    <mat-card-content>
+        <h1>{{ 'communitystore.welcome' | translate }}</h1>
+        <span [innerHTML]="'communitystore.introduction' | translate | safeHtml"> </span>
```
bartvanb, line 4: Just curious why you are using the safeHtml pipe here? I don't know a lot about the pipe, but wondering why the text inserted would potentially be unsafe html? If that's the case, maybe we should start using this more

vantage6-ui/src/app/pages/store/algorithm/list-public/algorithm-list-public.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    MatProgressSpinner,
+    TranslateModule,
+    SafeHtmlPipe
+  ]
+})
+export class AlgorithmListPublicComponent implements OnInit, OnDestroy {
+  @HostBinding('class') class = 'card-container';
+  isLoading = true;
+  routePaths = routePaths;
+  destroy$ = new Subject<void>();
```
bartvanb, line 35: I dont think we are using this `destroy` anywhere in this component (there are no subscriptions to anything) so I think better to remove it

vantage6-ui/src/app/pages/store/algorithm/read-public/algorithm-read-public.component.html
```diff
@@ -0,0 +1,7 @@
+
+<ng-container *ngIf="!isLoading; else loading">
+<app-display-algorithm [algorithm]="algorithm"></app-display-algorithm>
+</ng-container>
+<ng-template #loading>
+    <mat-spinner diameter="48"></mat-spinner>
+</ng-template>
```
bartvanb, line 7: nice clean component :-)

vantage6-ui/src/app/services/algorithm.service.ts
```diff
@@ -1,12 +1,18 @@
...
+  AlgorithmLazyProperties,
+  ArgumentType
+} from 'src/app/models/api/algorithm.model';
+import {ChosenCollaborationService} from './chosen-collaboration.service';
+import {AlgorithmStore} from 'src/app/models/api/algorithmStore.model';
+import {Pagination} from 'src/app/models/api/pagination.model';
+import {ChosenStoreService} from './chosen-store.service';
+import {isListTypeArgument} from '../helpers/algorithm.helper';
+import {getLazyProperties} from '../helpers/api.helper';
+import {environment} from "../../environments/environment";
```
bartvanb, line 15: preferably reformat this as well

bartvanb (commented):
I'm happy with this PR, it's structured now exactly as I imagined it.
My comments are very minor, I don't think it should be too much work to address them and then we can merge it :-)

vantage6-ui/src/app/pages/store/algorithm/list-public/algorithm-list-public.component.html
dsmits: Ok so it's the only solution I found to be able to add a link in the localization file. But I don't feel very good about it because it introduces a security leak. If you add other html in that translation it would also be converted. I'm going to do it more pragmatically

dsmits:
Ok I think i fixed all your comments. Fixing the formatting took me the longest :sweat_smile:.

vantage6-ui/src/app/pages/store/algorithm/list-public/algorithm-list-public.component.ts
bartvanb: You removed the OnDestroy but the variable is still here :-)

bartvanb (approved):
Just one final comment. If you can also fix the Codacy issue (unused import), feel free to merge afterward!

Issue #1892, opened 2025-04-24
[Feature Request] [k8s] Store database connection details in environment variables


frankcorneliusmartin:
See parent issue for motivation.

Use a config map to define environment variables, for example database `"A"`:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  ...
data:
  DATABASE_LABELS: "A, B, C"
  DATABASE_A_URI: ""
  DATABASE_A_TYPE: ""
  DATABASE_A_[*]: "any_other_params"
  ...
```

Then we need to store passwords and such in Kubernetes secrets.

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
data:
  DATABASE_A_USERNAME: bXktYXBw
  DATABASE_A_PASSWORD: Mzk1MjgkdmRnN0pi
```

- [x] Update the `@source_database.py` to be able to consume both `DATABASE_URI`, `DATABASE_USERNAME` and `DATABASE_PASSWORD` and orter `DB_PARAM_*`s (if present).
- [x] Update the `run()` in `container_manager.py` to create secrets for the *username* and *password* for the database.
- [x] Update the `self._set_database(self.ctx.config["databases"])` in `container_manager.py` to consume environment variables for the databases instead of the config files.
- [x] Update the charts 
	- [x] Remove `databases` section from the configuration file Leave the `fileBased` section (as we need to create file mounts for these)
	- [ ] ~~Add `configMap` for the db connection details~~
	- [x] Add `secrets` for the db connection details 

Note that the `secret` and `configMap` is optional as some sources don't require credentials (typically only file-based sources, but also DB connections can disable authentication). This allows other (parent charts) to specify these themselves.

frankcorneliusmartin:
https://github.com/vantage6/vantage6/pull/1899

Issue #1743, opened 2025-02-18
[Feature Request] [k8s] Use secrets for passwords in Helm template files


frankcorneliusmartin:
**Problem description**
Plain passwords are bad practice, see: https://stackoverflow.com/questions/61888274/how-to-hide-passwords-in-the-helm-values-yaml-file

frankcorneliusmartin:
@bartvanb what do you think about removing the need for a `config.yml` altogether?

bartvanb:
> @bartvanb what do you think about removing the need for a config.yml altogether?

That depends on what you propose as alternative :-) The important thing is that users can set all settings easily and securely, which alternative for that do you propose?

frankcorneliusmartin:
@bartvanb: I dont see any added value in creating both a `values.yaml` (k8s) and a `config.yaml` which basically copies a lot of stuff from the `config.yaml`. So I propose the we attach the CLI to create the values.yaml and that k8s delivers these settings in environment variables (as this issue proposes). 

I already implemented this for the DB settings if you want to see what I mean.

bartvanb:
Before closing this: we should also move the rest of the configuration to env vars

PR #1899, opened 2025-04-24
Feature/move db configuration to environment variables

frankcorneliusmartin:
Fixed: https://github.com/vantage6/vantage6/issues/1892

Related PRs: https://github.com/vantage6/vantage6/pull/1902, https://github.com/vantage6/vantage6/pull/1905 merge this one first


vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -1200,6 +1271,30 @@ def __delete_job_related_pods(self, run_io: RunIO, namespace: str):
         for job_pod in job_pods_list.items:
             self.__delete_pod(job_pod.metadata.name, namespace)
 
+        self.__delete_secret(run_io.container_name, namespace)
+
+    def __delete_secret(self, secret_name: str, namespace: str) -> None:
+        """
+        Deletes a secret in a given namespace
```
CunliangGeng, line 1278:
```suggestion
        Deletes a secret in a given namespace

```

charts/node/values.yaml
```diff
@@ -20,18 +20,46 @@ node:
...
+
     serviceBased:
+    # Name is the LABEL in the DATABASE_LABELS variable
     - name: postgres
       uri: postgres://vantage6:vantage6@vantage6-vantage6-postgres:5432/vantage6
       type: other
 
+      # The env is the DATABASE_[LABEL]_SOME_OTHER_KEY variable. So in this example
+      # case, the node will set the DATABASE_POSTGRES_USER and
+      # DATABASE_POSTGRES_PASSWORD environment variables.
```
bartvanb, line 58: can we choose a different database label, e.g. `myPostgres` instead of `postgres`? That way the envvars become e.g. `DATABASE_MYPOSTGRES_PASSWORD` which is easier to understand without having to read carefully

charts/node/values.yaml
```diff
@@ -20,18 +20,46 @@ node:
...
+  #
+  # Optionally, you can also specify additional environment variables for each
+  # database by setting the DATABASE_[LABEL]_SOME_OTHER_KEY variable.
+  #
+  # DATABASE_[LABEL]_SOME_OTHER_KEY: some_other_value
+  #
+  # It is recommended to do this through Kubernetes secrets. If you specify the
+  # details here, secrets will be used. In case you are dealing with a file-based
+  # database, the node will automatically mount the file to the /app/databases/
+  # directory.
```
bartvanb, line 42: I would delete this last sentence. It is only understandable by vantage6 experts and I'm not sure why it's necessary to say here.

bartvanb: Maybe an alternative would be 'File-based databases are automatically made available to your node'

dev/load-basic-fixtures.py
```diff
@@ -108,7 +108,6 @@ def create_node(index, collaboration, organization, task_namespace):
         node_config = template.render(
             {
                 "api_key": node["api_key"],  # Use the API key from node creation
-                "databases": {"default": node_test_database_file},
```
CunliangGeng, line 111: The line 42 `node_test_database_file = args.node_test_database_file` can be removed, and also update the `args` and the use of this script in `devspace.yaml`

charts/node/values.yaml
```diff
@@ -20,18 +20,46 @@ node:
...
     - name: postgres
       uri: postgres://vantage6:vantage6@vantage6-vantage6-postgres:5432/vantage6
       type: other
 
+      # The env is the DATABASE_[LABEL]_SOME_OTHER_KEY variable. So in this example
+      # case, the node will set the DATABASE_POSTGRES_USER and
+      # DATABASE_POSTGRES_PASSWORD environment variables.
+      env:
+        USER: vantage6
+        PASSWORD: vantage6
```
bartvanb, line 61: We are adding examples now in the `values.yaml`. I'm wondering if that is the right place to add examples? By adding them here, it means that people running the charts always have to overwrite them, right? I would see this more as part of documentation. 
(If you agree, we can make this part of a documentation issue to review what needs to be part of values.yaml's)

CunliangGeng (changes requested):

devspace.yaml
```diff
@@ -170,21 +170,17 @@ deployments:
...
       chart:
         name: ./charts/node
+      # See the env section in the dev section where the database environment variables
+      # are set.
       values:
         node:
-          taskDirHost: ${TASK_DIRECTORY}
-          image: ${NODE_IMAGE}
-          logging:
-            level: DEBUG
```
bartvanb, line 178: I would prefer to keep DEBUG logs for the dev environment

vantage6-algorithm-tools/vantage6/algorithm/decorator/source_database.py
```diff
@@ -22,20 +26,73 @@ def decorator(*args, mock_uri: str | None = None, **kwargs) -> callable:
...
+        For example, for a PostgreSQL database. The connection details are retrieved by
+        the node from the environment variables and passed on to the algorithm
+        container. This always follows the `DATABASE_*` naming convention.
+
+        For example when the nodes reads the following environment for database A, and
+        the user requests database `A`:
+        ```bash
+        DATABASE_A_URI="postgresql://user:password@host:port/database"
+        DATABASE_A_USERNAME="postgresql"
+        DATABASE_A_PASSWORD="password"
```
bartvanb, line 45: I'm not sure now if I understand your example connection strings. I the database URI above, the username is `user` and the password `password`, right? Then why is the username also `postgresql`? 
I'm starting to doubt my knowledge of SQL connection strings, there is no additional authentication layer, right? If there isn't, then why would we even pass the username/password separately if we already have the URL?

vantage6-algorithm-tools/vantage6/algorithm/decorator/source_database.py
```diff
@@ -22,20 +26,73 @@ def decorator(*args, mock_uri: str | None = None, **kwargs) -> callable:
...
+        DATABASE_A_OTHER_DETAILS="..."
+        ```
+
+        Then the algorithm container will pass the following environment variables to
+        the algorithm (note that the `A_` annotation is removed):
```
bartvanb, line 50: How does this work with multiple databases?

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -201,31 +201,66 @@ def _setup_policies(self, config: dict) -> dict:
...
+        This function will return a dictionary with contains a key for every data
+        source. Each key will have a dictionary with the following keys:
+        - uri: the URI of the database
+        - local_uri: the local URI of the database
+        - is_file: whether the database is a file
+        - is_dir: whether the database is a directory
+        - type: the type of the database
+        - env: the environment variables to be passed to the container
+
+        It reads these from the environment variables that are passed to the container.
```
bartvanb, line 217: Nice that you extended this docstring!
```suggestion
        It reads these from the environment variables that are passed to the node.
```

charts/node/values.yaml
```diff
@@ -20,18 +20,46 @@ node:
...
     serviceBased:
+    # Name is the LABEL in the DATABASE_LABELS variable
     - name: postgres
       uri: postgres://vantage6:vantage6@vantage6-vantage6-postgres:5432/vantage6
       type: other
 
+      # The env is the DATABASE_[LABEL]_SOME_OTHER_KEY variable. So in this example
+      # case, the node will set the DATABASE_POSTGRES_USER and
+      # DATABASE_POSTGRES_PASSWORD environment variables.
+      env:
```
bartvanb, line 59: From later in this PR *I think* that these environment variables are just examples and have nothing to do with the connection URI. If that's correct, could you modify these environment examples to something like `SOME_ENV_VAR = example`? 
The current env var examples are super confusing for me as I kept thinking they were somehow used to replace parts of the URI connection string.

bartvanb: Note that this comment not only applies here but also in other places in this PR

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -699,11 +766,15 @@ def _create_volume_mounts(
             environment_variables[ContainerEnvNames.DATABASE_TYPE.value] = db["type"]
 
             # additional environment variables for the database. These will be stored
-            # as {PREFIX}{KEY}=value in the container
+            # as {PREFIX}{KEY}=value in the container. These are red by the
```
bartvanb, line 769: red -> read

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -699,11 +766,15 @@ def _create_volume_mounts(
...
+            # `@source_database` decorator in the (data extraction) algorithm.
             if "env" in db:
                 for key in db["env"]:
                     env_key = f"{ContainerEnvNames.DB_PARAM_PREFIX}{key.upper()}"
-                    environment_variables[env_key] = db["env"][key]
+                    if key.lower() in ["password", "pass", "user", "username"]:
```
bartvanb, line 774: this seems a bit flaky. E.g. is doesn't work if someone has `postgres_pass` or so. Maybe we can simply pass all environment variables as secrets?

bartvanb (changes requested):
Not sure that I understood everything correctly, so also some questions remaining. 

I think it would be good to start on the documentation of stuff sooner rather than later, because I'm starting to notice that with several PRs it would be easier to understand the PR once the docs are also updated for that PR - let's discuss that also soon.

charts/node/templates/node-secrets.yml
```diff
@@ -0,0 +1,15 @@
...
+  name: {{ .Release.Name }}-node-secrets
+  namespace: {{ .Values.namespace | default .Release.Namespace }}
+data:
+  {{ range $db := concat .Values.node.databases.fileBased .Values.node.databases.serviceBased }}
+  DATABASE_{{ $db.name | upper }}_URI: {{ $db.uri | b64enc }}
+  DATABASE_{{ $db.name | upper }}_TYPE: {{ $db.type | b64enc }}
+  {{- if $db.env }}
+  {{- range $key, $value := $db.env }}
+  DATABASE_{{ $db.name | upper }}_{{ $key | upper }}: {{ $value | b64enc }}
+  {{- end }}
```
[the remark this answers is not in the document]
CunliangGeng: I think the whole idea or motivation is to replace `config.yaml` with env vars, as described in #1743, which is the recommended practices for k8s applications.

devspace.yaml
frankcorneliusmartin: Yes I agree, should not have removed that: 7f95c5d11ac7dba6e2ee9abd0a4c9cac3023aa8c

charts/node/values.yaml
frankcorneliusmartin: Ok I feel this is a more general issue and not specific for this issue. I created a separate issue for this: https://github.com/vantage6/vantage6/issues/1924

charts/node/values.yaml
frankcorneliusmartin: 5013d95088f09c49b6e7577240cf366f57ca7dd0

charts/node/values.yaml
frankcorneliusmartin: a24ba9bd48cdac9143695343e83a8b1a6483d07a

vantage6-algorithm-tools/vantage6/algorithm/decorator/source_database.py
```diff
@@ -3,13 +3,17 @@
...
 from vantage6.common import error
 from vantage6.common.globals import ContainerEnvNames
+from vantage6.algorithm.tools.util import get_env_var
 
 
 def source_database(func) -> callable:
     @wraps(func)
-    def decorator(*args, mock_uri: str | None = None, **kwargs) -> callable:
+    def decorator(
+        *args, mock_uri: str | None = None, mock_type: str | None = None, **kwargs
```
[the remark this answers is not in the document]
frankcorneliusmartin: Added param doc: 9ccec0ae3ab75fa5d1d62d40a1f94725fd9513b7

Note that I did not list all types here as the function itself has to deal with the argument type. So any type is allowed but depending on your `@data_extraction` function it is used or not.

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -238,21 +273,27 @@ def _set_database(self, config_databases: dict | list) -> dict:
...
             db_is_file = tmp_uri.exists() and tmp_uri.is_file()
             db_is_dir = tmp_uri.exists() and tmp_uri.is_dir()
 
             if db_is_file or db_is_dir:
                 local_uri = str(tmp_uri)
 
+            # Get additional environment variables
+            env = {}
+            for key in os.environ:
+                if key.startswith(f"DATABASE_{label.upper()}_"):
```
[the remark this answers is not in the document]
frankcorneliusmartin: c416bfb97b4684ac36a3ab32a074d053b9a776cf

dev/load-basic-fixtures.py
frankcorneliusmartin: 5b533711943b6d85965e27590a987092d7c99cc4, 692975f83e54222791d48d9c32ab83eddacf2d20 and fb355dd4b74c0885b687f2bee6619e51c8fb2f21

Sorry some of these had already been fixed, so that is why this is distributed over 3 commits.

charts/node/templates/node-secrets.yml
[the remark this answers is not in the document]
frankcorneliusmartin: @bartvanb 
> We could argue that file paths may also reveal private information but we could also choose to put the file-based dbs in the regular config.

I agree, probably this is not the case for file based dbs. I guess I did it for simplicity as there is also no real harm done and it gives very little debug info? Let me know if you think it is important to separate them.

@CunliangGeng 
Yes, you are right. However Bart suggestion is to use non-secrets for non-sensitive data. It is possible to use a `ConfigMap` instead of a `Secret` in k8s to create the environment variables.

vantage6-algorithm-tools/vantage6/algorithm/decorator/source_database.py
frankcorneliusmartin: I agree, this is an confusing example. I guess I forgot to remove the user/pass from the URI when I defined them separately.

fb355dd4b74c0885b687f2bee6619e51c8fb2f21

vantage6-algorithm-tools/vantage6/algorithm/decorator/source_database.py
frankcorneliusmartin: `@data_extraction` jobs only work for single data sources. `@federated` and `@central` **can** use multiple dataframes.

vantage6-node/vantage6/node/k8s/container_manager.py
frankcorneliusmartin: 955ff2bf36eaba974a573ddb95b9a27548227ff3

charts/node/values.yaml
frankcorneliusmartin: I made it consistent with the other suggestion you made.

28b4940e87e75734087d35376c24d124c59fd0df

vantage6-node/vantage6/node/k8s/container_manager.py
frankcorneliusmartin: 4cceb35b013a9643c990712cf94234f052b2175d

vantage6-node/vantage6/node/k8s/container_manager.py
frankcorneliusmartin: All env entries are now considered secret: 

df47a5916e42b94b418299383c5864f86813d2ba

dev/load-basic-fixtures.py
CunliangGeng: Line 29-31 should also be removed, they are not needed any more

CunliangGeng (approved):
:shipit: It's a great feature to use env vars for config. Just one more minor comment to solve, then you can merge it.

dev/load-basic-fixtures.py
frankcorneliusmartin: Thanks, done: 15b65e0e24178466601f26ab15b0fe5ede79c9cc

charts/node/templates/node-secrets.yml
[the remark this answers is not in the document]
bartvanb: OK let's leave it like this. I think what @CunliangGeng meant is that we aim to replace config.yaml *completely* by env vars, so then replacing the nonsensitive stuff here also makes sense

vantage6-algorithm-tools/vantage6/algorithm/decorator/source_database.py
bartvanb: Ah yes of course. My mind was with v4

charts/node/values.yaml
```diff
@@ -20,18 +20,45 @@ node:
...
       uri: /mnt/data/default.csv
       type: csv
+
+
     serviceBased:
-    - name: postgres
-      uri: postgres://vantage6:vantage6@vantage6-vantage6-postgres:5432/vantage6
+    # Name is the LABEL in the DATABASE_LABELS variable
+    - name: myPostgres
+      uri: postgres://vantage6-vantage6-postgres:5432/vantage6
```
bartvanb, line 52:
```suggestion
      uri: postgres://vantage6-postgres:5432/vantage6
```

bartvanb (approved):
Nice! Just one tiny suggestion, feel free to merge. Let me know when you do, then I'll start on the next PR in the chain

PR #1912, opened 2025-04-29
Option to not share algorithm logs with server

lsago:


docs/node/node_config.yaml
```diff
@@ -289,3 +289,8 @@ task_dir: C:\Users\<your-user>\AppData\Local\vantage6\node\mydir
 # for other organizations in your collaboration to understand why a task
 # is not completed. Obviously, no sensitive data is shared. Default true
 share_config: true
+
+# Whether or not to share algorithm logs with the server. Otherwise they will
+# only be displayed as part of the node logs. Default is true.
+# NOTE: It's recommented to set this to false when using real data
+share_logs: false
```
frankcorneliusmartin, line 296: As we might also want to include node logs at some points I would like to change this to `share_algorithm_logs` or something similar.

vantage6-node/vantage6/node/docker/docker_manager.py
```diff
@@ -175,6 +175,19 @@ def __init__(
...
+        # set defaults, warn about dangers, etc
+        self.share_logs = config.get("share_logs", True)
+        if self.share_logs:
+            self.log.warning(
+                "Algorithm logs are shared with the server. "
+                "This may contain sensitive information."
+            )
+            self.log.warning(
+                "Setting 'share_logs' to False is advisable in a production scenario."
+            )
```
frankcorneliusmartin, line 189: At the time the risk depends on the algorithms that are allowed on this node. 

I do agree with you it would be better to have a fixed pre defined number of errors, but since that feature is not completely implemented I would not want people to disable it in our production cases as debugging would become super difficult.

frankcorneliusmartin: Maybe this can be a little bit mitigated by encrypting the logs?

frankcorneliusmartin: @bartvanb how do you feel about this?

frankcorneliusmartin (changes requested):

vantage6-node/vantage6/node/docker/docker_manager.py
bartvanb: I also think that debugging without sharing logs might become super difficult, especially for when dealing with non-responsive or non-knowledgeable node admins. Encrypting the logs is not really a solution I think, because you would still share sensitive logging, only with fewer people.
The best solution would indeed be to have a more intricate error handling system. I think almost all leaks will come from tracebacks after unexpected errors - if we catch those and don't share sensitive stuff that would be better. Not sure we can make that 100% failsafe without looking into it though.

So it depends: do we want to put this responsibility with the algorithm developer or not? It's a balance between security and usability. I'm inclined to go for usability now, and A) remove these logs and B) instruct algo devs to pay attention to this in the algo dev docs.

bartvanb: It's a tricky balance indeed. I am satisfied with the compromise you suggest. We are working towards a better solution but as long as we don't have that, I agree that we should warn for it. I think vantage6 has reached a stage where it's used in mature use cases in which people may expect warnings for such things.
In saying that, actually I find that it's difficult for me to estimate the security/usability balance here as I'm not often directly involved in runs with real data, so I also think my opinion here is less important than yours and Frank's.

frankcorneliusmartin: I also agree with the compromise. In addition, we should consider adding a try and catch to the wrapper.. (that we can turn of for devbuilds of the algorithms)

vantage6-node/vantage6/node/docker/task_manager.py
```diff
@@ -176,7 +179,6 @@ def report_status(self) -> str:
         if self.status_code:
             self.log.error(f"Received non-zero exitcode: {self.status_code}")
             self.log.error(f"  Container id: {self.container.id}")
-            self.log.info(logs)
```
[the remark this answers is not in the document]
frankcorneliusmartin: Now they are always printed right? Regardless of their state? I am oke with removing this line. However maybe we should a char limit not to overload our logging system in case an algorithm goes completely wild with logging?

frankcorneliusmartin (approved):
There is one minor comment left, if you address it this PR is OK for me.

bartvanb (commented):
Sorry for late review. 
I just wanted to voice one concern: if the logs of the algorithm are not shared to the server, I feel that they should remain available on the node. IIRC the algorithm logs are printed to the node logs if there is an error, so in that case we should normally be able to find them back. However, the logs of successful algorithm runs would be lost forever. 
We could do a couple of things: 1) always print the algo logs (or just the last N lines in case there are many) in the node log, 2) create a log file that stores the logs of algorithm runs (similar to the separate proxy server log).
Option 2 sounds like a good solution to me but might be a bit much for this PR - we could create a separate issue if you agree. Option 1 could be a quick fix that can be included in this PR.
What do you guys think?

vantage6-node/vantage6/node/docker/task_manager.py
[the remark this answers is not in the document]
frankcorneliusmartin: So lets make a new issue for addressing both the char limit and storing the algorithm log files in a separate file. Because if we limit it here without an external file we lose information. Would that be oke?

frankcorneliusmartin:
I am OK with merging it is as it is now and create a separate issue to address:

- Limit the chars that are displayed in the node logs 
- Store the algorithm logs (per run or task) in a separate file

What do you think about this @bartvanb @lsago

bartvanb:
Agreed!

frankcorneliusmartin:
@lsago can you create the issues?

Issue #1913, opened 2025-04-30
[Bug] (k8s + sessions) The Job PODs can't reach their corresponding Node proxy in the dev-environment


hcadavid:
In the dev environment, multiple Nodes are deployed in the same POD, hence this POD provide access to each one's proxy using a different IP Port. For the Job PODs of each node to reach their corresponding proxy, the Node-Service POD must enable the same ports.

However, the Service is configured to expose only the default port. More importantly, the proxy ports for a second or a third Node are defined randomly: when creating the node, if the default port is used by another process (in this case, the first node), it choses a random one:

https://github.com/vantage6/vantage6/blob/2927a88327c862e3bf67a52edbedcaa381b2c5c4/vantage6-node/vantage6/node/__init__.py#L199

Given this, it is not posible to define which ports need to be forwarded on the Node-Service POD when creating it. 

This is an issue for the dev environment, where multiple nodes are created in the same POD, but will also be a problem on a real setup when the allocation of the default port fails, as the Node-Service POD is statically configured to forward the default port.

PR #1914, opened 2025-04-30
Enabling communication between the Jobs and their corresponding Node proxy

hcadavid:
These changes address issue #1913, by replacing the use of randomly generated port numbers with a fixed set of ports:

_Like in the original V6-node, in the new one you can configure the port that will be used for the proxy (PROXY_SERVER_PORT). The problem is, when the allocation of this default port fails (e.g., because it is being used by other process) a random one is chosen._ 

_The dev-environment is using this feature to spin multiple nodes within the same POD, without a ports conflict (so it always has one v6-node with the default port, and N-1 v6-nodes with random ones). However, with exception of the v6-node no.1, these nodes are not reachable by their corresponding jobs anyway, as the Node-service doesn't have such random ports enabled for forwarding. This is a -provisional- workaround to fix this by switching from random ports to pre-defined ones._

This solution implies:

- There should be a limit on the number of nodes that can be setup on the dev. environment .
- This limit correspond to the number of ports defined on `charts/node/values.yaml`, which in turn correspond to the list of fixed ports on globals.DEV_NODES_PREDEFINED_PORTS
- In general, for consistency, there should be a strong constraint on the availability of these ports. That is, the alternative ones would be used only if a previous one was used by another Node. Not having the default port during the startup should make the node to fail -so it is responsibility of the person doing the setup to either release the port or to change the value of the default one.

Another alternative may be creating a POD for each Node on the dev environment, so that the default port can always be used regardless of the number of Nodes, but I'm not sure how complex this change would be.

I used [this algorithm](https://github.com/hcadavid/v6-sessions-k8s-diagnostics/blob/9dbb7e4458c817a2553e184206eb7d0bd8d2304c/v6-session-basics/partial.py#L192) to test that each job can reach the Node-proxy API with the FQDN and proxy port given to them .

@frankcorneliusmartin @bartvanb could you give me your thoughts on this?

### DIFF
--- a/charts/node/templates/node-deployment.yml
+++ b/charts/node/templates/node-deployment.yml
@@ -27,7 +27,7 @@ spec:
         image: {{ .Values.node.image }}
         env:
         - name: PROXY_SERVER_PORT
-          value: "{{ .Values.node.proxyPort }}"
+          value: "{{ (index .Values.node.proxyPorts 0).port}}"
         command: ["vnode-local", "start", "-c", "/app/config/config.yml", "--dockerized"]
         volumeMounts:
         - name: task-files-root
--- a/charts/node/templates/node-service.yml
+++ b/charts/node/templates/node-service.yml
@@ -8,16 +8,9 @@ spec:
 {{ include "node.matchLabels" . | indent 4 }}
     component: node
   ports:
-    # These ports must match ones defined on DEV_NODES_PREDEFINED_PORTS
-    - name: node1
+{{- range .Values.node.proxyPorts }}
+    - name: {{ .name }}
       protocol: TCP
-      port:  7654
-      targetPort:  7654
-    - name: node2
-      protocol: TCP
-      port:  6171
-      targetPort:  6171
-    - name: node3
-      protocol: TCP
-      port:  3986
-      targetPort:  3986
+      port: {{ .port }}
+      targetPort: {{ .targetPort | default .port }}
+{{- end }}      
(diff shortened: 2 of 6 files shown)


bartvanb:
If we can prevent limiting the number of nodes in a dev env that would have my preference. And also, I'm wondering if we don't need to know the port once we go for a real deployment?

Maybe we can make the proxy port a configuration option? Then in de dev env we could set it up by assigning a random port to the node configuration file and the node would just get it from the config.

(Note that if we go that route we should take care that @frankcorneliusmartin also takes this into account when replacing the config with environment variables)

hcadavid:
> Maybe we can make the proxy port a configuration option? Then in de dev env we could set it up by assigning a random port to the node configuration file and the node would just get it from the config.

Sounds like a a good option! This would also require passing the list of ports to `load-basic-fixtures.py` as an argument.

Considering this, and for consistency, it would also be important to remove the port allocation-retries from the node:

https://github.com/vantage6/vantage6/blob/2927a88327c862e3bf67a52edbedcaa381b2c5c4/vantage6-node/vantage6/node/__init__.py#L195-L201

Regardless of the dev-environment problem, I think a sys-admins would prefer the proxy-port setting to be consistent in a production environment (and not having this randomly assigned ports). If the v6-node fails because of an already allocated port, this is something worth to make the node process to fail (instead of continue with the random port) and to pay attention on what's going on (e.g., another node may be already running).

@frankcorneliusmartin what do you think?

hcadavid:
I'm closing this PR, a different approach was succesfully implemented and tested on https://github.com/vantage6/vantage6/pull/1945

Issue #1782, opened 2025-03-05
[Change Request] Devspace is linked to pods, not services


bartvanb:
**Description**
When implementing #1779, I found out that the ports that we are attaching the devspace to, are ports of the pods (double check that it's not the container!). It would be nicer if the services were used directly, because this way we are not testing that the services are configured properly in the dev environment. But we need to find out whether this is also possible with devspaces

CunliangGeng:
It's not possible to fort forward a k8s service in devspace. Alternatively, we could manually use `kubectl port-forward svc/myservice ...` to forward a service.

But you may not want to forward all services (except the UI one), because most services (`ClusterIP` type) are only used by pods within the k8s cluster. 

As for the UI service, I think we should change its type from `ClusterIP` to `NodePort` or `LoadBalancer`, this will make the service directly reachable from outside of the k8s cluster (so we don't need port forward any more in devspace for UI).

bartvanb:
Cool that you found a way to forward the service! I feel like I don't know enough at present about the k8s services so can't comment on whether using `NodePort` or `LoadBalancer` is appropriate - maybe @hcadavid / @frankcorneliusmartin can comment?

I do think that in a development environment it's nice not to only be able to access the UI via localhost. E.g. it's nice to be able to use postman or something to access endpoints of the server / algorithm store separately. So I would suggest doing the same for server and store as well (and it would also be nice to be able to access the database, there are also issues for that somewhere)

bartvanb:
Let me clarify myself: What I understood from Cunliang's post was that he intended the UI to be available on localhost:7600 but that the server and algorithm store would no longer be available on localhost:7601 and localhost:7602, as is the case at present. The assumption was that the server and store are only used internally. 
My point is that I *do* want server+store to remain available on localhost, because one might not just use UI but could also use a Python client to test stuff (or even use the API directly via e.g. Postman), and then they are required to be exposed. I get that you might then use the Cluster IP address but that is inconvenient because it's not fixed so you need to look it up and change it every time.

CunliangGeng:
>Let me clarify myself: What I understood from Cunliang's post was that he intended the UI to be available on localhost:7600 but that the server and algorithm store would no longer be available on localhost:7601 and localhost:7602, as is the case at present. The assumption was that the server and store are only used internally.

Some misunderstandings here. The server and algorithm store are still available on localhost:7601 and localhost:7602, but they are forwarded directly from pods instead of from services:
[...]

PR #1925, opened 2025-05-08
Change/1782 port forward UI k8s service

CunliangGeng:
Solve #1782.

Note that:
1. Docker k8s supports load balancer, so the external IP would be `localhost` for dev env
2. Microk8s does does not natively have load balancer, so to use LoadBalancer service for dev env, you can use https://metallb.io/ to enable load balancer, or use `kubectl port-forward svc/vantage6-server-vantage6-frontend-service  7600:7600` to port forwarding the service. (these should be put to doc)

### DIFF
--- a/charts/server/values.yaml
+++ b/charts/server/values.yaml
@@ -98,7 +98,5 @@ ui:
 
   # The image of the UI
   image: harbor2.vantage6.ai/infrastructure/ui:latest
-
-
-
-
+  # External port for the UI
+  port: 7600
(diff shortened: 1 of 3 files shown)


charts/server/values.yaml
```diff
@@ -98,7 +98,5 @@ ui:
 
   # The image of the UI
   image: harbor2.vantage6.ai/infrastructure/ui:latest
-
-
-
-
+  # External port for the UI
```
frankcorneliusmartin, line 101: I would call this `service` port, as it is not to be confused with the port that is used by the user. Typically a nginx ingress would define that port.

frankcorneliusmartin (approved):

PR #1926, opened 2025-05-08
Feature/only central compute action can get token from server

frankcorneliusmartin:
This PR is based on #1905 as I needed some stuff from that PR. Merge that one first :)


vantage6-server/tests_server/test_resources.py
```diff
@@ -2866,8 +2875,19 @@ def test_create_task_permissions_as_container(self):
...
+        # Test wrong collaboration_id
+        results = self.app.post(
+            "/api/task",
+            headers=headers,
+            json=task_json,
+        )
+        self.assertEqual(results.status_code, HTTPStatus.UNAUTHORIZED)
+
+        # test wrong action
+        task_json["action"] = AlgorithmStepType.FEDERATED_COMPUTE
```
bartvanb, line 2890: I think `task_json["collaboration_id"] = col.id` from line 2899 should be moved here: otherwise we are testing both with wrong action AND still with wrong collaboration ID, so we cannot really test whether we get 401 because of the one or the other

bartvanb (approved):
Feel free to merge after addressing my one comment

vantage6-server/tests_server/test_resources.py
frankcorneliusmartin: fc9982c35ed7bf069c6b96b9e999405adee2635f

PR #1927, opened 2025-05-08
1916 change request make algorithm cards more descriptive

dsmits:
Added more info on the cards when you are listing the algorithms. Let me know what you think.

Selection_408


vantage6-ui/src/app/components/algorithm/display-algorithms/display-algorithms.component.ts
```diff
@@ -25,4 +27,27 @@ export class DisplayAlgorithmsComponent {
...
+  }
+
+  getStatusClass(status: AlgorithmStatus): string {
+    switch (status) {
+      case AlgorithmStatus.Approved:
+        return 'status-badge-approved';
+      case AlgorithmStatus.Rejected:
+        return 'status-badge-rejected';
+      case AlgorithmStatus.Removed:
+        return 'status-badge-removed';
```
bartvanb, line 46: The status `AlgorithmStatus.Replaced` is missing here - it would default incorrectly to the pending status while the algorithm is replaced by a newer version. Could you make a separate case (+ css etc) for that one? Alternatively, you could update the 'removed' case as effectively that version of the algorithm is no longer in effect (but I'd then maybe rename the text to 'invalidated' or so)

bartvanb (commented):
One tiny comment, and one hesitation: I wonder if it makes sense to show the algorithm status in the card everywhere, because on the most-used pages, all algorithms will have status approved (the exceptions are the pages in the algorithm store management section). So I wonder if that adds anything on most pages. Maybe there could be a boolean input determining if the status is shown on that page?

Also, note to all of us: if we show the version on the algorithm card, we should take more care to update it ;-)

vantage6-ui/src/app/components/algorithm/display-algorithms/display-algorithms.component.ts
dsmits: You're right, making it the pending styling would be incorrect. Replaced will now get the same styling as removed. The status name will still be passed along, this function is just for the styling, not the text.

vantage6-ui/src/app/pages/store/algorithm/list-public/algorithm-list-public.component.html
```diff
@@ -9,7 +9,8 @@ <h1>{{ 'communitystore.welcome' | translate }}</h1>
 <app-page-header title="{{ 'communitystore.algorithms' | translate }}"></app-page-header>
 <ng-container *ngIf="!isLoading; else loading">
     <app-display-algorithms [algorithms]="algorithms"
-                            [routeOnClick]="routePaths.communityStoreAlgorithms"></app-display-algorithms>
+                            [routeOnClick]="routePaths.communityStoreAlgorithms"
+                            [showStatus]="true"></app-display-algorithms>
```
bartvanb, line 13: I'm wondering if this should be 'true' - the status will be approved for all algorithms. I would like to include `[showStatus=true]` mostly on all places in the algorithm store admin components where this component is used. Could you add that in this PR? Creating separate issue is also fine but it's just small so I think we could include it

bartvanb (approved):
One more comment - I'm also fine with moving that to separate issue but I think it's small enough.

PR #1928, opened 2025-05-08
Expose databases for db GUI

### LINKED ISSUE #1794: [Feature Request] Easily interact with the database in devspace
**Problem description**
During development it can be very helpful to interact with the database. For v4 we regularly used SQLite DB browser but now we are running a Postgres Pod

**Desired solution**
- Think of / provide nice way to interact with database, preferably via some browser software
- Document it

CunliangGeng:
Fix #1794.

### DIFF
--- a/charts/store/values.yaml
+++ b/charts/store/values.yaml
@@ -52,7 +52,7 @@ database:
   password: vantage6
 
   # The name of the database
-  name: store
+  name: vantage6
 
   #hostpath of the database mount
   volumePath: /mnt/data_store
--- a/devspace.yaml
+++ b/devspace.yaml
@@ -79,8 +79,10 @@ pipelines:
         vantage6-store
       start_dev \
         vantage6-server \
+        vantage6-server-db \
         vantage6-ui \
-        vantage6-store
+        vantage6-store \
+        vantage6-store-db
 
   node:
     run: |-
@@ -243,7 +245,18 @@ dev:
       bindAddress: ${PORTS_BIND_ADDRESS}
     logs:
       enabled: true
-
+  
+  vantage6-server-db:
+    labelSelector:
+      app: vantage6-server
+      component: postgres
+    ports:
+      # expose the server database port
+      - port: 5432:5432
+        bindAddress: localhost
+      - port: 5432:5432
+        bindAddress: ${PORTS_BIND_ADDRESS}
+  
   vantage6-store:
     imageSelector: ${STORE_IMAGE}
     ports:
@@ -264,6 +277,16 @@ dev:
       startContainer: true
     command: ["/bin/sh", "/vantage6/vantage6-algorithm-store/dev_store.sh"]
 
+  vantage6-store-db:
+    labelSelector:
+      app: store
+      component: postgres
+    ports:
+      # expose the store database port
+      - port: 5433:5432
+        bindAddress: localhost
+      - port: 5433:5432
+        bindAddress: ${PORTS_BIND_ADDRESS}
 
   # This puts the store container in development mode.
   vantage6-ui:


frankcorneliusmartin:
Hi @CunliangGeng thanks for addressing this. In the original issue Bart also requests a UI for the DB. Can we also add a pgadmin container for this?

charts/store/values.yaml
```diff
@@ -52,7 +52,7 @@ database:
   password: vantage6
 
   # The name of the database
-  name: store
+  name: vantage6
```
bartvanb, line 55: The effect that this has, is that if I'm logging in to a DB UI, I now see a database named 'vantage6', right? Then why not keep the name 'store' as that would be more obvious?

devspace.yaml
```diff
@@ -264,6 +277,16 @@ dev:
       startContainer: true
     command: ["/bin/sh", "/vantage6/vantage6-algorithm-store/dev_store.sh"]
 
+  vantage6-store-db:
+    labelSelector:
+      app: store
```
bartvanb, line 282: should this be `vantage6-store`?

charts/store/values.yaml
bartvanb: I just noticed that for the server this is also called `vantage6`. Maybe make that one `vantage6-server` and this `vantage6-store`?

bartvanb (commented):
Nice work! I like not directly running `pgadmin` (@frankcorneliusmartin Cunliang explains this in the linked issue) but giving devs the option to choose their own. Especially since we don't spin up the database UI that often, so it's nice to make the dev env a bit lighter to run by not automatically including it.

I have mostly questions about how exactly it works now. Apart from the questions below, I was wondering: with the current setup, are the server and store databases both exposed on localhost:5432? The usernames/passwords for both are `vantage6 / vantage6` as specified in the values.yamls. Isn't this a conflict?

Finally, when I wanted to test this, I got
```
start_dev: [forward ports: unable to listen on any of the requested ports: [{5432 5432} {5432 5432}], error selecting pod: timeout: couldn't find a pod / container in time with pod name: vantage6-store-postgres-deployment-6464bb8d7d-mlh48, waiting for pod to become ready: timeout: couldn't find a pod / container in time with image selector: harbor2.vantage6.ai/infrastructure/algorithm-store:BvdXfQV]
```
Did it work for you @CunliangGeng?

charts/store/values.yaml
CunliangGeng: Changing `store` to `vantage6` is meant to make different databases have the same name, so that users don't need to remember different db names when connecting databases in GUI (there they have to fill the database name). With this change, users can use only `vantage6` for name, username and password of all databases.

In GUI, users can give different names to each connection, so they can still differentiate each database.

devspace.yaml
CunliangGeng: No, it's `store`.

image

CunliangGeng:
@bartvanb 

The server's db is exposed on 5432, and store's db on 5433.
Here is how I connect to server db
image
and store db
image



> As for the error: start_dev: [forward ports: unable to listen on any of the requested ports: [{5432 5432} {5432 5432}], error selecting pod: timeout: couldn't find a pod / container in time with pod name: vantage6-store-postgres-deployment-6464bb8d7d-mlh48, waiting for pod to become ready: timeout: couldn't find a pod / container in time with **image selector**: harbor2.vantage6.ai/infrastructure/algorithm-store:BvdXfQV]

I don't have this error, it looks even the `vantage6-store-store-deployment` is not running since there is no pod with the image selector. We did not use image selector for database (only label selector).

Maybe try purge everything then rerun (populate-server and start-dev)?  or rebuild images, purge everything and then rerun?

charts/store/values.yaml
bartvanb: OK fair enough, I guess if we document it, both are fine so let's leave it like this!

bartvanb (approved):
I guess my trouble from running this comes from the fact that I've built custom images in order to work with keycloak. But seeing your screenshots, I am confident it's working how I'd like it to work, so I'm fine with merging it and then I'll simply use it later :-)

Issue #1940, opened 2025-05-13
[Bug] unit tests failing


dsmits:
Unit tests in branch `release/4.11` are failing. Or actually, the tests are not failing but there is an error message after the report that everything runs ok. Also, the job hangs until infinity.

I have run a git bisect and narrowed it down to a single commit where `main` is merged into `release/4/11`. I don't find any clear changes that would have caused this issue.

The stacktrace complains about there not being a table "authenticatable". The error occurs in the method [`handle_password_recovery`](https://github.com/vantage6/vantage6/blob/e99bb54ad0ffacffaebf6405b2e867f5fb52f2f4/vantage6-server/vantage6/server/resource/recover.py#L115). This method is called by [`recover.post`](https://github.com/vantage6/vantage6/blob/e99bb54ad0ffacffaebf6405b2e867f5fb52f2f4/vantage6-server/vantage6/server/resource/recover.py#L326) but it runs in the background. It seems like it is not relevant to the tests because the tests succeed. I wonder if this background task should have been prevented somehow.

## Stacktrace
```python
Traceback (most recent call last):
  File "/home/dsmits/projects/CARRIER/vantage6/venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1900, in _execute_context
    self.dialect.do_execute(
  File "/home/dsmits/projects/CARRIER/vantage6/venv/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: no such table: authenticatable

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/home/dsmits/projects/CARRIER/vantage6/vantage6-server/vantage6/server/resource/recover.py", line 154, in _handle_password_recovery
    user = User.get_by_username(username) if username else User.get_by_email(email)
  File "/home/dsmits/projects/CARRIER/vantage6/vantage6-server/vantage6/server/model/user.py", line 223, in get_by_username
    result = session.query(cls).filter_by(username=username).one()
  File "/home/dsmits/projects/CARRIER/vantage6/venv/lib/python3.10/site-packages/sqlalchemy/orm/query.py", line 2870, in one
    return self._iter().one()
[...]

PR #1943, opened 2025-05-15
Fix import order

bartvanb:
Aim is to fix unittests on `release/4.11` #1940. Not done yet


wbaccinelli (approved):
I admit that I don't fully understand what prometheus does, but it fixed the problem and looks good.

PR #1823, opened 2025-03-18
Change/merge federated central type with step type

bartvanb:
Note that we should merge this first and then #1820

PR #1945, opened 2025-05-16
Enabling support for central functions with an arbitrary number of nodes on the dev environment

hcadavid:
This PR enables a proper integration of an arbitrary number of nodes within the dev environment by (1) ensuring that the proxy of each one of the N nodes setup via devspace can be reached through the node Service, and (2) by fixing the proxy client so that now works with the sessions-enabled v6-server. Addresses issue #1913

To this end, the approach (based on  @bartvanb suggestion) is as follows:

- The node_proxy_port variable is now part of the Node configuration file.
- The node (on the _init_ method) uses the port defined in this variable to initialize the proxy.
- The behavior of retrying new random ports when the default one is not available was removed to ensure consistency in the dev. environment, and also considering that it may be more
  convenient for a sysadmin to be aware that the default port is being used (e.g., another node instance may be still running there).
- The load_fixtures script was updated so that it set these ports to each node in sequence, that is, the first node will run on port NODES_STARTING_PORT_NUMBER, the second on
  NODES_STARTING_PORT_NUMBER+1, the third on NODES_STARTING_PORT_NUMBER+2, and so on. That is, for the dev environment to work, the host must have a range of ports available: [NODES_STARTING_PORT_NUMBER .. NODES_STARTING_PORT_NUMBER+NUMBER_OF_NODES-1]
- Based on the NUMBER_OF_NODES and NODES_STARTING_PORT_NUMBER values (defined in the devspace config), the ports on the Node service are dynamically included. 
- The helm template of the node Service is defined in a way that it open multiple ports only when working in the dev environment. When the  NUMBER_OF_NODES and NODES_STARTING_PORT_NUMBER are not available (when using helm directly, no devspace), the default port defined in values.yml is used.

To check that each job was talking to the right node, I used this [diagnostics algorithm](https://github.com/hcadavid/v6-sessions-k8s-diagnostics/blob/48efc9da4bcf85e4404f169f37503c902b082253/v6-session-basics/partial.py#L195) (note the port 7654 used in the first node and the 7658 used on the fifth one):

image

image

 I also tested (with a [python client](https://github.com/hcadavid/v6-sessions-k8s-diagnostics/blob/main/notebooks/session-csv-extract-test.ipynb)) this implementation of [central average](https://github.com/hcadavid/v6-sessions-k8s-diagnostics/blob/48efc9da4bcf85e4404f169f37503c902b082253/v6-session-basics/partial.py#L251) and worked as expected:

image

However, as seen above, there is still one limitation: I need to pass the session_id to the central function so that it can also pass it to the client (the "1" next to "Age" argument is the session_id). I know the proxy client should set this value under the hood, but I'm leaving it as is for the time being in the interest of time.

@frankcorneliusmartin I also [explored the alternative](https://github.com/vantage6/vantage6/tree/fix/release/5.0/jobs2proxycomm-headless-srv) of switching from a K8S-deployment + a K8S service to K8S-StatefulSet + K8S Headless service we discussed previously. This allows to reach the POD running the Nodes directly -without the need of setting up port-forwarding on the service. However, this setup is not compatible with devspace. In DevSpace it is not possible to create the ephemeral development PODs as part of the StatefulSet. So when running the dev environment with this, the StatefulSet is created with the node POD, but it is later destroyed by running in conflict withthe ephemeral POD (which is not part of the StatefulSet).


dev/load-basic-fixtures.py
```diff
@@ -32,11 +32,19 @@
 parser.add_argument(
     "--task-namespace", type=str, default="vantage6-tasks", help="Task namespace"
 )
+parser.add_argument(
+    "--starting-port-number",
+    type=int,
+    default=7654,
+    help="The port number to be allocated for the first node. Additional nodes will use consecutive ports incremented by 1 from this value.",
```
bartvanb, line 39: Please specify that the port number is allocated to the *proxy server* of the first node.

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -389,6 +390,9 @@ def create(
                 "description": description,
                 "organizations": organization_json_list,
                 "databases": self.parent.databases,
+                "session_id": session,
+                "method": input_["method"],
```
bartvanb, line 394: In #1823, we moved method out of the input field (it fixes #809). I am wondering if method here is still part of the `input_` or if it is (as I expect) a separate field. I see now that the AlgorithmClient was not correctly updated in #1823 so that should be done here. Note that the docstring should be updated accordingly.

Also, you tested these changes by building a new algorithm right? I'm surprised that you didn't encounter this there (maybe you specified the method both inside and outside of the input_?)

vantage6-node/vantage6/node/__init__.py
```diff
@@ -169,49 +166,34 @@ def __proxy_server_worker(self) -> None:
...
-                exit(1)
+        try:
+            os.environ["PROXY_SERVER_PORT"] = str(node_proxy_port)
+            http_server.serve_forever()
 
-        if not port_assigned:
-            self.log.error(
-                f"Unable to assing a port for the node proxy after {try_number} attempts"
+        except OSError as e:
+            self.log.info(
```
bartvanb, line 189:
```suggestion
            self.log.error(
```

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -343,7 +343,11 @@ def run(
         RunStatus
             Returns the status of the run
         """
-        self.log.debug("Setting up algorithm run %s", run_id)
+        init_org_ref = task_info.get("init_org", {})
+        init_org_id = init_org_ref.get("id") if init_org_ref else None
+        self.log.debug(
+            f"[Algorithm job run {run_id} -requested by org {init_org_id}] Setting up algorithm run:{self.ctx}"
```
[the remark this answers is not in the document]
bartvanb: Finally, I don't think you want to print `self.ctx`?

vantage6-node/vantage6/node/k8s/run_io.py
```diff
@@ -8,6 +8,12 @@
 import pyarrow.parquet as pq
 import pandas as pd
 
+import concurrent.futures.thread  # Defensive import to avoid "can't register atexit after shutdown" runtime errors
+
+# Pyarrow lazily import concurrent.futures.thread during runtime,
+# which attempts to register an atexit handler.
+# Importing it upfront prevents late imports during Python interpreter shutdown.
```
bartvanb, line 15: Could you put the explanation above the import? And could you suppress the linting error with something like `# pylint: disable=something`? That should also remove some Codacy issues

bartvanb (commented):
Minor comments, nice that we don't require too much changes any more to get central functions to work!

CunliangGeng (changes requested):
The changes match what you described, looking good to me.
But one thing is missing: the devspace allows user to manually add more nodes (see line 363 of devspace.yaml) by adding a new node config file to the `dev/.data` folder, then a new node container will be started (it seems all node containers are restarted). In this case, the new node containers won't have a Service for it, then the users have to manually redeploy a Service to make it work. Can you automate this process?

dev/load-basic-fixtures.py
hcadavid: Fixed on a new commit

vantage6-node/vantage6/node/__init__.py
```diff
@@ -169,49 +166,34 @@ def __proxy_server_worker(self) -> None:
...
+                f"Error while trying to start the proxy server at {proxy_host}:{node_proxy_port}"
             )
+            self.log.info(
+                f"Check that port {node_proxy_port} is not being used by another process."
+            )
+            self.log.info("%s: %s", type(e), e)
+            self.log.info("Shutting down the node...")
```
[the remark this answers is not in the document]
hcadavid: Fixed on a new commit

vantage6-node/vantage6/node/k8s/container_manager.py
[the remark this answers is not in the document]
hcadavid: Replaced with %s style, removed the self.ctx from the log message

vantage6-node/vantage6/node/k8s/run_io.py
hcadavid: I documented this further in issue https://github.com/vantage6/vantage6/issues/1950, and referenced it from the code. Including a flag to disable the unused-import linter warning

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
hcadavid: Good catch! I added the `method` argument to the `algorithm/client/__init__` method and updated the json_body accordingly (no longer taking it from `input_`). Now in my [test algorithm](https://github.com/hcadavid/v6-sessions-k8s-diagnostics/blob/e0f7432714e6eb10e736ba0c83a0bbed61e30e9d/v6-session-basics/partial.py#L249) the client can be used within a central function like this:

```python

    task = client.task.create(name="central-fedavg", description="subtask",
                    organizations=ids,
                    method="federated_avg",             
                    input_={
                        "args": ["Age"],
                        "kwargs": {}
                    }, session=session_id)
```

Two questions for @frankcorneliusmartin and @bartvanb :

- Should the session also be implicit (setting under the hood the same session used by the central function)? If so, I'm not sure where it could be taken from.
- The body of the request of the proxy client has `federated compute` as a fixed action. I'm not sure this should be an argument... for instance, is it possible that a central function calls another central function?


```json
json_body = {
    "name": name,
    "image": self.parent.image,
    "collaboration_id": self.parent.collaboration_id,
    "description": description,
    "organizations": organization_json_list,
    "databases": self.parent.databases,
    "session_id": session,
    "method": method,
    "action": "federated compute",
}

```

hcadavid:
Thanks @CunliangGeng for pointing this out. You are right, in this case the manually added nodes will start, but their jobs won't be able to `talk' back to the node's server-proxy. I don't think this can be automated, as in that case the node port is arbitrarily defined by the user in the config file (if using the change proposed in this PR, where the proxy node is defined in the config file), or randomly defined at runtime by the node (if using the traditional approach).

@frankcorneliusmartin in the code you mention this is a possibility, but is this manual addition of nodes something desirable?


> The changes match what you described, looking good to me. But one thing is missing: the devspace allows user to manually add more nodes (see line 363 of devspace.yaml) by adding a new node config file to the `dev/.data` folder, then a new node container will be started (it seems all node containers are restarted). In this case, the new node containers won't have a Service for it, then the users have to manually redeploy a Service to make it work. Can you automate this process?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
bartvanb: > Should the session also be implicit (setting under the hood the same session used by the central function)? If so, I'm not sure where it could be taken from.

Yes I think a child task should always take the session of their parent. This type of thing would normally be present in the algorithm container token I think (see [here](https://github.com/vantage6/vantage6/blob/release/5.0/vantage6-server/vantage6/server/resource/token.py#L351) - doesn't seem to be in there yet. @frankcorneliusmartin Do you remember how exactly algorithm tokens work on v5 now?

> The body of the request of the proxy client has federated compute as a fixed action. I'm not sure this should be an argument... for instance, is it possible that a central function calls another central function?

We have claimed to support such a 'grandparent' task structure in the past but we never encountered that someone wanted to do it. I don't have a strong opinion for either, but if it's settable, I think it should have federated compute as default

bartvanb (approved):
I am OK with merge after resolving my minor comments. I should note that I have not tested this so it would be nice if Cunliang or Frank can do that before we merge it

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -359,13 +365,20 @@ def run(
 
         # Verify that an allowed image is used
         if not self.is_docker_image_allowed(image, task_info):
-            self.log.critical(f"Docker image {image} is not allowed on this Node!")
+            self.log.critical(
+                "[Algorithm job run %s -requested by org %s] Docker image %s is not allowed on this Node!",
```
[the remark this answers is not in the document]
hcadavid: Fixed

devspace.yaml
frankcorneliusmartin, line 1: I know the variable is already quite long but I would like to have the word proxy in there? 
`NODES_STARTING_PROXY_PORT_NUMBER` to emphasize that this is for the proxy

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -389,6 +393,9 @@ def create(
                 "description": description,
                 "organizations": organization_json_list,
                 "databases": self.parent.databases,
+                "session_id": session,
+                "method": method,
+                "action": "federated compute",
```
frankcorneliusmartin, line 398: Cant we use the enum for this? https://github.com/vantage6/vantage6/blob/99518256a61fa342dea1ca918ca7f7dd7c3ba25c/vantage6-common/vantage6/common/enum.py#L36-L48

frankcorneliusmartin (approved):
Some minor things to address, after that feel free to merge this PR.

CunliangGeng:
@frankcorneliusmartin  Can you ask this comment (https://github.com/vantage6/vantage6/pull/1945#issuecomment-2890976101)  before merging the PR?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
hcadavid: Done

frankcorneliusmartin:
> @frankcorneliusmartin in the code you mention that this is a possibility, but is this feature of manually adding nodes something desirable? Perhaps the nodes can be initialized through devspace, and then -once the ports are configured- the developer can customise the node settings and restart the dev environment?

I think this is desired as this gives flexibility to test different configurations? However I also see that this is not the most important thing to focus on, so I propose we store this request in a separate issue

vantage6-node/vantage6/node/globals.py
```diff
@@ -53,5 +53,9 @@
 DATABASE_BASE_PATH = "/app/databases/"
 
 # Must be consistent with node pod configuration
-PROXY_SERVER_HOST = "http://v6proxy-subdomain.vantage6-node.svc.cluster.local"
-PROXY_SERVER_PORT = 4567
+# http://vantage6-node-node-service.vantage6-dev.svc.cluster.local:7654
+PROXY_SERVER_HOST = "http://vantage6-node-node-service.vantage6-dev.svc.cluster.local"
```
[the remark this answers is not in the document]
hcadavid: The PROXY_SERVER_HOST is now dynamically defined considering {{ .Release.Name }} and the namespace. As there is no way to have default proxy server host (as it depends on the namespace) it is now mandatory to set this env. variable before starting the node.

PR #1947, opened 2025-05-19
Modify port numbers that server and store databases are exposed at in…

bartvanb:
… the dev env

Two reasons for this change:
- for some miraculous reason, 5432 didn't work on my setup (maybe a weird WSL thing) - changing it to another port number solved the issue
- we have all our dev components exposed in ports 76xy


CunliangGeng (approved):
LGTM!

PR #1958, opened 2025-05-21
Checking if devspace init pipeline succeeds. Without checking, server…

bartvanb:
… was not populated upon errors

This is just an idea - let me know what you think. We could expand on this by allowing user to 'continue anyway' or so


CunliangGeng (approved):
Nice addition of error checking!

PR #1961, opened 2025-05-26
Ensure purging data works in WSL

bartvanb:
Similar to what we changed when creating the directories is now also required for deleting them :-)

### DIFF
--- a/dev/purge_data.sh
+++ b/dev/purge_data.sh
@@ -26,6 +26,7 @@ replace_wsl_path() {
   if [[ "$path" == $WSL_REFERENCE_PATH* ]]; then
     path="${WSL_REGULAR_PATH}${path#${WSL_REFERENCE_PATH}}"
   fi
+  echo "$path"
 }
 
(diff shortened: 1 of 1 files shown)


dev/purge_data.sh
```diff
@@ -17,14 +17,26 @@ usage() {
...
+
 if [[ "$1" == "--help" ]]; then
   usage
 fi
 
-POPULATE_MARKER=$1
-TASK_DIRECTORY=$2
-SERVER_DATABASE_MOUNT_PATH=$3
-STORE_DATABASE_MOUNT_PATH=$4
+POPULATE_MARKER=$(replace_wsl_path "$1")
```
CunliangGeng, line 36: I did not do wsl replacement for this marker file in `populate_server.sh`, since it's assumed to be in `.devspace/...`. But I think it's more robust to do it. Can you add it?

frankcorneliusmartin (approved):

PR #1962, opened 2025-05-26
Change/#1955 fix node image validation

frankcorneliusmartin:
Have a look at the issue. I've posted a question there that I am not sure about.


bartvanb (approved):
Looks good. Only thing I want to note, is that there are still a lot of other Docker things in the `common/docker/addons.py`. Is there already an issue to remove those before v5.0? It's probably part of the CLI redo

Issue #1741, opened 2025-02-18
[Feature Request] [k8s] Integrating/adapting network policies from the PoC


hcadavid:

bartvanb:
![Image](https://github.com/user-attachments/assets/e34e17ea-d518-45c5-a330-f517a0f25bb6)

Just to document what is needed here

hcadavid:
Addressed in https://github.com/vantage6/vantage6/pull/1980

Issue #1259, opened 2024-05-13
[Feature Request] Action based provisioning of algorithm containers


frankcorneliusmartin:
**Problem description**

We could improve the security by applying the following:

* [x] `Compute` action containers should not be able to reach the original data source. First #154 and #943 need to be implemented
* [x] `Data extraction` and `Pre processing` should not be able to reach any thing in the network(?) Should they be able to use the whitelising / ssh-tunnel features?
* [x] Token only needs to be requested for the compute container
* [x] Only `Compute` results are returned to the vantage6-server

Maybe other rules could also be applied to different actions.

(note that this issue is referenced (#1247) to in the code, so you can search for points of interest)

frankcorneliusmartin:
Implemented:

* Only Compute results are returned to the vantage6-server
* Token only needs to be requested for the compute container

The other two require more work

frankcorneliusmartin:
@hcadavid I know you are working on #1741, can you have a look at

> Data extraction and Pre processing should not be able to reach any thing in the network(?) Should they be able to use the whitelising / ssh-tunnel features?

hcadavid:
@frankcorneliusmartin sure, as the 'action type' is given to the `run` method of the `container_manager` we can add metadata to the job accordingly (so that we can define step-specific network policies). I need to complete https://github.com/vantage6/vantage6/pull/1914 before integrating these policies. 

However, the policies tested on the PoC (not yet on the release/5.0 branch) already isolate the Job nodes, regardless of their type. How to include the whitelisting and ssh-tunnels features are is yet to be defined. 

For instance, it wouldn't be possible to use the 'whitelisted' hosts of the traditional v6-node configuration file, as the PODs can't define their own policies at runtime. If the node setup is made through helm-charts, the network policies should be defined as a helm template, and the whitelisted hosts would be defined on its corresponding `values.yaml`. Hence, the person doing the setup would have to edit this additional file. @frankcorneliusmartin @bartvanb  What do you think?

Issue #1740, opened 2025-02-18
[Feature Request] [k8s] Service mesh (Isteo) alternative for VPN


frankcorneliusmartin:
**Problem description**
We should implement the VPN flow using Isteo

hcadavid:
Note:

When implementing the Isteo layer, federated_ and central_ compute may need to communicate with containers on other (k8s and v6) nodes. Then, existing network policies should be revisited:

https://github.com/vantage6/vantage6/blob/1cb86ad1994efa1c41ecd97f606b8a7a06fbc11f/charts/node/templates/job-network-policy.yml#L127-L146

Issue #1739, opened 2025-02-18
[Feature Request] [k8s] Replace SSH Tunnel, Whitelisting and docker services


frankcorneliusmartin:
**Problem description**
These have been removed from the code base. It should be documented on how to approach these use cases with the k8s node

frankcorneliusmartin:
maybe a solution to replace the SSH tunnel: https://stackoverflow.com/questions/65123401/how-to-access-hosts-localhost-from-inside-kubernetes-cluster

bartvanb:
Whitelisting supported, SSH tunnel is not necessary because whitelisting is much better, docker services is now a k8s service that is automatically available if you spin it up in the internal k8s dns

PR #1980, opened 2025-05-28
Action-based network policies 

hcadavid:
This PR enables action-tailored network policies for the job pods (https://github.com/vantage6/vantage6/issues/1259 and https://github.com/vantage6/vantage6/issues/1741) as follows:

- The container manager now set as a 'role' label the algorithm step type, so that the network policies (within the tasks namespace) can be tailored for each one.
- A first implementation of such pairing rules are included, where by default all inbound/outbound traffic is disabled, and the central and federated compute actions are defined as special cases: the central action can reach the node proxy, and the data-extraction one can reach a whitelisted database (how to enable access to white-listed external databases is not enabled but shown in comments).
- The values of the ActionType enumerations were renamed (removing spaces) so that these can be used as the role label previously described.

**Note**: for an algorithm to work with this PR the [AlgorithmStepType](https://github.com/vantage6/vantage6/blob/4cb75fb1e4de936940fd4d7b66205c3179507350/vantage6-common/vantage6/common/enum.py#L36), the algorithm must be rebuilt with the new enumeration values (which uses underscores instead of spaces) - this is not on the current version of `commons` (`v.5.0.0-a.17`).

The following are the logs that result from running a ['federated' diagnostics algorithm](https://github.com/hcadavid/v6-sessions-k8s-diagnostics/blob/79b80c3b14cd39bfae92cb4e3766e392f906c8ca/v6-session-basics/partial.py#L270). Note the assigned 'role' labels, and how it impacts the result of the diagnostics: a central_compute function can't access only the proxy, while the federated_compute can't access neither external networks nor the internal k8s one.

image
```
info > wrapper for v6-session-basics
info > Reading input file /app/input
info > Dispatching ...
info > Module 'v6-session-basics' imported!
[[32minfo [0m] - Validating function action: central_compute
HOST env var: http://vantage6-node-node-service.vantage6-dev.svc.cluster.local
PORT env var: 7654
>>>>>Proxy FQDN vantage6-node-node-service.vantage6-dev.svc.cluster.local solved as 10.152.183.21
Host architecture:x86_64
IPv4 Addresses:
lo: 127.0.0.1
eth0: 10.1.71.46

IPv6 Addresses:
lo: ::1
eth0: fe80::5c90:e8ff:fedb:16a1%eth0
External DNS reachable (socket connection to port 53 test) :ENABLED
HTTP connection failed: HTTPConnectionPool(host='www.google.com', port=80): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7a9224fb5150>: Failed to establish a new connection: [Errno 101] Network is unreachable'))
Internet access (http connection test) :DISABLED
info > Checking if the FQDN of the node proxy (vantage6-node-node-service.vantage6-dev.svc.cluster.local:7654) can be resolved... 
info > FQDN of the node proxy (vantage6-node-node-service.vantage6-dev.svc.cluster.local:7654) resolved as 10.152.183.21... 
info > Port 7654 can be opened on the proxy (vantage6-node-node-service.vantage6-dev.svc.cluster.local) IP address: 10.152.183.21
V6-proxy status :REACHABLE at vantage6-node-node-service.vantage6-dev.svc.cluster.local:7654
```

image

````
info > wrapper for v6-session-basics
info > Reading input file /app/input
info > Dispatching ...
info > Module 'v6-session-basics' imported!
[[32minfo [0m] - Validating function action: federated_compute
HOST env var: http://vantage6-node-node-service.vantage6-dev.svc.cluster.local
PORT env var: 7654
Unable to resolve Proxy FQDN vantage6-node-node-service.vantage6-dev.svc.cluster.local - DNS disabled for this POD
Host architecture:x86_64
IPv4 Addresses:
lo: 127.0.0.1
eth0: 10.1.71.36

IPv6 Addresses:
lo: ::1
eth0: fe80::38fe:15ff:fe61:51a0%eth0
External DNS reachable (socket connection to port 53 test) :ENABLED
HTTP connection failed: HTTPConnectionPool(host='www.google.com', port=80): Max retries exceeded with url: / (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x72bbf50b4dc0>: Failed to resolve 'www.google.com' ([Errno -3] Temporary failure in name resolution)"))
Internet access (http connection test) :DISABLED
info > Checking if the FQDN of the node proxy (vantage6-node-node-service.vantage6-dev.svc.cluster.local:7654) can be resolved... 
info > Unreachable proxy: FQDN could not be resolved
V6-proxy status :DISABLED or unreachable at vantage6-node-node-service.vantage6-dev.svc.cluster.local:7654
````

### CODE REVIEW

File: charts/node/templates/job-network-policy.yml [hunk 1]
```diff
@@ -1,12 +1,139 @@
...
+  - {}
+  # # This policy could be further refined to allow traffic to the v6-server only
+  # egress:  
+  # - to:
+  #   # Kubernetes NetworkPolicies do not support domain names or DNS-based rules directly
+  #   - ipBlock:
+  #       cidr: 203.0.113.25/32  # Replace with your actual Postgres server IP
+  #   ports:
+  #   - protocol: TCP
+  #     port: 9876
```
bartvanb, line 23: This part is not very clear to me: would this enable traffic to the v6 server directly? Why would we want to allow that if the node can already communicate via the proxy? So can we not better remove this?

If it turns out we need this, maybe it's better to create an issue to make this configurable?

charts/node/templates/job-network-policy.yml
```diff
@@ -1,12 +1,139 @@
...
+
+
+---
+
+# DEFAULT policies for the 'tasks' namespace (inboud/outbound traffic blocked)
+# DO NOT CHANGE
```
bartvanb, line 29: I think this 'Do not change' line should be removed. All things configurable should be collected to e.g. set in the CLI - all other stuff should not be changed de facto

charts/node/templates/job-network-policy.yml
```diff
@@ -1,12 +1,139 @@
...
+  # # Example of how to 'whitelist' an external PostgreSQL database
+  # # (outside the k8s cluster)
+  # egress:  
+  # - to:
+  #   # Kubernetes NetworkPolicies do not support domain names or DNS-based rules directly
+  #   - ipBlock:
+  #       cidr: 203.0.113.25/32  # Replace with your actual Postgres server IP
+  #   ports:
+  #   - protocol: TCP
+  #     port: 5432
```
bartvanb, line 118: Should we link the issue for whitelisting here? Or place a comment there that something like this should be incorporated there?

charts/node/templates/job-network-policy.yml
```diff
@@ -1,12 +1,139 @@
...
+#   namespace: {{ .Values.node.taskNamespace }}
+# spec:
+#   podSelector:
+#     matchLabels:
+#       role: federated_compute
+#   policyTypes:
+#   - Ingress
+#   - Egress
+#   ingress: []
+#   egress: []
```
bartvanb, line 139: As far as I'm concerned, it's better to delete stuff that we're not using

charts/node/templates/node-deployment.yml
```diff
@@ -5,6 +5,8 @@ metadata:
   namespace: {{ .Values.namespace | default .Release.Namespace }}
   labels:
 {{ include "node.labels" . | indent 4 }}
+    name: {{ .Values.namespace | default .Release.Namespace }}
+    app: v6-node-server-proxy
```
bartvanb, line 9: I think I'm missing something, but why proxy? This is generic node deployment right?

bartvanb (commented):
Looks pretty good, some questions

charts/node/templates/job-network-policy.yml
hcadavid: Actually what this allows is the proxy to reach out the server. The network policy, as defined, allows the node-proxy to reach any other host on the Internet, so that it can reach the corresponding v6-server. This example in the comments is just to point out that the rule could be further refined to ensure that the node-proxy can _only_ reach one specific server. This may desirable for security considerations (e.g., a backdoor attack on the node wouldn't be able to make its way out). I updated the comments to make this more clear.

charts/node/templates/job-network-policy.yml
hcadavid: Referencing this PR on https://github.com/vantage6/vantage6/issues/1739

bartvanb: Thanks! I would expect the whitelisting to be configurable somehow but whether that is via a config file or helm chart is a bit bigger scope than this PR

charts/node/templates/job-network-policy.yml
bartvanb: I think that in a v5 session, a federated compute will take the data from a local parquet file. It should otherwise be completely isolated from the internet so I don't see us adding policies for it soon (and if so, we could take the extraction / central compute rules as example). So I still prefer deleting it.

frankcorneliusmartin: This is an interesting part. 

In general I agree with Bart, both `central_` and `federated_` compute should only be able to access the session data within the session storage. However... when we are going to implement the Isteo layer I expect that `federated_` and `central_` compute are going to communicate with containers on other (k8s and v6) nodes. Maybe good to move this code to that issue: https://github.com/vantage6/vantage6/issues/1740

charts/node/templates/job-network-policy.yml
```diff
@@ -1,12 +1,146 @@
...
+  {{- end }}
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
+    ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
```
frankcorneliusmartin, line 93: I am not sure what these ports represent. Are these the ports the the `central_` can reach? In that case these should be the ports that the node service is listening on? 

I am just wondering why these are hard coded?

frankcorneliusmartin (commented):
Som small questions/comments.

We should include the diagnostics image in the local algorithm store so we can easier validate the network policies. Did you add them in your setup already?

charts/node/templates/job-network-policy.yml
hcadavid: @frankcorneliusmartin the interpretation of this rule also requires considering the matchLabel (kubernetes.io/metadata.name: kube-system).  This is required to enable the job pods to access K8S's internal DNS, which is always listening on port 53. If this rule is not included, the modules within the job PODs won't be able to resolve the Node-service's FQDN. 

```
  - to:
    - namespaceSelector:
        matchLabels:
          kubernetes.io/metadata.name: kube-system
    ports:
    - protocol: UDP
      port: 53
    - protocol: TCP
      port: 53
```

An additional rule for actually enabling the communication with the Node, given the port(s) defined for it, is dynamically generated in the template:

https://github.com/vantage6/vantage6/blob/1cb86ad1994efa1c41ecd97f606b8a7a06fbc11f/charts/node/templates/job-network-policy.yml#L65-L84

hcadavid:
> Som small questions/comments.
> 
> We should include the diagnostics image in the local algorithm store so we can easier validate the network policies. Did you add them in your setup already?

Nope, I'll do it after making the changes suggested by both of you.

charts/node/templates/job-network-policy.yml
```diff
@@ -1,12 +1,146 @@
...
+  # # This policy could be further refined to allow outbound traffic ONLY to an
+  # # specific v6-server.
+  # egress:  
+  # - to:
+  #   # Kubernetes NetworkPolicies do not support domain names or DNS-based rules directly
+  #   - ipBlock:
+  #       cidr: 203.0.113.25/32  # v6-server IP address
+  #   ports:
+  #   - protocol: TCP
+  #     port: 9876 # v6-server port
```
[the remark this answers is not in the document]
hcadavid: I removed the commented code example and created issue https://github.com/vantage6/vantage6/issues/2037 to enable the dynamic generation of 'egress' rules for data_extraction tasks based on the settings defined in config.yml.

Issue #1675, opened 2025-01-16
[Feature Request] [k8s] Remove unused code


frankcorneliusmartin:
**Problem description**
Since we removed quite a bit of code when moving to kubernetes we might have ended up with deadcode. We should remove this luggage.

- docker addons?
- Functions used by the task manager
- CLI commands (e.g. `--mount-src`, etc.)
- ...

bartvanb:
Related #1735

PR #1981, opened 2025-05-28
Change/#1675 remove unused code from docker

frankcorneliusmartin:


Makefile
```diff
@@ -281,8 +241,13 @@ publish:
 	cd vantage6-server && make publish
 	cd vantage6-algorithm-store && make publish
 
+# Default test subpackages if none specified
+TEST_SUBPACKAGES ?= common,cli,algorithm-store,server
+
 test:
-	coverage run --source=vantage6 --omit="utest.py","*.html","*.htm","*.txt","*.yml","*.yaml" utest.py
+	$(eval TEST_ARGS := $(shell echo $(TEST_SUBPACKAGES) | tr ',' ' ' | sed 's/^/--/;s/ / --/g'))
```
bartvanb, line 248: I think this command could be simplified: we probably can do both without the `eval` and the `shell`, right?

utest.py
```diff
@@ -2,35 +2,60 @@
...
+        success_common = run_tests(common_test_suites)
+        success = success and success_common
 
     # run CLI tests
-    cli_test_suites = find_tests(str(Path(__file__).parent / "vantage6"))
-    success_cli = run_tests(cli_test_suites)
+    # if args.cli or args.all:
+    #     cli_test_suites = find_tests(str(Path(__file__).parent / "vantage6"))
+    #     success_cli = run_tests(cli_test_suites)
+    #     success = success and success_cli
```
bartvanb, line 42: This should not be commented out - we will forget to reinstate it. Why did you?

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
```diff
@@ -408,9 +408,9 @@ def create(
                 json=json_body,
             )
 
-    class VPN(ClientBase.SubClient):
+    class Network(ClientBase.SubClient):
```
bartvanb, line 411: It makes more sense to me to delete this entire class. The VPN endpoints that are being called here will be removed in v5. So maybe just remove all this AND remove the endpoints in the server?

vantage6-algorithm-tools/vantage6/algorithm/tools/wrappers.py
bartvanb, line 1: I feel that we could leave this in the infrastructure code. If we do, algorithm developers could simply do 

```
# in their __init__.py
from vantage6.algorithm.tools.wrappers import load_data, load_csv_data, *  # or one of them
```
And they would have all the predefined wrappers in their image without needing to write them themselves. 

That is especially useful if using `session_restrict_to_same_image = True`

Note that we could do a similar thing for preprocessing actually

vantage6-client/vantage6/client/__init__.py
```diff
@@ -808,7 +808,7 @@ def list(
             is_online: bool, optional
                 Filter on whether nodes are online or not
             ip: str, optional
-                Filter by node VPN IP address
+                Filter by node internal IP address
```
bartvanb, line 811: Maybe better to create an issue to simply delete the `ip` field. Or check if we're actually using it

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -23,35 +23,6 @@
...
-    kill_now = False
-
-    def __init__(self) -> None:
-        signal.signal(signal.SIGINT, self.exit_gracefully)
-        signal.signal(signal.SIGTERM, self.exit_gracefully)
-
-    # pylint: disable=unused-argument
-    def exit_gracefully(self, *args) -> None:
-        """Set kill_now to True. This will trigger the container to stop"""
-        self.kill_now = True
```
bartvanb, line 38: Don't we need this in the (to be reinstated) functionality to kill tasks?

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -23,35 +23,6 @@
...
-        docker_client = docker.from_env()
-        docker_client.ping()
-    except Exception as exc:
-        log.error("Cannot reach the Docker engine! Please make sure Docker is running.")
-        log.exception(exc)
-        log.warning("Exiting...")
-        exit(1)
-
-
 def running_in_docker() -> bool:
```
bartvanb, line 26: I think this function can and should also be deleted - note that that does require changes to other files

vantage6-common/vantage6/common/docker/addons.py
```diff
@@ -23,35 +23,6 @@
 log = logging.getLogger(logger_name(__name__))
 
 
-class ContainerKillListener:
```
bartvanb, line 26: The imports in line 17+18 are the ONLY docker functions we are still using from docker after this PR (and the other one with node image policy) is merged. It would be nice to remove the docker package dependency completely (maybe separate issue?)

vantage6-server/vantage6/server/resource/port.py
```diff
@@ -46,9 +46,9 @@ def setup(api: Api, api_base: str, services: dict) -> None:
         resource_class_kwargs=services,
     )
     api.add_resource(
-        VPNAddress,
-        api_base + "/vpn/algorithm/addresses",
-        endpoint="vpn_address",
+        InternalAddress,
```
bartvanb, line 49: These endpoints shouldn't really do anything any more. We should remove them (and also the `Port` sqlalchemy model and everything related to it)

bartvanb (changes requested):
Nice work, I have a number of comments and questions, some of which may be turned into separate issues to be tackled later

vantage6-common/vantage6/common/docker/addons.py
frankcorneliusmartin: I agree, this was already logged into https://github.com/vantage6/vantage6/issues/1675

vantage6-algorithm-tools/vantage6/algorithm/client/__init__.py
frankcorneliusmartin: Removed it: 11489d0e999c9ad61663d03f1a62cb7616a527d4

vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
```diff
@@ -545,17 +545,17 @@ def get(self, is_encrypted: bool = True) -> dict:
             }
 
     # TODO implement the get_addresses method before using this part
-    # class VPN(SubClient):
+    # class Network(SubClient):
```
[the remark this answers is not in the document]
frankcorneliusmartin: 11489d0e999c9ad61663d03f1a62cb7616a527d4

vantage6-server/vantage6/server/resource/port.py
frankcorneliusmartin: I removed it: 4b2d4dc2daa1fc9732f422a28d343a3e40c71605

vantage6-client/vantage6/client/__init__.py
```diff
@@ -841,14 +838,14 @@ def list(
...
                 collaboration = self.parent.collaboration_id
+
             params = {
                 "page": page,
                 "per_page": per_page,
                 "name": name,
                 "organization_id": organization,
                 "collaboration_id": collaboration,
                 "study_id": study,
-                "ip": ip,
```
bartvanb, line 851: Now that we are removing the VPN stuff, the IP field is not really in use at the moment I think. Maybe we should remove it in its entirety. However, we could also leave it in because we will want to re-add this functionality of course. Both options are fine with me, but if we leave it in for now, I do want it mentioned in the Istio issue that if we don't start using this field there, we should remove it.

bartvanb (approved):
One comment which you can solve by making further changes or commenting on existing issue and merge

vantage6-client/vantage6/client/__init__.py
frankcorneliusmartin: I don't understand, we removed it here right?

bartvanb: Sorry I guess I was unclear. What I was talking about was removing the `Authenticatable.ip` field entirely. It still exists in that model but is now not used I think (it might be set sometimes but nothing is done with its value)

bartvanb:
I unlinked the issue to be resolved by this PR because the work is not finished

PR #2053, opened 2025-06-18
Change/minor deployment changes part 2

frankcorneliusmartin:
Some difficulties/mistakes I encountered whilst deploying the node for the first time:

* Added the option to customize the `storageClass` and `size` for the tasks and (file based) database volumes
* The file based volume is only created when there are any file based databases (this resulted in an error previously)
* The `PV` (not namespaced) for the (file based) databases is automatically deleted when the `PVC` (namespaced) is deleted. We do not want to keep a volume with outdated databases. When the node is created again this volume should be freshly created and populated. I did *not* do this for the tasks volume as we might want to keep the task and session folders when we redeploy the node. So this still needs to be manually deleted if required.
* The `*-node-secrets` are only created when there is not an existing `*-node-secrets` file. This allows the secrets file to be created outside of our install (e.g. you can deploy a DB and immediately create the secret, this way you do not have to store the secret somewhere else in the meantime)


charts/node/templates/node-secrets.yml
```diff
@@ -1,16 +1,18 @@
+{{- if not (lookup "v1" "Secret" (.Values.namespace | default .Release.Namespace) (printf "%s-node-secrets" .Release.Name)) }}
```
bartvanb, line 1: If I understood this well, we only create a new secret if there is not already an old secret resource available.

Are we sure that we do not want to throw an error instead? This way old secrets could cause a deployment not to work, and it might not be easy to debug ("I am sure that I set the secret correctly now, why is it still not working :-@"). 
Isn't it better to make sure we remove everything properly in the CLI where appropriate?

My comment is of course also for the other files modified here. 
And finally: *if* we are going to change this, why does this PR only include the changes for the node and not all k8s resources?

charts/node/values.yaml
```diff
@@ -7,19 +7,27 @@ node:
...
-
   proxyPort: 7654
 
+  # Persistence configuration, these settings are both used for the file based databases
+  # and the task directory.
+  persistence:
+    tasks:
+      storageClass: local-storage
+      size: 10Gi
+      hostPath: /tmp/vantage6/tasks
```
bartvanb, line 26: unrelated to this PR but I'm wondering if `/tmp/` is a good place to store task data on a node. If the machine reboots, we lose them all I think (assuming Ubuntu VM)?

charts/node/templates/node-secrets.yml
frankcorneliusmartin: No I want to use this so that secrets can be created outside of vantage6 which could be a very common case when dealing with external databases:

> The *-node-secrets are only created when there is not an existing *-node-secrets file. This allows the secrets file to be created outside of our install (e.g. you can deploy a DB and immediately create the secret, this way you do not have to store the secret somewhere else in the meantime)

The only alternative that I have is that we create a boolean in the `values.yml` to not create the secret, but I find that quite ugly. Let me know what you think.

bartvanb: I understand the reasoning behind this now and agree that this is better.

So my only remaining question: should we also do this elsewhere?

PR #2063, opened 2025-06-20
Add `just_uri` option to `@data` decorator

lsago:
Sometimes algorithm developers want tighter control over how the data is loaded into a dataframe.


frankcorneliusmartin:
The code seems fine to me, but I wondered why you chose to do it in the same decorator and not just create a new one?

bartvanb:
Agreed, `@data_uri` sounds like an excellent idea

lsago:
Closing it :) I may open a new one for something like `@data_uri`. We've run into a related bug where the order of the databases is not predictable once they reach the partial function – was applying this patch for my algos. So, maybe a "key word" databases specification would be nice. So far I've just hackily worked around it in the algorithms we've been using, but will open a new issue for that as well.

Issue #1863, opened 2025-04-01
[Feature Request] UI sessions - support multiple dataframes


bartvanb:
**Problem description**
Multi-dataframe tasks are not supported in the UI yet

**Desired solution**
These should be supported

bartvanb:
Also, should multi-database extraction tasks be supported? Not sure we will encounter those

PR #2065, opened 2025-06-23
Feature/#2061 new dataframes decorator in the algorithm tools

frankcorneliusmartin:
* I updated the image `v6-session-basics` with additional functions. I did not want to break existing environments so I decided to duplicate functions. Once this PR is merged the `v6-session-basics` should be updated and the `dev/algorithm-store.json` should also be updated.
* Please have a thorough look at the UI code, as this was done with the help of chad (yes I did check the code ;-))

### DIFF
--- a/dev/v6-session-basic-algorithm-store.json
+++ b/dev/v6-session-basic-algorithm-store.json
@@ -8,9 +8,9 @@
   "functions": [
     {
       "name": "global_sum_dev_many",
-      "display_name": "Sum for multiple dataframes",
+      "display_name": "Sum for a given column in one or more dataframes",
       "standalone": true,
-      "description": "Compute the global sum of a federated dataset",
+      "description": "Compute the global sum of a given column in one or more dataframes",
       "ui_visualizations": [],
       "step_type": "central_compute",
       "arguments": [
@@ -143,7 +143,7 @@
     },
     {
         "name": "sum_many",
-        "display_name": "Sum for multiple dataframes",
+        "display_name": "Sum for a given column in one or more dataframes",
         "standalone": true,
         "description": "Sum one column for each dataframe",
         "ui_visualizations": [],


vantage6-algorithm-tools/vantage6/algorithm/decorator/data.py
```diff
@@ -97,29 +113,35 @@ def decorator(
...
-
-                dataframe_name = dataframes[i]
+            for source, requested_dataframes in zip(sources, dataframes_grouped):
                 # read the data from the database
-                info("Reading dataframe")
-                data_ = _read_df_from_disk(dataframe_name)
+
+                # if the source is not "many", we can just add the first (and only)
+                # dataframe to the arguments
+                if str(source).lower() != "many":
```
bartvanb, line 139: Could you use a constant for using "many" everywhere (or, as I suggest somewhere above, preferably "multiple")? Then, if we ever change it, it's not so much work

vantage6-client/vantage6/client/__init__.py
```diff
@@ -1913,9 +1916,15 @@ def _parse_arg_databases(databases: list[dict] | str) -> list[dict]:
...
+            # It is common to only specify a single level of databases, we assume
+            # that its not a many databases argument and convert it so that every
+            # requested dataframe is handled as a single argument in the algorithm.
+            if isinstance(databases, list) and not isinstance(databases[0], list):
+                databases = [[db] for db in databases]
+
+            for db in [db for sublist in databases for db in sublist]:
                 try:
                     label_input = db.get("label")
                 except AttributeError:
```
bartvanb, line 1930: The value error below here is no longer really accurate. But also "list of list of dicts" becomes a hard-to-read error. Maybe something like "Each database should be a string specifying a database label, or a dictionary with at least a 'label' key". Also might be nice to mention the misbehaving db in the error

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -802,8 +801,23 @@ def _create_volume_mounts(
             # available in the session.
             self._validate_dataframes(databases_to_use, run_io)
 
+            # group and sort on the position of the database in the argument list
+            databases_to_use_sorted = sorted(
+                databases_to_use, key=lambda x: x["position"]
+            )
+            grouped_databases = groupby(
+                databases_to_use_sorted, key=lambda x: x["position"]
+            )
```
bartvanb, line 810: Could you explain why we have to sort things? Before, we relied on the user providing proper order right?

vantage6-node/vantage6/node/k8s/container_manager.py
```diff
@@ -802,8 +801,23 @@ def _create_volume_mounts(
             # available in the session.
             self._validate_dataframes(databases_to_use, run_io)
 
+            # group and sort on the position of the database in the argument list
+            databases_to_use_sorted = sorted(
+                databases_to_use, key=lambda x: x["position"]
```
bartvanb, line 806: Does this work if the user did not provide a "position" key? There is no validation on whether the user provided that key I think?

vantage6-server/vantage6/server/resource/common/input_schema.py
```diff
@@ -430,7 +430,13 @@ def validate_databases(self, databases: list[dict] | None):
         """
         if databases is None:
             return  # some algorithms don't use any database
-        for database in databases:
+
+        if isinstance(databases, list) and not isinstance(databases[0], list):
+            raise ValidationError(
+                "Databases must be a list of lists of dictionaries or None"
+            )
```
bartvanb, line 437: I thought we were also going to allow something like `[["a, "b"], "c"]`? Or is that just what we accept in the client and convert there to this format?

vantage6-server/vantage6/server/resource/dataframe.py
```diff
@@ -287,7 +288,7 @@ def post(self, session_id):
                 method=extraction_details["method"],
                 organizations=extraction_details["organizations"],
                 # TODO FM 10-7-2024: we should make a custom type for this
-                databases=[{"label": source_db_label, "type": "source"}],
+                databases=[[{"label": source_db_label, "type": "source"}]],
```
bartvanb, line 291: I assume this also has to be changed for preprocessing tasks, or is the original format also accepted?

vantage6-ui/src/app/components/forms/compute-form/create-analysis-form.component.html
```diff
@@ -187,8 +187,10 @@
           <form class="form" [formGroup]="dataframeForm">
             <ng-template matStepLabel>{{ "task-create.step-dataframe.title" | translate }}</ng-template>
             <mat-form-field *ngIf="dataframes.length > 0" subscriptSizing="dynamic">
-              <mat-label>{{ "task.dataframe" | translate }}</mat-label>
-              <mat-select formControlName="dataframeId">
+              <mat-label>{{ "task.dataframe" | translate }}{{ isFirstDatabaseMany() ? 's' : '' }}</mat-label>
+              <mat-select
+                formControlName="dataframeId"
+                [multiple]="isFirstDatabaseMany()">
```
bartvanb, line 193: I think multiple-dataframe tasks are not really supported in the UI at present right? I guess that's why you only look at the first one

vantage6-ui/src/app/components/forms/compute-form/create-analysis-form.component.ts
```diff
@@ -382,9 +388,12 @@ export class CreateAnalysisFormComponent implements OnInit, OnDestroy, AfterView
   }
 
   async handleSubmit(): Promise<void> {
-    if (this.isSubmitting) return;
-    if (this.isFormInvalid()) return;
+    if (this.isFormInvalid()) {
```
bartvanb, line 391: I'd still include the check for `isSubmitting` - even if this function should be disabled in the HTML in that case (but the same actually also goes for isFormInvalid)

vantage6-ui/src/app/models/api/session.models.ts
```diff
@@ -12,8 +12,8 @@ export enum SessionSortProperties {
 }
 
 export enum SessionScope {
-  Collaboration = 'col',
-  Organization = 'org',
+  Collaboration = 'collaboration',
+  Organization = 'organization',
```
bartvanb, line 16: Have you checked if this has side effects? Is it used anywhere where it has to be matched to data coming from backend?

vantage6-ui/src/assets/localizations/en.json
```diff
@@ -344,7 +345,10 @@
           "short_header": "Only for UI"
         }
       },
-      "databases": "Databases",
+      "databases": {
+        "title": "Databases",
+        "many": "Can multiple databases be used?"
```
bartvanb, line 350: This question is maybe a bit vague for me. Maybe something like "Does the algorithm accept multiple dataframes in this single argument?"

bartvanb (changes requested):

dev/v6-session-basic-algorithm-store.json
```diff
@@ -114,7 +140,33 @@
           "description": "Dataframe to use for computing the sum"
         }
       ]
-    }
+    },
+    {
+        "name": "sum_many",
+        "display_name": "Sum Many",
```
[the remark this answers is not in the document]
frankcorneliusmartin: Same as the reply above.

vantage6-algorithm-tools/vantage6/algorithm/decorator/data.py
```diff
@@ -44,11 +47,11 @@ def _read_df_from_disk(df_name: str) -> pd.DataFrame:
     return pd.read_parquet(dataframe_file)
 
 
-def data(number_of_databases: int = 1) -> callable:
+def dataframe(*sources: str | int) -> callable:
```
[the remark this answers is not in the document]
frankcorneliusmartin: e4d7d501735b6367490264cd96d86e4d264709ee

vantage6-algorithm-tools/vantage6/algorithm/decorator/data.py
frankcorneliusmartin: I changed it here: 31fc8c0796827c66cc185cf2f2926720f05a74a6. In case you change it you still need to update the docstrings, fields in the db (optional) and the algorithm store UI.

vantage6-server/vantage6/server/resource/dataframe.py
frankcorneliusmartin: I am not sure what you mean? I also adjusted the preprocessing endpoint?:

```python
response, status_code = self.create_session_task(
            session=session,
            databases=[
                [
                    {
                        "dataframe_id": dataframe.id,
                        "type": TaskDatabaseType.DATAFRAME,
                    }
                ]
            ],
...
```

vantage6-server/vantage6/server/resource/dataframe.py
```diff
@@ -573,7 +574,12 @@ def post(self, id):
         response, status_code = self.create_session_task(
             session=session,
             databases=[
-                {"dataframe_id": dataframe.id, "type": TaskDatabaseType.DATAFRAME}
+                [
+                    {
+                        "dataframe_id": dataframe.id,
+                        "type": TaskDatabaseType.DATAFRAME,
+                    }
+                ]
```
frankcorneliusmartin, line 582: https://github.com/vantage6/vantage6/pull/2065/files/62fc2b8902b183d6ca879de37b1e5ae076128b2e#r2170924293

vantage6-server/vantage6/server/resource/token.py
```diff
@@ -154,6 +154,26 @@ def post(self):
 
         # We store the task metadata in the token, so the server can verify later on
         # that the container is allowed to access certain server resources.
+
+        # TODO: FM this is not super nice generated AI code.
```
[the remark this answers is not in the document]
frankcorneliusmartin: I looked at the code and decided to remove the TODO: ce386baee960bfea409213bf8213631640c2f22d

vantage6-ui/src/app/components/forms/compute-form/create-analysis-form.component.html
frankcorneliusmartin: Yes you are right, I considered for a second that you have algorithms that support wild things like:

```python
@dataframe("multiple", 2)
def f(dfs, df, df):
  pass
```

But I also saw it would be a lot of effort to implement this in the UI, and probably its never(?) going to be used, so decided it was not worth it. So the two types that are supported in the UI are:

* `@dataframes` (or the same `@dataframe("multiple")`
* `@dataframe(2)` (or any other positive int)

It that acceptable for you?

vantage6-ui/src/app/components/forms/compute-form/create-analysis-form.component.ts
```diff
@@ -777,10 +792,25 @@ export class CreateAnalysisFormComponent implements OnInit, OnDestroy, AfterView
...
+        } else {
+          this.columns = [];
+        }
+      } else if (dataframeID) {
+        // For single select
+        const dataframe = this.dataframes.find((_) => _.id === Number(dataframeID));
+        if (dataframe) {
+          this.setColumns(dataframe);
+        } else {
+          this.columns = [];
```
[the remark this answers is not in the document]
frankcorneliusmartin: e8537eab0fb1b65185603d15bce487597d5d4d0e

vantage6-ui/src/assets/localizations/en.json
```diff
@@ -240,7 +240,8 @@
       "title": "Define parameters",
       "description": "Provide parameter values for the selected algorithm function",
       "valid-json": "Please provide valid JSON input.",
-      "upload-or-fill-in-json": "Upload a JSON file or fill in the JSON input field"
+      "upload-or-fill-in-json": "Upload a JSON file or fill in the JSON input field",
+      "warning-column-list-many-dataframes": "This function accepts multiple dataframes in a single argument. The selection of columns are from the first selected dataframe."
```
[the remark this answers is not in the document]
frankcorneliusmartin: https://github.com/vantage6/vantage6/issues/2076

vantage6-server/vantage6/server/resource/dataframe.py
bartvanb: That code change was exactly what I meant, but I didn't see that that was the preprocessing endpoint during review. Thanks!

vantage6-ui/src/app/components/forms/compute-form/create-analysis-form.component.html
bartvanb: I mentioned this point in [#1863](https://github.com/vantage6/vantage6/issues/1863)

Issue #2055, opened 2025-06-18
[Feature Request] External KeyCloak provider


frankcorneliusmartin:
In the IDEA4RC project (and maybe others in the future) vantage6 is used as part of a bigger architecture including its own instance of keycloak. In such instance it would be useful to allow a non internal keycloak instance. 

This could also be of interest when using service from Azure or AWS that provide a good configure keycloak out of the box.

Requirements (specifically for the IDEA4RC project):

- [ ]  Nodes (and users?) need to be able to login using *direct access grant**
- [ ] KeyCloak paths needs to be settable in node and server
- [ ] Custom claim needs to be renamed to `vantage6_client_type`
- [ ] Block the creation and deletion of users and nodes as we do not want to modify the existing KeyCloak
- [ ] vantage6 users needs to be able to couple to the external auth module (this also goes for the root user)

bartvanb:
Everything required is now implemented and it's working with an Azure server running keycloak :-)

PR #2066, opened 2025-06-24
Added options to user create+delete that the user is not created/dele…

bartvanb:
…ted in vantage6-managed keycloak but instead is left untouched

Partial fix of #2055


vantage6-server/vantage6/server/resource/user.py
```diff
@@ -491,18 +506,28 @@ def post(self):
...
+                    }
                 ],
             }
         )
+
+    def _verify_user_in_keycloak(self, data):
+        keycloak_admin: KeycloakAdmin = getKeyCloakAdminClient()
+        keycloak_id = keycloak_admin.get_user_id(data["username"])
+        if keycloak_id is None:
+            raise BadRequestError("User does not exist in Keycloak")
```
frankcorneliusmartin, line 530: You need to update the docstring to include this.
```
400:
            description: Username or email already exists
```

vantage6-server/vantage6/server/resource/user.py
```diff
@@ -870,19 +909,27 @@ def delete(self, id):
...
+                    id,
                 )
                 for task in user.created_tasks:
                     task.delete()
 
-        self._delete_user_in_keycloak(user)
+        if params.get("delete_from_keycloak", True):
+            self._delete_user_in_keycloak(user)
+        else:
+            log.warning("User id=%s will not be deleted from Keycloak", id)
```
frankcorneliusmartin, line 923: I think this level is too high, they where already quite explicit when setting the flag `delete_from_keycloak`.
```suggestion
            log.info("User id=%s will not be deleted from Keycloak", id)
```

frankcorneliusmartin (approved):
* [x] Some minor comments
* [x] We need to create an option that the User KeyCloak management should be handled by the server (as a config option) and not by client arguments.

vantage6-server/vantage6/server/resource/user.py
frankcorneliusmartin: 27ae10a

Issue #2052, opened 2025-06-18
[Feature Request] Clearly show credentials generated in populate server


bartvanb:
**Problem description**
It is not clear to developers in the v5 dev env which credentials are created

**Desired solution**
Header with all information of what will be created in populate_server process when asking the question y/n

**Additional context**
Discussed here: https://github.com/vantage6/vantage6/pull/1982#discussion_r2154080612

PR #2067, opened 2025-06-24
Refactor development scripts and update devspace.yaml configuration. …

frankcorneliusmartin:
I addressed some small issues in our devspace environment.
* Devspace starts once, even when you populate it with data
* You only get prompted once to login (so no window mayhem)
* Improved logging so you get a report of the details what happened during population of the server https://github.com/vantage6/vantage6/issues/2052

*Sorry the refactor made it unclear that I moved a lot of code, and did not really introduce a lot of new stuff*


bartvanb:
Please fix codacy issues

PR #2077, opened 2025-06-27
Made minimal requirements to make the `MockClient` work with the `@da…

frankcorneliusmartin:
I only did the bare minimum, consider this a quickfix for making the `MockClient` work with the `@dataframes` decorator. Should be merged after #2065

### DIFF
--- a/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
+++ b/vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
@@ -106,9 +106,7 @@ class MockAlgorithmClient:
             org_data = []
             for dataset in org_datasets:
                 db = dataset.get("database")
-                if isinstance(db, pd.DataFrame):
-                    df = db
-                elif isinstance(db, dict):
+                if isinstance(db, pd.DataFrame) or isinstance(db, dict):
                     df = db
                 else:
                     df = load_data(


vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
```diff
@@ -17,8 +17,8 @@
 
 class MockAlgorithmClient:
     """
-    The MockAlgorithmClient mimics the behaviour of the AlgorithmClient. It
-    can be used to mock the behaviour of the AlgorithmClient and its
+    The MockAlgorithmClient mimics the behavior of the AlgorithmClient. It
+    can be used to mock the behavior of the AlgorithmClient and its
```
bartvanb, line 21: You like US english better than British?

vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
```diff
@@ -107,6 +108,8 @@ def __init__(
                 db = dataset.get("database")
                 if isinstance(db, pd.DataFrame):
                     df = db
+                elif isinstance(db, dict):
+                    df = db
```
bartvanb, line 112: Really nitpicking but can we not do `if isinstance(db. dataframe) or isinstance(df, dict)`? Saves scrolling

bartvanb (approved):
Why is this not simply included in the other PR? The other PR now leaves the mock client in a broken state so it would make sense just to add it there

vantage6-algorithm-tools/vantage6/algorithm/tools/mock_client.py
frankcorneliusmartin: Actually I don't. d8e984bd3abc125b3f68e7073ccc98f3c02837ae. I also changed my spellcheck in the editor to GB.

PR #2088, opened 2025-06-30
Add JWT decode leeway configuration option in ServerApp

frankcorneliusmartin:
* Tested it with a custom build in the IDEA4RC project
* I did not disable the IAT check as this was not easily configurable in the Flask-JWT-extended function interfaces and this seems like a proper fix.


bartvanb (approved):
Could you please add this comment (or something of along those lines) to make sure that we still understand why in a few years?

"Leeway is provided for the token IAT to prevent errors that token is not yet valid, which can happen if server times are drifting slightly. Note that this may be removed after updating to pyjwt > 3.0, where the validation behaviour changes - but be sure to check that."

PR #2111, opened 2025-07-09
add auth chart to release.yml and Makefile

CunliangGeng:


Makefile
```diff
@@ -183,11 +183,27 @@ ui-image:
 		$(if ${_condition_push},--push .,.)
 
 helm-charts:
+# Update Helm dependencies
```
bartvanb, line 186: could you please indent the comments properly

Makefile
```diff
@@ -183,11 +183,27 @@ ui-image:
...
 	helm package charts/common -d charts/
 	helm package charts/node -d charts/
 	helm package charts/store -d charts/
 	helm package charts/server -d charts/
+# Remove chart dependencies for dev purposes
+	rm -rf charts/auth/charts
+	rm -rf charts/common/charts
+	rm -rf charts/node/charts
+	rm -rf charts/store/charts
+	rm -rf charts/server/charts
```
bartvanb, line 203: If I understand correctly: if you're doing this locally, the deps will be in these directories and this is cleanup?

Also, for loop might be nice here since we're doing 5 times 3 identical actions (it would actually be nice in more places - maybe do it here immediately?)

Makefile
bartvanb: There is a setting in vscode (and derivatives) to replace all tabs automatically by spaces, which is very convenient for these cases :-)

PR #2114, opened 2025-07-09
Feature/#2069 build single image devspace

bartvanb:


dev/get_rebuild_command.sh
```diff
@@ -0,0 +1,29 @@
...
+    echo "Please provide only one argument." >&2
+    echo "Usage: ./dev/get_rebuild_command.sh [--server|--node|--store|--ui]" >&2
+    exit 1
+fi
+
+# check if arg is equal to server, node, store, or ui
+image_to_build=$1
+if [[ "$image_to_build" == "--server" || "$image_to_build" == "--node" || "$image_to_build" == "--store" || "$image_to_build" == "--ui" ]]; then
+    # remove the -- from the arg
+    image_name=$(echo "$image_to_build" | sed 's/^--//')
```
CunliangGeng, line 22:
```suggestion
    image_name=${image_to_build#--}
```

CunliangGeng: see bash "Remove prefix" in the cheatsheet https://devhints.io/bash

dev/get_rebuild_command.sh
```diff
@@ -0,0 +1,29 @@
...
+    exit 0
+elif [ $# -gt 1 ]; then
+    echo "Please provide only one argument." >&2
+    echo "Usage: ./dev/get_rebuild_command.sh [--server|--node|--store|--ui]" >&2
+    exit 1
+fi
+
+# check if arg is equal to server, node, store, or ui
+image_to_build=$1
+if [[ "$image_to_build" == "--server" || "$image_to_build" == "--node" || "$image_to_build" == "--store" || "$image_to_build" == "--ui" ]]; then
```
CunliangGeng, line 20: It's ok to use `if-else` here. Might be more clean if using `case`, see "Case/switch" in https://devhints.io/bash

PR #2134, opened 2025-07-15
WIP setup of setting password policy in keycloak

bartvanb:
I have created something to ensure that Keycloak only accepts passwords that are sufficiently complicated (>=8 characters, upper/lower case etc). This is done now by starting a sidecar container on the Keycloak pod that waits until keycloak is initialized and then sets the password policy by calling the Keycloak admin API. I did it this way because my preferred way of setting this up in the realmconfiguration to be automatically configured did not work - and no clear error message was provided.

I am not yet happy with this implementation, for the following reasons:

- I don't like the sh script directly in values.yaml - this should be a separate script. 
- The script should get the admin credentials as arguments - now they are hardcoded admin/admin. Ideally, we don't need to provide these credentials in multiple places in the values.yaml

I had difficulties putting the script in a proper volume, as is explained in the helm chart docs at https://artifacthub.io/packages/helm/bitnami/keycloak. Of course, we could build a separate docker image to configure the keycloak but that seems a bit overkill.


CunliangGeng:
> I did it this way because my preferred way of setting this up in the realmconfiguration to be automatically configured did not work - and no clear error message was provided.

I had a look at the error, the cause is that the passwords for realm admin must be updated to follow the password policy. I pushed two commits to remove the sidecar approach and directly use the configuration. It works well, as you can see from the keycloak console:

image


**👉 I did not check if the UI part works well or not for both admin and users, you could continue this work to verify it. V6 users' passwords might also need to be updated.**

PR #2138, opened 2025-07-15
add ruff config file

CunliangGeng:


.ruff.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[lint]
+extend-select = ["I"]     # Enable linting for imports
+
+[lint.isort]
+known-first-party = ["vantage6"]
+known-third-party = ["docker"]
+force-single-line = false
+combine-as-imports = true
+section-order = ["future", "standard-library", "third-party", "v6-common", "first-party", "local-folder"]
```
bartvanb, line 12: Sorry for not being more exhaustive before, but I think this can be extended such that the import order is:

```
vantage6.common
vantage6.client
vantage6.algorithm.tools
vantage6.cli
vantage6.backend.common
vantage6.node / store / server    <--- these three don't matter as they should never occur in the same file so choose whatever you prefer
```

That way we have even more structure :-)

Note that more structures would be possible since not all these combinations will occur (e.g. backend-common imports only occur in server and store so never conflicts with CLI), but it's good just to pick one order for consistency I guess

bartvanb (changes requested):
One more comment: can you add `ruff` in vantage6/common/setup.py under `extras_require["dev"]`? That way it will be auto-installed when some one does `make install-dev`

.ruff.toml
CunliangGeng: Updated in the [8431908](https://github.com/vantage6/vantage6/pull/2138/commits/8431908fe31920339dad26678314859a9f9f52b7) and [3cc80cf](https://github.com/vantage6/vantage6/pull/2138/commits/3cc80cf71b0053737a46cb383a9aa0a333ce2c3b).

If we do not force the structure, they will be in alphabetical order.

bartvanb (approved):
Perfect :-)

PR #2139, opened 2025-07-16
Fixed small typo (classic tutorial IS outdated)

dsmits:


docs/algorithms/index.rst
```diff
@@ -12,7 +12,7 @@ in the :ref:`classic-algo-tutorial`, an FL algorithm is build from scratch.
 
 .. warning::
 
-    Note that the classic tutorial is not outdated, and the commands may be
+    Note that the classic tutorial not outdated, and the commands may be
```
bartvanb, line 15: I think the word *not* should be removed right? Now the word *is* is removed

bartvanb (changes requested):

docs/algorithms/index.rst
dsmits: So apparently I opened up a PR that actually makes the text worse, not better 😅 .

Fixed it now

bartvanb (approved):

PR #2272, opened 2025-09-17
Change/format ruff instead of black

### LINKED ISSUE #2249: [Change Request] Adapt code style to ruff
**Description**
We currently have `black` but ruff is 99.9% compatible but just more extensive in its functionality. Applying ruff settings should not change as many lines as when we moved to black.

On v5 branch, we now have a ruff.toml which formats to ruff style, so we should definitively move to that.

- [ ] Adapt pipeline to check for ruff formatting instead of black.
- [ ] Run it on the code once to comply everywhere.

bartvanb:
Fix #2249


bartvanb:
Now it's done :-)

PR #2286, opened 2025-09-25
In dev environment cleanup, ensure that any running vantage6 tasks ar…

bartvanb:
…e also cleaned up

This is achieved by running an additional python script. This uses refactored code from the k8s manager to cleanup jobs

### DIFF
--- a/dev/cleanup_task_jobs.py
+++ b/dev/cleanup_task_jobs.py
@@ -86,6 +86,7 @@ def main() -> int:
         print(f"No Vantage6 task jobs found to delete in namespace '{namespace}'")
     else:
         print(f"Deleted {deletions} Vantage6 task job(s) in namespace '{namespace}'")
+    return 0
 
 
 if __name__ == "__main__":


dev/cleanup_task_jobs.py
```diff
@@ -0,0 +1,92 @@
...
+
+from vantage6.common.globals import APPNAME
+
+from vantage6.node.k8s.task_cleanup import delete_job_related_pods
+
+
+def _is_vantage6_task_job(job: k8s_client.V1Job) -> bool:
+    # Vantage6 task jobs can be identified by their name, which is of the form
+    # "vantage6-run-<run_id>"
+    return job.metadata.name.startswith(f"{APPNAME}-run-")
```
frankcorneliusmartin, line 19: This is ok, however adding a separate annotation to the job container would be preferred.

dev/cleanup_task_jobs.py
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        return int(annotations.get("run_id"))
+    except ValueError:
+        print(f"Job '{job.metadata.name}' has no run_id annotation, skipping")
+        return None
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description="Cleanup Vantage6 task jobs")
+    parser.add_argument("--namespace", type=str, default=None, help="Task namespace")
```
frankcorneliusmartin, line 43: I doubt that `None` is the correct default? Better would be `default` or simply exit when not provided (in that case best to make it an non optional argument)

dev/cleanup_task_jobs.py
```diff
@@ -0,0 +1,92 @@
...
+    # Vantage6 task jobs can be identified by their name, which is of the form
+    # "vantage6-run-<run_id>"
+    return job.metadata.name.startswith(f"{APPNAME}-run-")
+
+
+def _get_jobs(
+    namespace: str, batch_api: k8s_client.BatchV1Api
+) -> list[k8s_client.V1Job]:
+    try:
+        return batch_api.list_namespaced_job(namespace=namespace).items
```
frankcorneliusmartin, line 26: In case you decide to add a annotation/label, you can filter here immediately: https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/BatchV1Api.md#list_namespaced_job using the `label_selector`

dev/cleanup_task_jobs.py
```diff
@@ -0,0 +1,92 @@
...
+            core_api=core_api,
+            batch_api=batch_api,
+        )
+        deletions += 1
+
+    if deletions == 0:
+        print(f"No Vantage6 task jobs found to delete in namespace '{namespace}'")
+    else:
+        print(f"Deleted {deletions} Vantage6 task job(s) in namespace '{namespace}'")
+
```
frankcorneliusmartin, line 89: I think it would be good to explicitly return a `0` here.

frankcorneliusmartin (approved):
Minor comments, do with them what you want.

PR #2287, opened 2025-09-25
Change/#1510 and reinstate the dev network commands

bartvanb:


bartvanb:
This is ready for review now. There are a couple of things that I'd like to improve on still - some have to be documented alternatively.

This is my list of notes:

- [ ] For Linux users (or others not using Docker desktop), k8s_node_name is not `docker_desktop` as for others. They now have an option in `v6 sandbox new` to set this (`--k8s-node-name`) but it's far from trivial. Minimal solution is clear documentation, ideal solution is to derive the k8s node name in the command itself (maybe ask for user confirmation depending on how certain we are)
  - [ ] I think we need this option also for other new commands! I don't think e.g. v6 server new will produce usable config files otherwise
  - [ ] Potential solution is maybe `v6 use k8s-node` or so?
- [ ] Port forwarding is crashing regularly. It would be good to have a script that checks these processes and restarts port forward if it crashes
- [ ] The `additional-config` flags to provide additional configuration in the sandbox will I think not work, as it will create conflicts in the values.yamls. Haven't tested this, but it is bad practice anyway. We should improve this by prefering the additional config over the regular config (but not keeping both)
- [ ] Refactoring to do: resources (aka fixtures in this case) are created in three different ways for `v6 server import`, `v6 dev start` and `v6 sandbox new`. I partially refactored this already to sync the latter two processes a bit, but this can still be much cleaner. IMHO the populate scripts should be used everywhere. Now the import command has it's own stuff which is more or less duplicate from elsewhere.
- [ ] `v6 stop` commands don't clean up pv/pvs/secrets. They should. The dev pipeline already has a solution for this that should be refactored to work for both.
- [ ] `v6 node stop` does NOT stop the port forwarding process as of yet.

Issue #2213, opened 2025-08-06
[Change Request] Last things to do before releasing v5.0


bartvanb:
**Description**
This issue is meant to gather specific, more extensive tests that we need to do before releasing v5.0, that for some reason or other, we cannot do yet.

Things that might be done at any time point:
- [x] Check that new algorithm template creator is fully correct
- [x] Check that the `@metadata` decorator works: #1904 
- [ ] Write release notes

Last things:
- [ ] Check that CSP headers in the UI nginx configuration are set properly. 
  - We cannot test this now because we are using the dev Docker image for the UI in the devspace environment
  - We can test it once CLI commands to run UI with proper image are in place
- [ ] Release new tag on the algorithm template creator repo
  - [ ] Ensure doing the changes listed here: https://github.com/vantage6/v6-algorithm-template/issues/13
- [ ] Setting up community services
- [ ] Resolve remaining v5+ TODOs
  - [ ] In the `helm_install()` function, remove the `--devel` argument that currently is needed to allow using prerelease charts
  - [ ] remove `--prerelease=allow` note in quickstart section of the docs - it is necessary only while no final release is present
- [ ] Finalize release notes with security fixes

Asap after releasing 5.0 (some may also be done before):
- [ ] Deploying community services to latest version
- [ ] Release one v5 algorithm that contains the default extraction methods
  - [ ] Should the other images also contain them?
- [ ] Migrating community algorithms to v5
  - [ ] Release the v5 version of v6 diagnosistics as the `latest` image: merge https://github.com/vantage6/v6-diagnostics/pull/8 and then do `make publish`
  - [ ] After releasing average algorithm, validate that Python client docs are functional: https://docs.vantage6.ai/en/main/user/pyclient.html
- [ ] Making algorithms available in v5 community store
- [ ] Release the security advisories (see #1932)

PR #2298, opened 2025-09-30
Reinstated the `@metadata` decorator using environment variables inst…

frankcorneliusmartin:
…ead of JWT token extraction


bartvanb (changes requested):
OK fine by me. I think it would be important to test that this really works, maybe we can add that to #2213? And can you also add that the docs need to be written to one of the docs issues (or create a new one)? Otherwise we would probably forget

frankcorneliusmartin:
No worries. I added a paragraph of documentation about this decorator and added to the #2213 that we need to verify that the decorator actually works. Let me know if we can merge it like this.

frankcorneliusmartin:
I now also tested it using the (new) mock client. Ill leave the remark in #2213 to check it in a real network.

PR #2302, opened 2025-10-02
Bugfix/#2300 temporary fix for keycloak images

### LINKED ISSUE #2300: [Bug] Keycloak bitnami chart no longer supported
**Describe the bug**
Bitnami container images have been (re)moved from docker hub, see: https://github.com/bitnami/containers/issues/83267. This breaks the dev environment and deployment.

We should update to the best practice, however we can also choose to go with the legacy repository for the sort term.

frankcorneliusmartin:
Do not close #2300 as this is a temporary fix.


bartvanb (commented):
hate that this is necessary but good that it is fixed temporarily

PR #2318, opened 2025-10-16
Complete v6 server import

bartvanb:
- Fix todo's in v6 server import
  - better checks on input file
  - other minor stuff
- fix delete_dependents for collaboration
- change image pull policy to always pull (it was not using new versions of dev images but also in future it is important to use latest node/server/... image when starting the chart)

### DIFF
--- a/vantage6/vantage6/cli/server/import_.py
+++ b/vantage6/vantage6/cli/server/import_.py
@@ -1,6 +1,7 @@
 import click
 import requests
 import yaml
+from marshmallow import Schema, ValidationError, fields, post_load
 
 from vantage6.common import error, info
 from vantage6.common.globals import (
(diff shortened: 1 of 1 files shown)


vantage6-server/vantage6/server/resource/collaboration.py
```diff
@@ -664,6 +665,8 @@ def delete(self, id):
                     node.delete()
                 for study in collaboration.studies:
                     study.delete()
+                for session in collaboration.sessions:
+                    session.delete()
```
frankcorneliusmartin: I think we need the dataframes in this case too?

bartvanb: The `session.delete()` function already deletes the dataframes

frankcorneliusmartin (approved):

PR #2325, opened 2025-10-21
Improve exception logging

bartvanb:


vantage6-backend-common/vantage6/backend/common/resource/error_handling.py
```diff
@@ -86,6 +86,7 @@ def wrapper(*args, **kwargs):
             ForbiddenError,
             InternalServerError,
         ) as e:
+            log.exception(e)
```
frankcorneliusmartin: There is good reason why we do not log the exceptions for (most) of these exceptions, it would clutter the log file of the server when bots (or weird users) are touching our API. So I would suggest making this a little less verbose.

These exceptions can all be triggered by users: `UnauthorizedError`, `NotFoundError`, `BadRequestError` and `ForbiddenError`. I would suggest muting these completely. 

I think logging the `InternalServerError` makes sense as we don't know what is going on. So a simple suggestion would be to move the `InternalServerError` to the other except:

```python
except (Exception, InternalServerError) as e:
    log.exception(e)
```

Any particular reason you needed more verbose exception handling? Alternatively we can make an option as en env var so that these errors are logged.

frankcorneliusmartin (changes requested):

PR #2340, opened 2025-10-23
Fix node restart

bartvanb:
Also added temp sleep to make it work, but that should be resolved in other issue


bartvanb:
> > Also added temp sleep to make it work, but that should be resolved in other issue
> 
> I guess you mean the `--timeout` and `--wait` flags here? Because indeed this will return a non zero exit code if the resources have not stopped within this timeframe.

Nope I first had an ugly temporary `sleep` solution which the second commit fixed. So that comment was no longer valid

PR #2358, opened 2025-10-27
Bugfix/#2233 prepro df not ready

bartvanb:


vantage6-server/vantage6/server/resource/task.py
```diff
@@ -747,6 +758,32 @@ def _delete_subtasks(task: db.Task) -> None:
...
 
+    def _handle_last_session_task(self, task: db.Task) -> None:
+        """
+        Handle the last session task for a dataframe.
+        """
+        if task.action == AlgorithmStepType.DATA_EXTRACTION:
+            raise ForbiddenError(
+                "You cannot delete the data extraction task for a dataframe. Delete "
+                "the dataframe instead."
+            )
```
frankcorneliusmartin: This should also be covered by `task.dataframe.ready()`? As if it isnt ready, there is no data frame to delete?

PR #2385, opened 2025-11-03
Check that job pods are available before selecting them from a list

### LINKED ISSUE #1967: [Feature Request] Check for `FAILED` status when waiting for the POD to be running
**Problem description**
The moment you start watching it might have already failed. This we need to check for statuses that are > running.

## Original TODO in the code
https://github.com/vantage6/vantage6/blob/9ec91291deb9d6669ef397c77619d377f3125804/vantage6-node/vantage6/node/k8s/container_manager.py#L611-L616

bartvanb:
Fix #1967 - note that the original problem from that issue is no longer present in the code, however, when I checked the current code, I saw something that potentially could fail so fixed that

### DIFF
--- a/vantage6-node/vantage6/node/k8s/container_manager.py
+++ b/vantage6-node/vantage6/node/k8s/container_manager.py
@@ -581,6 +581,7 @@ class ContainerManager:
         # policy
 
         status = self.__wait_until_pod_running(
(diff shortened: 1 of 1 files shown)


frankcorneliusmartin (approved):
Nice catch!

PR #2388, opened 2025-11-03
`on_ping` event now also emits the sync event and the restores the ro…

frankcorneliusmartin:
…oms that the node is in.

- [ ] test it (my laptop is burning up from building idea images, so I prosponed this for now)

### DIFF
--- a/vantage6-server/vantage6/server/websockets.py
+++ b/vantage6-server/vantage6/server/websockets.py
@@ -395,6 +395,10 @@ class DefaultSocketNamespace(Namespace):
         """
         auth = db.Authenticatable.get_by_keycloak_id(session.auth_id)
 
+        # There is a bug https://github.com/vantage6/vantage6/issues/2386 where the node
+        # disconnects on the server side but the node is still connected to the server.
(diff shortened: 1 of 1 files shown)


vantage6-server/vantage6/server/websockets.py
```diff
@@ -381,6 +394,15 @@ def on_ping(self) -> None:
         ping and sets them as online.
         """
         auth = db.Authenticatable.get_by_keycloak_id(session.auth_id)
+
+        if auth.status != AuthStatus.ONLINE.value:
```
bartvanb: I would like to include a comment here why this stuff has been added, because normally, I would never expect this condition to be true

bartvanb (commented):
Two minor questions. I do think it's quite important to test this properly before merging it, as it is an essential part of the code

frankcorneliusmartin:
> Two minor questions. I do think it's quite important to test this properly before merging it, as it is an essential part of the code

I agree, I need to collect some additional data from the IDEA4RC setup before we merge this. Let me first wrap up the recordings.

Issue #2345, opened 2025-10-24
Provide way of asking for/specifying k8s node name in CLI


bartvanb:
For Linux users (or others not using Docker desktop), k8s_node_name is not `docker_desktop` as for others. They now have an option in `v6 sandbox new` to set this (`--k8s-node-name`) but it's far from trivial. Minimal solution is clear documentation, ideal solution is to derive the k8s node name in the command itself (maybe ask for user confirmation depending on how certain we are)

-  I think we need this option also for other new commands! I don't think e.g. v6 server new will produce usable config files otherwise
- Potential solution is maybe v6 use k8s-node or so?

PR #2389, opened 2025-11-04
Feature/#2345 specify k8s node cli

bartvanb:
This turned out to be more of a refactoring job than anticipated. Now at least the CLI seems ready to store more kubernetes variables in a stored k8s config file.

I solved #2345 not by creating a new command that asks to provide the k8s node name from the options, but for now simply automatically selecting the one that is active, and only offer feedback when user has switched k8s node (which one they want to use)


frankcorneliusmartin (approved):
Neat, this made it much more clean

PR #2390, opened 2025-11-05
Change/#2346 option manage ports in k8s

bartvanb:
- Adding development option `forward_ports` in the server and algorithm store
- Use this option in the sandbox (not in devspace as that forwards its own ports
- The option uses k8s' `NodePort` resource. This option was chosen because ingress and loadbalancer are not included in k8s installations by default and we want to keep the sandbox option as easy as possible to run
- the option exposes ports 30760-30762 for server, store and UI. Note that we could not use 76xx because nodeport is only possible in `30000-32767` range
- This PR also includes some minor cleanup / bugfixes / docs updates that I stumbled upon (e.g. also renames `v6 algorithm generate-algorithm-json` to `v6 algorithm generate-store-json`)

### DIFF
--- a/docs/server/components/yaml/algorithm_store_config.yaml
+++ b/docs/server/components/yaml/algorithm_store_config.yaml
@@ -92,9 +92,9 @@ store:
   # the server that it communicates with locally
   dev:
     # Specify the URI to the host. This is used to generate the correct URIs to
-    # communicate with the server. On Windows and Mac, you can use the special
-    # hostname `host.docker.internal` to refer to the host machine. On Linux, you
-    # should normally use http://172.17.0.1.
+    # communicate with the server. On Windows and Mac, when using Docker Desktop, you
+    # can use the special hostname `host.docker.internal` to refer to the host machine.
+    # On Linux, you should normally use http://172.17.0.1.
     host_uri: http://host.docker.internal
 
     # disable review process - all submitted algorithms are automatically accepted, which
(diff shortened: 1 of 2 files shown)


charts/server/templates/server/server-backend-deployment.yml
```diff
@@ -27,14 +27,9 @@ spec:
           value: {{ .Values.server.keycloakUrl }}
         - name: KEYCLOAK_REALM
           value: {{ .Values.server.keycloakRealm }}
-{{- if .Values.server.dev }}
-  {{- if .Values.server.dev.store_address }}
+{{- if and .Values.server.dev .Values.server.dev.store_address }}
```
frankcorneliusmartin: That is one ugly way of defining an `and` condition 🔥.

vantage6/vantage6/cli/sandbox/config/core.py
```diff
@@ -298,7 +296,7 @@ def __auth_config_return_func(self, extra_config: dict) -> dict:
                 "production": False,
                 "no_password_update_required": True,
                 "redirectUris": [
-                    f"{HTTP_LOCALHOST}:7600",
+                    f"{HTTP_LOCALHOST}:30760",
```
frankcorneliusmartin: Shouldn't this be `Ports.SANDBOX_SERVER.value`

frankcorneliusmartin (approved):
Very little to do here. For me, can you confirm:

* `v6 dev` (using devspace) uses port ranges `76xx` - uses port forwaring
* `v6 node/server uses port ranges `76xx` - has no external bindings
* `v6 sandbox` uses port ranges `3076x` - uses nodeports

vantage6-server/configs/postgress.yaml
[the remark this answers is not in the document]
bartvanb: They do seem ancient, and I could not see them being used anywhere. No idea what they used to be for

PR #2401, opened 2025-11-11
Add uluru as v5 name to the repo code

bartvanb:


bartvanb:
> You probably checked this, but I want to confirm. The image is free to use and requires no attribution?

Yes, Wikimedia commons free to use

PR #2412, opened 2025-11-18
Prevent failure of central task when jwt secret key is not set by ens…

### LINKED ISSUE #2368: [Bug] `JWT_SECRET_KEY` needs to be set
**Describe the bug**
In our `release/5.0` server configuration the `jwt.secret` needs to be set. Else this will not end up as the `JWT_SECRET_KEY` in `config.yml`. Somehow the internal logic to generate such a key is broken.

**To Reproduce**
* Start the server without the `jwt.secret`
* Start a central compute task. 
* Look in the server logs

Key error at:
```bash
File "/vantage6/vantage6-server/vantage6/server/resource/token.py", line 192, in post
    {"sub": container}, self.config["jwt_secret_key"], algorithm="HS256" 
```

bartvanb:
…uring the key is present in self.config after server initialization

### DIFF
--- a/charts/server/templates/server/server-backend-configmap.yml
+++ b/charts/server/templates/server/server-backend-configmap.yml
@@ -8,10 +8,8 @@ data:
     algorithm_stores: []
     api_path: {{ .Values.server.apiPath }}
     description: {{ .Values.server.description }}
-    {{- if .Values.server.jwt }}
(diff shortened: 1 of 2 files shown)


frankcorneliusmartin (approved):
Learned something new today; `{}.setdefault(.)`. Probably would have saved a bunch of if statements in the code if I knew this one earlier.

PR #2422, opened 2025-11-26
Feature/#2037 whitelisting v5

frankcorneliusmartin:


charts/node/templates/job-network-policy.yml
```diff
@@ -100,4 +116,97 @@ spec:
...
+{{- end }}
+
+---
+
+{{- if
+    and
+        .Values.node.whitelist
+        (hasKey .Values.node.whitelist "federated_compute")
+        (hasKey .Values.node.whitelist.federated_compute "egress")
+}}
```
bartvanb: I like this structure that we only define the policies if they are different from the default

docs/node/node_config.yaml
```diff
@@ -148,6 +148,45 @@ node:
     # if internet connection is not available. Default value is true.
     require_algorithm_pull: true
 
+  # Whitelisting settings to specify external addresses algorithms are allowed to
+  # connect to. Only Kubernetes network policies are supported. See:
+  #   https://kubernetes.io/docs/concepts/services-networking/network-policies
```
bartvanb:
```suggestion
  # https://kubernetes.io/docs/concepts/services-networking/network-policies
```

charts/node/values.yaml
```diff
@@ -69,3 +69,10 @@ node:
 
   prometheus:
     enabled: false
+
+  whitelist:
+    data_extraction:
+    preprocessing:
+    federated_compute:
+    central_compute:
+    postprocessing:
```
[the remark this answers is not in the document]
frankcorneliusmartin: Yes this is valid syntax it basically nulls the values: https://www.educative.io/answers/how-to-represent-null-values-in-yaml

PR #2437, opened 2025-12-04
Bugfix/move away from bitnami keycloak

bartvanb:
~~WIP - working on running Keycloak without Bitnami chart~~

- Run keycloak with the keycloak operator
  - new CLI command to install it
  - ensure dev env works with it
  - ensure sandbox works with it
  - introduce some new wait stages (to replace this done automatically by the bitnami chart)

While working on this, I ran into several other things that I decided to fix, such as:
- updated uv.lock
- Fix bug with default values whitelisting not rendering well in chart
- ...

### DIFF
--- a/vantage6-backend-common/vantage6/backend/common/__init__.py
+++ b/vantage6-backend-common/vantage6/backend/common/__init__.py
@@ -153,7 +153,14 @@ class Vantage6App:
         self.app.config["JWT_DECODE_LEEWAY"] = self.ctx.config.get(
             "jwt_decode_leeway", 10
         )
-        self.app.config["JWT_PUBLIC_KEY"] = self._get_keycloak_public_key()
+        try:
+            self.app.config["JWT_PUBLIC_KEY"] = self._get_keycloak_public_key()
+        except Exception as e:
+            log.exception(e)
+            log.error(f"Failed to get keycloak public key: {e}")
+            log.error("This means that you cannot login as a user")
+            log.error("Exiting...")
+            exit(1)
         self.app.config.setdefault("JWT_TOKEN_LOCATION", ["headers"])
 
         # Mail settings


charts/auth/templates/_helpers.tpl
frankcorneliusmartin: Any good reason to use the helpers from the common template?

vantage6-backend-common/vantage6/backend/common/__init__.py
```diff
@@ -121,13 +122,21 @@ def validate_required_env_vars(self) -> None:
...
+                    f"{os.environ.get(RequiredServerEnvVars.KEYCLOAK_URL.value)}/realms"
+                    f"/{os.environ.get(RequiredServerEnvVars.KEYCLOAK_REALM.value)}",
+                    timeout=100,
+                )
+                key = response.json()["public_key"]
+                return f"-----BEGIN PUBLIC KEY-----\n{key}\n-----END PUBLIC KEY-----"
+            except KeyError as e:
+                log.error(f"Failed to get keycloak public key: {e}")
+                time.sleep(retry_delay)
+        raise Exception("Failed to get keycloak public key")
```
frankcorneliusmartin: Shouldn't we catch this in `_configure_flask_base` (or preferably higher), I traced it back and could not find any catch. This will then lead to a supper ugly exit with exception.

vantage6/vantage6/cli/configuration_create.py
```diff
@@ -77,20 +77,14 @@ def add_common_server_config(
...
-        "Where is your server database located on the host machine?",
-        default=f"{Path.cwd()}/dev/.db/db_pv_server",
-    ).unsafe_ask()
+    service_name = "hub" if instance_type == InstanceType.SERVER else "store"
 
-    config["database"]["k8sNodeName"] = q.text(
-        "What is the name of the k8s node where the databases are running?",
-        default="docker-desktop",
-    ).unsafe_ask()
+    config = add_database_config(config, instance_type)
```
frankcorneliusmartin: I am confused. Should this not be in the `not production` case? Now we first ask them for the mount points and then we can also add an external database (in the `production` case). I think it is always one of them, not both?

charts/auth/templates/database/auth-database-pvc.yml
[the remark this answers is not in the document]
bartvanb: 0336b8da2 - it was actually not superfluous but poorly named: this file contained the PVC and the other the PV. (can I blame AI?)

charts/auth/values.yaml
```diff
@@ -1,149 +1,94 @@
...
-  # the resources.
+
+  production: true
+  tls:
+    enabled: true
+    autoGenerated: true
+
+  dev:
+    forward_ports: false
+    local_auth_port_to_expose: 30764
```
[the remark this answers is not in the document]
bartvanb: Nice double negative :-)
Yes it has to be here. A similar section is present for all charts except the node - it is used to determine whether or not to publish a `NodePort` (which we use in the sandbox)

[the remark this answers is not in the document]
bartvanb: Ah right, now I get what you mean. No they shouldn't be - I deleted it 43508d344

vantage6-backend-common/vantage6/backend/common/__init__.py
frankcorneliusmartin: I am not suggesting continueing after catching it at a higher level, merely that we get a better error message (instead of the exception log) and then exit.

PR #2458, opened 2025-12-22
SMTP server v5 in keycloak

### LINKED ISSUE #2007: [Bug] Email service not functional in initial Keycloak setup
**Describe the bug**
We have not set up emails being sent from keycloak to e.g. inform users that their account is temporarily blocked. 

**Expected behavior**
- [ ] Set up email sending via keycloak
- [ ] Check if we can delete all the SMTP service code from server/store

### LINKED ISSUE #2008: [Feature Request] Reinstate account recover options from before
**Problem description**
Initial implementation of keycloak does not include a lot of features that we used to support

**Desired solution**
Reinstate this functionality:
- [ ] Change password
- [ ] maximum number of password attempts, then block for X time
- [ ] Reset password and MFA
- [ ] Setting how long a token is valid

bartvanb:
~~Merge #2454 before reviewing this~~

- Fix #2007 
  - Add SMTP email service to keycloak 
  - Reinstate algorithm store emails
- Fix #2008 
  - 'forgot password' functionality
  - brute force protection
- option to allow server admin to require email verification

What still needs to happen before #2007 is fixed is to clean up the server email code and to re-enable store emails

### DIFF
--- a/vantage6-algorithm-store/vantage6/algorithm/store/resource/review.py
+++ b/vantage6-algorithm-store/vantage6/algorithm/store/resource/review.py
@@ -393,6 +393,7 @@ class Reviews(ReviewBase):
             algorithm.save()
 
         # send email to the reviewer to notify them of the new review
+        # TODO next time we change something here, update to kwargs instead of args
         Thread(
             target=self._send_email_to_reviewer,
             args=(


charts/store/templates/server/store-configmap.yml
```diff
@@ -73,6 +73,28 @@ data:
...
     {{- end }}
     port: {{ .Values.store.internal.port }}
+    base_url: {{ .Values.store.baseUrl }}
     root_user:
       username: {{ .Values.store.keycloak.adminUsername }}
-      v6_hq_uri: {{ .Values.store.vantage6HQUri }}
\ No newline at end of file
+      v6_hq_uri: {{ .Values.store.vantage6HQUri }}
+    {{- if .Values.store.smtpServer }}
+    smtpServer:
```
frankcorneliusmartin: Can we make this `smtp_server` to be consistent?
```suggestion
    smtp_server:
```

(If you accept this change, it needs to be changed where its referenced too.. obviously.)

vantage6-hq/vantage6/hq/__init__.py
```diff
@@ -7,6 +7,7 @@
 
 import importlib.metadata
 import os
+from pathlib import Path
 
 from gevent import monkey
```
frankcorneliusmartin: Also here should the Monkey patch go first.

frankcorneliusmartin (changes requested):

charts/auth/values.yaml
```diff
@@ -27,13 +27,36 @@ keycloak:
     ssoSessionIdleTimeout: 1800
     passwordPolicy: "length(8) and upperCase(1) and lowerCase(1) and digits(1) and specialChars(1)"
     editUsernameAllowed: false
+    resetPasswordAllowed: false
```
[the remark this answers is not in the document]
bartvanb: This is disabled by default because there is no SMTP server by default. If no emails can be sent, this should be set to false to prevent that we have users clicking a reset button that will not work (I think)

charts/store/templates/server/store-configmap.yml
bartvanb: eded90a29 - note that I only changed the references in the configMap and *not* those in the values.yaml. I might also do that to be consistent, but in the keycloak chart is *has to be* `smtpServer` or it will go wrong there, and then we have two versions and things are bound to go wrong some day.

vantage6-algorithm-store/vantage6/algorithm/store/resource/review.py
```diff
@@ -399,6 +400,8 @@ def post(self):
                 self.mail,
                 review,
                 g.user.username,
+                algorithm.name,
+                algorithm.developer.username,
                 self.config,
                 request.headers.get("store_url"),
             ),
```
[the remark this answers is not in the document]
bartvanb: 187b88e44 - I appreciate the suggestion but I want to be sure that it works and that would cost me a non-negligible amount of testing time. So I put a TODO to improve this next time we touch this code.

PR #2490, opened 2026-02-11
Change/#2266 device authorization flow

bartvanb:
Fix the issue with localhost forwards in v5 by implementing device authorization flow. This will still try to open a URL in the browser when a user does `client.authenticate()` - but if that fails, it will print instructions on how to open a browser and submit a printed code, so that logging in will also work on a VM.

This also allows us to remove the previously required forwards to the localhost:7681 HTTP server spun up by the client, which has now been deleted.

### DIFF
--- a/vantage6/vantage6/cli/auth/install.py
+++ b/vantage6/vantage6/cli/auth/install.py
@@ -226,7 +226,7 @@ def _wait_for_operator_ready(
     start_time = time.time()
     deployment_name = "keycloak-operator"
 
-    while time.time() - start_time < timeout:
+    while (time.time() - start_time) < timeout:
         try:
             command = [
                 "kubectl",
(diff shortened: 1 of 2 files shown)


vantage6-client/vantage6/client/__init__.py
```diff
@@ -155,97 +154,201 @@ def _get_logger(level: str) -> logging.Logger:
...
+            return
 
-                    # Send success response
-                    self.send_response(200)
-                    self.send_header("Content-type", "text/html")
-                    self.end_headers()
-                    self.wfile.write(b"Login successful! You can close this window.")
+        # check if browser can be opened to directly authenticate, otherwise fallback to
+        # console output
+        self._maybe_open_browser_for_device_flow(device_data)
```
frankcorneliusmartin: I had a got laugh about this name. I think try sounds better, maybe makes it sound like it might happen. 
```suggestion
        self._try_open_browser_for_device_flow(device_data)
```

frankcorneliusmartin (approved):
I assumed this implemented: https://github.com/keycloak/keycloak-community/blob/main/design/oauth2-device-authorization-grant.md

PR #2491, opened 2026-02-12
Set organization id in keycloak

### LINKED ISSUE #1994: [Change Request] Set organization ID of user+node in Keycloak
**Description**
Keycloak tokens should provide server and store with the organization ID of the user. In v4, this is stored separately in the server and store and synced by server-store communication (the store needs org IDs to check if reviews are done from multiple organizations). In v5, server-store communication is largely eliminated as both just talk to Keycloak. 

- [x] Add user attribute organization ID in keycloak
- [x] Set it for the user and ~~node~~ upon creating the account - **I did not implement it for the node because this is not required at present**
- [ ] ~~Change it in PATCH endpoints~~ **not necessary - patch user doesn't allow updating organization_id for both server and store**
- [ ] Use it in store for algorithm review policies (currently disabled I think)
- [ ] Remove it from store database
- [ ] Remove it from server database and use keycloak-provided ID where necessary

An alternative way of fixing this is by having the store request this information from the server but that is not the nicest way to do this

bartvanb:
Fix #1994.

- Create user attribute `organization_id` in keycloak
- Set it when creating a user in HQ
- Obtain it in the algorithm store when creating a user there

There are a couple of things different from the original plan in the issue:
- I didn't set `organization_id` for the nodes in keycloak - there is no use case because only HQ needs to know this info. We can always do it later when useful
- I did not remove organization_id as field in the HQ and store databases. This was the original plan in the issue, but I think that is very inconvenient in HQ because it is used often in SQLAlchemy relations. In the store, it could be removed but that is added work for not much benefit - we can always do that later if necessary.

### DIFF
--- a/vantage6-algorithm-store/tests_store/resources/test_user.py
+++ b/vantage6-algorithm-store/tests_store/resources/test_user.py
@@ -77,7 +77,9 @@ class TestUserResource(TestResources):
         """Test the user create."""
         # Mock the keycloak admin client
         get_keycloak_id_for_user_mock.return_value = "test-keycloak-id"
-        get_organization_id_from_keycloak_mock.return_value = None
+        # Simulate that Keycloak user has an organization_id attribute so creation
+        # succeeds
+        get_organization_id_from_keycloak_mock.return_value = 1
 
         self.register_user(authenticate_mock=authenticate_mock)
 


vantage6-algorithm-store/vantage6/algorithm/store/resource/user.py
```diff
@@ -277,19 +280,13 @@ def post(self):
...
-        #         "msg": f"User '{data['username']}' not found in the Vantage6 server."
-        #     }, HTTPStatus.BAD_REQUEST
-        # user_org = server_response.json()["data"][0]["organization"]["id"]
+        # Retrieve organization_id from Keycloak user attributes
+        organization_id = get_organization_id_from_keycloak(user_id)
+        if organization_id is None:
+            log.warning(
+                "User '%s' does not have organization_id attribute in Keycloak",
+                data["username"],
+            )
```
frankcorneliusmartin: Are you sure you want to continue here instead of returning an error code?

Issue #2509, opened 2026-02-23
`k8sNodeName` should not be obligatory for the hub charts


bartvanb:
The `k8sNodeName` variable is used in the charts of HQ, store and auth. This variable designates the name of the kubernetes node. In production scenarios, however, this name is not always a constant as there may be several worker nodes. 

The reason it is there at present, is that is used for persistent volumes. For the SQL db's, #2494 ensures that the PV/PVCs are no longer required. However, the k8s node name is still required for logging (both HQ and store) and prometheus (HQ). This must no longer be the case.

_See comments by @frankcorneliusmartin in https://github.com/vantage6/vantage6/pull/2494#discussion_r2840818309_

bartvanb:
In #2509, there is a temporary workaround to simply not store some stuff that is otherwise stored in PVCs. Moving this to next minor, where we should figure out a way to store logs etc in a cloud environment

Issue #2452, opened 2025-12-12
[Change Request] Complete `v6 hub` functionality


bartvanb:
**Description**
Recently, a first version of `v6 hub new` was built to generate configs for all central component charts (i.e. all except the node). This however is aimed at generating production-ready charts and has not been tested yet. There are very likely things not working.

- [ ] Improve `v6 hub new` to ensure that it works
- [ ] Implement `v6 hub start` and `stop` - which should be very easy - simply a call to those start commands

bartvanb:
The idea in doing this now is to create a `hub` chart that deploys the subcharts auth, hq and store. That can then also function as a way to automatically set up ingress / load balancing.

- [x] Create hub chart
  - [x] Add subcharts
  - [x] Provide ingress / load balancing to let components communicate with one another
- [x] Test the hub chart in a cloud environment

bartvanb:
Fixed by #2530 and tested in AKS

PR #2521, opened 2026-03-04
Feature/#2452 hub chart

### LINKED ISSUE #1924: [Feature Request] Create a `values-prod.yml` for the production setup that we deliver with the helm chart
**Problem description**
Or make the `values.yml` by default the production one. Currently we are adding examples in the values.yml which might be confusing. We can also consider to comment out all optional things (I've seen other projects do the same @CunliangGeng what do you think about this solution)

bartvanb:
Introducing a hub chart that can run the auth, hq and store charts together. 

I am not satisfied yet with this, so I put this in draft. I'll notify when ready for next step

What this PR contains:
- hub chart - fix most of #2452 - we still need to check that everything works in test cloud deployement before closing that issue
  - introduce global values that are used in the subcharts, so that we don't have to define them in each subchart
  - added hub helm chart to release process
  - include wait processes in chart deployment so that store+HQ don't crash while auth initializes, and HQ doesn't crash while store is starting up
- simplified values.yaml for auth chart as some could be put more easily in the chart itself
- Fix #1924
- Update example yaml files in the docs with all these changes
- Temporary workaround for #2509 so that k8snodename does not need to be set, so that we can have production envs, but that also means that in that case log files are not stored and prometheus cannot be enabled for HQ.

TODO:
- [x] add hub config yaml to docs
- [x] cleanup `_kubectl` functions in the CLI
- [x] cleanup wait scripts in the helm charts
- [x] update v6 hub new to use global settings and not generate separate yamls
- [x] figure out if log files are still stored somewhere if using the hub chart
- [x] update devspace to use hub chart OR check that `v6 dev` still works and postpone it


charts/hub/values.yaml
```diff
@@ -0,0 +1,53 @@
+# Default values for the hub parent chart.
+#
+# The keys below correspond to the subchart names so that their values.yaml
+# structures can be provided from this single hub chart values file.
+hq:
+  enabled: true
+  # Copy or override values from charts/hq/values.yaml here as needed, e.g.:
```
frankcorneliusmartin:
```suggestion
  # Copy or override values from charts/hq/values.yaml here as needed.
```

charts/hub/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+auth:
+  enabled: true
+  # Copy or override values from charts/auth/values.yaml here as needed.
+
+store:
+  enabled: true
+  # Copy or override values from charts/store/values.yaml here as needed.
+
+global:
+  # Shared Keycloak configuration (used as default if subcharts don't override)
```
frankcorneliusmartin: Why did you choose to make the subcharts leading? I think it would make more sense if this Hub chart is always the one who overwrites?

vantage6/vantage6/cli/configuration_manager.py
```diff
@@ -250,6 +255,56 @@ def get_config_template(self) -> str:
         return super()._get_config_template(AUTH_TEMPLATE_FILE)
 
 
+class HubConfigurationManager(ConfigurationManager):
```
frankcorneliusmartin: The docstings of the methods in this class are not complete.

vantage6/vantage6/cli/hub/remove.py
```diff
@@ -0,0 +1,35 @@
...
+def cli_hub_remove(
+    ctx: HubContext, name: str, system_folders: bool, force: bool
+) -> None:
+    """
+    Function to remove a hub.
+
+    Parameters
+    ----------
+    ctx : HubContext
+        Hub context object
```
frankcorneliusmartin: `system_folders` is missing

vantage6/vantage6/cli/configuration_manager.py
```diff
@@ -250,6 +255,56 @@ def get_config_template(self) -> str:
...
+    """
+    Maintains the hub's configuration.
+    """
+
+    def __init__(self, name, is_sandbox: bool = False, *args, **kwargs):
+        super().__init__(conf_class=HubConfiguration, name=name, is_sandbox=is_sandbox)
+
+    @classmethod
+    def from_file(cls, path: Path | str, is_sandbox: bool = False) -> Self:
```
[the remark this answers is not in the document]
bartvanb: Agree that the future stuff looks way better. Anyway, let's leave this as is

PR #2569, opened 2026-03-24
Undo 4.14 so that we can release 5.0 into main

bartvanb:
Just wait to merge this until we have the v5 PR


bartvanb:
This should be closed since we now have more changes in main that we want to put into 5.0.

PR #2593, opened 2026-04-24
Release 4.15.0

bartvanb:
Main point of this release is to create images in the Github container registry now that harbor2 is offline. 

Before merging, check that the release date is correct


.github/workflows/release.yml
```diff
@@ -148,20 +148,22 @@ jobs:
       stage: ${{ needs.version.outputs.stage }}
       tag_latest: ${{ needs.version.outputs.tag_latest }}
     steps:
-      - name: Login to harbor2.vantage6.ai
-        env:
-          USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          PASSWORD: ${{ secrets.DOCKER_TOKEN }}
-        run: docker login harbor2.vantage6.ai -u $USERNAME -p $PASSWORD
+      - name: Log in to GHCR
+        # https://github.com/docker/login-action/releases/tag/v4.1.0
```
frankcorneliusmartin:
```suggestion
        # https://github.com/docker/login-action/releases/tag/v4.1.0
```
```suggestion
        # pinned to: https://github.com/docker/login-action/releases/tag/v4.1.0
```

docs/node/configure.rst
frankcorneliusmartin, line 1: The sha's here are dead. Which is not really an issue as these are more as examples, agreed?

docs/release_notes.rst
```diff
@@ -1,6 +1,26 @@
 Release notes
 =============
 
+4.15.0
+
+*30 April 2026*
```
frankcorneliusmartin: I guess this date will be correct if you merge it on Thursday, but this is a reminder if it isnt

tools/update-discord.py
frankcorneliusmartin, line 1: I geuss this is here because we still have the discord stuff in the release 4 branch? As we removed it from 5: https://github.com/vantage6/vantage6/pull/2559/files. If we keep maintaining 4 we also need to remove this.

frankcorneliusmartin (approved):
Shall we also do A PR for both 4.x and 5.x to run ruff over the entire codebase? This way we can avoid having these changes clutter the actual code changes.

.github/workflows/release.yml
bartvanb: added this to all instances of this command

docs/node/configure.rst
bartvanb: Yes just examples, I didn't assume original was real sha actually

tools/update-discord.py
bartvanb: Ideally yes. But let's do this only if we run into issues with this - I'd prefer to not invest the effort until Discord is offline + v5 is online. And then only when we see that it is necessary

bartvanb:
Just tested 4.15.0rc3 and that is working as intended. We could release 4.15 now, however note that the community store is still offline, so we'd need to make it clear that not everything is working

PR #2594, opened 2026-05-01
Merge/4.15 into 5.0

bartvanb:
Wait for merge - first verify this still runs in v6 dev. 
Before/After merge: run pipeline to verify that works again


bartvanb:
I'm merging this now because `v6 dev` is running OK. However, I haven't tested algorithms running because as long as we haven't rerun the pipeline to create v5 base algorithm images, we cannot build algorithm images via their github pipelines. So I'll first merge this and run another RC, then test again with v6 dev

PR #2604, opened 2026-05-12
Change/prevent doubles config list wsl

bartvanb:
`v6 hub/node/... list` commands in v5 showed duplicates if run from within WSL. Also, `v6 hub list` showed also nodes and vice versa. This PR addresses both those issues

### DIFF
--- a/vantage6-common/vantage6/common/context.py
+++ b/vantage6-common/vantage6/common/context.py
@@ -353,7 +353,7 @@ class AppContext(metaclass=Singleton):
             return {
                 "log": mount_path / "log",
                 "data": mount_path / "data",
-                "config": mount_path / "config",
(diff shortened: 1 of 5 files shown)


frankcorneliusmartin (approved):
The code looks good. However I am unsure why 

> for WSL, there is no distinction between user and system folders, so we
> only print the user folders

Is the case. Is that because our setup? Because obviously there are user and system folders in WSL.

bartvanb:
See https://github.com/vantage6/vantage6/blob/release/5.0/vantage6-common/vantage6/common/context.py#L348. So there are user/system folders in WSL, but we can't use them as we're using /mnt/wsl folders as that is the only place mounting stuff works

PR #2606, opened 2026-05-12
Improve looking for running services to include hub services when HQ/…

bartvanb:
…auth/store is requested. This fixes e.g. v6 hq attach

### DIFF
--- a/vantage6/vantage6/cli/common/utils.py
+++ b/vantage6/vantage6/cli/common/utils.py
@@ -120,16 +120,28 @@ def find_running_service_names(
         svc_ends_with = f"user-{instance_type.value}"
     else:
         svc_ends_with = f"-{instance_type.value}"
+    is_part_hub = False
+    if instance_type in [
+        InstanceType.HQ,
+        InstanceType.ALGORITHM_STORE,
+        InstanceType.AUTH,
+    ]:
+        is_part_hub = True
+        if only_system_folders:
+            alternative_svc_ends_with = "system-hub"
+        elif only_user_folders:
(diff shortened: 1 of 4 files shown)


vantage6/vantage6/cli/common/utils.py
```diff
@@ -120,16 +120,28 @@ def find_running_service_names(
...
         release_name = release.get("name", "")
 
         # Check if this is a vantage6 service
-        is_matching_service = (
-            release_name.startswith(svc_starts_with)
-            and release_name.endswith(svc_ends_with)
-            and instance_type.value in release_name
+        is_matching_service = release_name.startswith(svc_starts_with) and (
+            release_name.endswith(svc_ends_with)
+            or (is_part_hub and release_name.endswith(alternative_svc_ends_with))
```
frankcorneliusmartin, line 144: That is one hard to read condition. I dont know whether ruff accepts something like this, else maybe break it into multiple vars?
```suggestion
        is_matching_service = 
            (
                release_name.startswith(svc_starts_with) 
                and (
                    release_name.endswith(svc_ends_with)
                    or (
                        is_part_hub 
                        and release_name.endswith(alternative_svc_ends_with)
                    )
                )
            )       
```

frankcorneliusmartin (approved):

vantage6/vantage6/cli/common/utils.py
bartvanb: I copied your suggestion and ruff autoformats it back to the original ;-) so ignoring this comment even though I agree with the sentiment

PR #2641, opened 2026-06-25
Bugfix/prevent node runs without algorithm isolation

bartvanb:


vantage6/vantage6/cli/globals.py
```diff
@@ -46,7 +46,7 @@
 ALGORITHM_TEMPLATE_REPO = "gh:vantage6/v6-algorithm-template.git"
 
 # image to use for diagnostics in `v6 test` commands
-DIAGNOSTICS_IMAGE = "ghcr.io/vantage6/algorithm/diagnostic:uluru"
+DIAGNOSTICS_IMAGE = "ghcr.io/vantage6/algorithm/diagnostic:5"
```
frankcorneliusmartin, line 49: Why? Uluru is the preffered name or is this beter security wise?

vantage6-node/vantage6/node/k8s/network_isolation.py
frankcorneliusmartin, line 1: Isn't this a bit over engineered? There is a lot of code just to make sure that the `ISOLATION_PROBE_CANDIDATES` are not in the `whitelist` section, which is stupid as the domainnames in this list are truely example domains (I probably would be more worried that a proxy would block the traffic and the wrong conclusion is drawn that the isolation came from the network policies)

If we remove that check we also do not have to provide the `whitelist` in the config map consumed by this node. 

So to summarize, I think I would prefer:

* Not to check the probes against our whitelist, just to keep things simple (remove quite a bit of code and the config map gets cleaner)
* Probe all domains, as one might be blocked due to proxy reasons

charts/node/templates/node-configmap.yml
```diff
@@ -89,4 +90,8 @@ data:
       {{- range .Values.node.algorithm_env }}
       - {{ . }}
       {{- end }}
+    {{- end }}
+    {{- if .Values.node.whitelist }}
+    whitelist:
+{{ .Values.node.whitelist | toYaml | indent 6 }}
```
frankcorneliusmartin: Do I understand correctly that you supply these to the node so you can check that the isolation works that is created by k8s network policies? 

There is a bit of a risk that this option gives the idea that the isolation is handled by the pod.

frankcorneliusmartin (changes requested):

vantage6/vantage6/cli/globals.py
bartvanb: This stems for the work I did on migrating community algorithms from v4 to v5. I found that we used a mix of `4`/`cotopaxi` and `5`/`uluru` tags. Instead of a mix, we should use either the `4`/`5` tags or the named tags. I opted for using the version number tags for all community algorithms, because it is in line with how other packages do it (e.g. `ubuntu:22` / `python:3.13` / `alpine:13`)

charts/node/templates/node-configmap.yml
bartvanb: Yes correct. Now that I no longer check the whitelist, this is no longer necessary so I'll remove it

vantage6-node/vantage6/node/k8s/network_isolation.py
bartvanb: Fair comment. I guess I went down that rabbit hole because our example whitelists were actually using `example.com`. But I already changed those in the current version of the PR

charts/node/templates/node-configmap.yml
bartvanb: https://github.com/vantage6/vantage6/pull/2641/commits/a39a224643ca3effaabd11fdfd598cdf27707096

vantage6-node/vantage6/node/k8s/network_isolation.py
bartvanb: https://github.com/vantage6/vantage6/pull/2641/commits/a39a224643ca3effaabd11fdfd598cdf27707096

vantage6/vantage6/cli/globals.py
frankcorneliusmartin: Ok fair. ✅

frankcorneliusmartin (approved):
LGTM when linting erros are being resolved

frankcorneliusmartin:
I see now this is related to the ruff update which is handled in a separate PR #2682

PR #2660, opened 2026-07-15
Gaurd the local-storage class in the subcharts when a parent chart is…

frankcorneliusmartin:
… used

This was a problem when I used `v6 sandbox new`, I got the error:
```bash
Error: INSTALLATION FAILED: 2 errors occurred:
        * storageclasses.storage.k8s.io "local-storage" already exists
        * storageclasses.storage.k8s.io "local-storage" already exists
```
After which the deployment in k8s stalls.

> The hq, auth and store subcharts each need a "local-storage" StorageClass on
  docker-desktop for their PersistentVolumes. A StorageClass is a cluster-scoped
  resource, so it must be created exactly once. Because all three subcharts are
  rendered in a single hub release, their per-subchart lookup guard cannot see
  the others (nothing exists yet at render time), which caused every subchart to
  emit its own copy and Helm to fail with "StorageClass local-storage already
  exists". The subchart templates are therefore guarded with `.Chart.IsRoot` and
  the hub creates the single shared StorageClass here.


bartvanb (approved):

frankcorneliusmartin:
Mergen as the `ruff` errors are coming from a new ruff version, for which another PR is open #2682

PR #2662, opened 2026-07-16
Added `reason` argument to websocket `on_disconnect` handlers for nod…

### LINKED ISSUE #2655: [Bug] SocketIO signature mismatch
* `on_disconnect()` signature error — Missing reason parameter that newer versions of python-socketio pass
   Added reason: str = None parameter
* `on_connect()` signature error — Missing _auth parameter
   Added _auth: dict | None = None parameter

frankcorneliusmartin:
…e and server.

Closes #2655

### DIFF
--- a/vantage6-hq/vantage6/hq/websockets.py
+++ b/vantage6-hq/vantage6/hq/websockets.py
@@ -238,7 +238,11 @@ class DefaultSocketNamespace(Namespace):
             # delete any data on the node stored on HQ (e.g. configuration data)
             self.__clean_node_data(auth)
 
-        self.log.info(f"{session.name} disconnected (reason: {reason})")
+        self.log.info(
+            "%s disconnected (reason: %s)",
+            session.name,
+            reason if reason else "unknown"
+        )
 
         # cleanup (e.g. database session)
         self.__cleanup()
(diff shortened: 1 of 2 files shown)


vantage6-hq/vantage6/hq/websockets.py
```diff
@@ -232,7 +238,7 @@ def on_disconnect(self) -> None:
             # delete any data on the node stored on HQ (e.g. configuration data)
             self.__clean_node_data(auth)
 
-        self.log.info(f"{session.name} disconnected")
```
bartvanb, line 235: Maybe nice to add a check if reason is None, otherwise set text to 'unknown' or so?

Also, I thought we changed log styles everywhere to `%s` instead of `f""` - but not very important - it is a super super minor performance improvement. But then this could be

```suggestion
        self.log.info(
            "%s disconnected (reason: %s)", 
            session.name, 
            reason if reason else "unknown"
        )
```

bartvanb: Same below btw

bartvanb (approved):

vantage6-hq/vantage6/hq/websockets.py
frankcorneliusmartin: I am stubborn, I really like the f-strings over the formatting. But I agree, lets not mix styles.  94bd9caff8ee3cbf6a840ad97ee227eecbf08dbb

PR #2663, opened 2026-07-16
Fixed the `return` should be `continue` bug and refactored the module…

### LINKED ISSUE #2649: [Bug] Spreaking worker (`__poll_task_results`) dies when task metadata failed to retrieve from the server
**Describe the bug**
There is a `return` where there should have been a `continue`, resulting in that if the clause is reached, the node will never be able to send results to the server.
 
80300ab629f1af61393395adfb53fb4e62649a35

frankcorneliusmartin:
… a bit

Fixes #2649

### DIFF
--- a/vantage6-node/vantage6/node/__init__.py
+++ b/vantage6-node/vantage6/node/__init__.py
@@ -540,42 +540,15 @@ class Node:
         results : Result
             Result of the completed run
         """
-        # FIXME: why are we retrieving the result *again*? Shouldn't we just store
-        # the task_id when retrieving the task the first time?
-        response = self.client.request(f"run/{results.run_id}")
-        task_id = response.get("task", {}).get("id")
-
-        init_org_id = None
-        if not task_id:
-            self.log.error(
-                "Task id for run id=%s could not be retrieved", results.run_id
-            )
-        else:
-            init_org = self.client.request(f"task/{task_id}").get("init_org")
-            if not init_org:
-                self.log.error(
-                    "Initiator organization from task (id=%s) could not be retrieved!",
-                    task_id,
-                )
-            else:
-                init_org_id = init_org.get("id")
-
         data = {
             "log": results.logs,
             "status": results.status.value,
             "finished_at": datetime.datetime.now().isoformat(),
+            "result": results.data,
         }
-        if init_org_id:
-            # The result can only be encrypted when the initiator organization is
-            # known. Without it the run is still marked as finished - the result data
-            # is lost, but HQ stops re-sending this run indefinitely.
-            data["result"] = results.data
-        else:
-            self.log.error(
-                "Marking run id=%s as finished without result", results.run_id
-            )
-
-        self.client.run.patch(id_=results.run_id, data=data, init_org_id=init_org_id)
+        self.client.run.patch(
+            id_=results.run_id, data=data, init_org_id=results.init_org_id
+        )
 
         # notify HQ, other nodes, and clients about algorithm status change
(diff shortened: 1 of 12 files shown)


vantage6-node/vantage6/node/__init__.py
```diff
@@ -571,6 +530,58 @@ def __send_updates_finished_tasks(self) -> None:
...
+        Send the outcome of a completed algorithm run to HQ and notify the other
+        parties of the status change.
+
+        Parameters
+        ----------
+        results : Result
+            Result of the completed run
+        """
+        # FIXME: why are we retrieving the result *again*? Shouldn't we just store
+        # the task_id when retrieving the task the first time?
```
bartvanb: As we're already refactoring this, can we get rid of this FIXME?

vantage6-node/vantage6/node/__init__.py
```diff
@@ -571,6 +530,58 @@ def __send_updates_finished_tasks(self) -> None:
...
+            "finished_at": datetime.datetime.now().isoformat(),
+        }
+        if init_org_id:
+            # The result can only be encrypted when the initiator organization is
+            # known. Without it the run is still marked as finished - the result data
+            # is lost, but HQ stops re-sending this run indefinitely.
+            data["result"] = results.data
+        else:
+            self.log.error(
+                "Marking run id=%s as finished without result", results.run_id
```
bartvanb: Shouldn't we set status to failed in this case?

bartvanb: Hmm I guess I have a question about the issue in general: when does it happen that the task_id is not successfully retrieved from the server? My guess is because of connection issues?

If it is indeed connection issues, shouldn't we just try again later? Now we are patching the run with an empty result but that has the disadvantage that the task will never be successfully stored. I think this is better than the old situation (crashing thread) but it may not be optimal yet

vantage6-node/vantage6/node/__init__.py
frankcorneliusmartin: You are right to point this out, I did not want to make this PR any bigger than really addressing the early `return` as that was the bug, this was just for efficiency. However since this is landing in 5.0.3 I refactored further, should be optimized. 

In order to test it, I had to merge #2693 and #2692 into this PR. 

I must say I went into the rabbit hole to make a fix that I like as it required met to make the `init_org` are required setting (which it was already without enforcing it).

frankcorneliusmartin: First to make the `init_org_id` mandatory in the data models at the HQ (was already always supplied): 259ac4fcbdc614c29ca010c0281279d945656e7c

Then the node can assume this parameters is always set: 48a7143051f1e49231235698ae67362110f63f91 simplifying the code quite a bit.

vantage6-node/vantage6/node/__init__.py
frankcorneliusmartin: By the changes 259ac4f and 48a7143 this case no longer exists as the `init_org_id` is *always* present. The code that obtains the runs to be executed

```python
self.client.run.list(
            include_task=True, state=TaskStatusQueryOptions.OPEN.value, task_id=task_id
        )
```

should therefore always contain the `init_org_id`.

frankcorneliusmartin:
Did some manual testing with this PR and seems to work as expected at least in the semi happy flow. 

* Starting tasks for dataframes, preprocessing and analysis
* Manually killing them

bartvanb (approved):

PR #2665, opened 2026-07-20
Fix DetachedInstanceError at socket-emit sites in HQ

### LINKED ISSUE #2656: [Bug] Database Session Errors (DetachedInstanceError)
Run object detached in PATCH endpoint (run.py:743) — Trying to serialize run after socketio.emit detached it

Fixed: Re-fetch run from database before schema serialization
Node attributes detached in websockets.py — Accessing node.id, node.name, node.organization.id after session expired

Fixed: Load all node attributes early while session is active, pass as separate parameters
Task object detached in logging (task_post_base.py) — Trying to access task.collaboration.name for logging

Fixed: Use task.collaboration_id directly instead of lazy-loading

frankcorneliusmartin:
## Summary

Nodes and users occasionally saw a 500 (`DetachedInstanceError`) on things like run
status updates and task creation, because `socketio.emit()` yields the greenlet under
gevent and a concurrent handler's session cleanup could detach the ORM object the
current request still needed to read from. This PR fixes five such sites — three from
the linked issue, two more found by auditing `event.py` — by hoisting needed attributes
into locals before emitting, re-fetching before serialization, or reordering writes
ahead of the emit. A broader root-cause fix (`expire_on_commit=False` on both
sessionmakers) was considered and deliberately deferred; see the note below for why.

Closes #2656

## A note on an alternative, more root-cause fix

Before reading the rest of this PR: there is a broader alternative fix we considered and
deliberately did not implement here — `expire_on_commit=False` on the two sessionmakers
in `vantage6-backend-common/vantage6/backend/common/base.py`
(`session_flask_requests` and `session_outside_flask_requests`).

Right now both are built with autocommit/autoflush settings only, so `expire_on_commit`
defaults to SQLAlchemy's `True`. That means every `.save()` (i.e. every commit) expires
*all* loaded ORM instances, turning every later attribute access into a fresh query.
That's the dominant trigger behind this whole bug class: it's what makes a `run` or
`task` object fragile the moment a session is cleared after a commit, anywhere in the
codebase, not just at the sites fixed here.

Setting `expire_on_commit=False` would remove that trigger globally in one change,
instead of requiring every future `emit()`/lazy-relationship call site to be fixed one
at a time (as this PR does). We chose **not** to make that change in this PR because:

- It's a global behavioral change affecting every session in HQ and the algorithm
  store, not scoped to the bug being fixed.
- There is close to no existing test coverage for the websocket layer, so a regression
  from stale in-memory attribute values (e.g. any column with `server_default` or
  `onupdate` computed by the DB) would not be reliably caught. We checked and found no
  such columns in the current HQ/algorithm-store models, but that's a point-in-time
  fact, not a guarantee against future models introducing one.
- Given the stability risk of a global session-behavior change with thin test
  coverage backing it up, we preferred the narrower, reviewable, site-by-site fix in
  this PR for now.

**We should consider `expire_on_commit=False` as a deliberate follow-up**, ideally paired
with expanded websocket test coverage so a regression would actually be caught.

## What this does

Fixes `DetachedInstanceError` at several socket-emit sites in HQ. All are the same
shape: HQ commits a change, then calls `socketio.emit()`/`emit()` before reading
attributes off the same ORM objects for logging or serialization. Because HQ runs on
`gevent_uwsgi` with a Redis message queue, `emit()` performs network I/O and yields the
greenlet — during which a concurrent handler's `__cleanup()` can clear the shared
database session, detaching the objects the current handler is still holding.

Fixed in three complementary ways, applied per call site as appropriate:

- **Hoist**: read every ORM attribute needed after an emit into a local variable
  *before* the first emit in that code path.
- **Re-fetch**: re-load an object from the database immediately before it is
  serialized into a response, since it may have been detached in the interim.
- **Reorder**: where safe, do database writes / side-effect-free reads before emitting
  rather than after, removing the hazard rather than working around it.

Sites fixed:
- `resource/run.py` — `SingleRun.patch`: hoist before the two status-update emits,
  re-fetch before the response is serialized.
- `websockets.py` — `_handle_node_connection`, `on_disconnect`, `on_ping`,
  `__alert_node_status`, `_send_dataframe_deletion_instructions`,
  `on_algorithm_status_change`: hoisted node/task attributes; `__alert_node_status`
  and `_send_dataframe_deletion_instructions` now take plain ids/names rather than ORM
  objects, since they emit inside a loop.
- `resource/common/task_post_base.py` — `post_task`: log before emitting rather than
  after (keeps `task.collaboration.name` in the log line), re-fetch before the response
  is serialized.
- `resource/event.py` — two sites not covered by the original bug report:
  `KillNodeTasks.post` (hoist before emit) and `kill_task()` (mark runs killed
  *before* emitting the kill instruction, so the DB is consistent even if the runs are
  detached afterward — this one previously risked leaving runs silently un-killed).

Additionally, every socket handler body in `websockets.py` is now wrapped in
`try/finally` so `DatabaseSessionManager.clear_session()` always runs, even if the
handler raises partway through. Previously an exception mid-handler would leak the
session for the lifetime of the greenlet.

## Approach

Considered making the fix global via `expire_on_commit=False` on both sessionmakers in
`vantage6-backend-common`, which would remove the dominant trigger (every `.save()`
currently expires all loaded instances) across the whole codebase in one change. Decided
against it for this PR — it's a broader behavioral change than this bug warrants, and
there is close to no test coverage for the websocket layer to catch a regression if the
assumption about safety (no `server_default`/`onupdate` columns in these models) turns
out to be wrong somewhere I didn't check. Left as a candidate for a follow-up.

Also considered mirroring an existing fix for the same bug on an older branch. Diverged
from it in a few places: it introduces dual old/new calling conventions on private
methods with no old callers to support, and wraps some emits in broad
`except Exception` that logs and swallows the failure. Neither seemed worth carrying
forward — the private-method shims are dead weight, and silently losing a websocket
message seems worse than letting the failure surface.

## Deviations from the plan

- `on_algorithm_status_change` needed one more hoist than planned: `run.task.required_by`
  is read in a loop that emits on each iteration, the same hazard as
  `__alert_node_status`. Same technique as planned, just one more site than originally
  scoped.
- `on_ping` needed a small reorder in addition to the planned `try/finally`: the status
  write is now persisted before the reconnection-recovery branch runs (which itself
  emits), so the write isn't left stranded against an session that recovery already
  cleared.
- Two bugs surfaced in the new tests themselves during verification, not in the fix:
  one read a detached attribute after the simulated session clear (fixed by capturing
  ids upfront and running inside a request context, matching how `kill_task` runs in
  production); the other initially skipped this repo's DB test lifecycle and broke five
  unrelated tests in `test_cleanup.py` (fixed by inheriting `TestResourceBase`).
- `black` isn't installed in this environment; used `ruff format` (its black-compatible
  formatter) instead.

## Testing

- `test_run.py::test_run_patch_survives_session_clear_during_emit` — patches
  `SocketIO.emit` to clear the session as a side effect, reproducing the race
  deterministically. Confirmed this fails with `500 != 200` against the pre-fix code
  and passes after.
- `test_task.py::test_task_post_survives_session_clear_during_emit` — same technique for
  task creation.
- `test_task.py::test_kill_task_marks_runs_killed_when_session_cleared_on_emit` — asserts
  runs actually end up `KILLED` in the database even when the session clears mid-emit.
- `test_websockets.py::test_cleanup_runs_when_handler_raises` (new file) — asserts
  `clear_session()` is still called for six handlers when the handler body raises.
- Full HQ suite: 143/143 passing (`uv run python utest.py --hq`).
- `ruff format --check` and `ruff check` clean on all changed files.

### DIFF
--- a/vantage6-hq/tests_hq/resources/test_run.py
+++ b/vantage6-hq/tests_hq/resources/test_run.py
@@ -2,6 +2,9 @@ import datetime
 import logging
 import uuid
 from http import HTTPStatus
+from unittest.mock import patch
+
+from flask_socketio import SocketIO
 
 from vantage6.common import logger_name
 from vantage6.common.enum import AlgorithmStepType, RunStatus
@@ -12,6 +15,7 @@ from vantage6.hq.model import (
     Run,
     Task,
 )
+from vantage6.hq.model.base import DatabaseSessionManager
 
 from .test_resource_base import TestResourceBase
 
@@ -136,3 +140,47 @@ class TestResources(TestResourceBase):
             col.delete()
             org1.delete()
             org2.delete()
+
+    def test_run_patch_survives_session_clear_during_emit(self):
+        """
+        A socket emit yields the greenlet, after which the database session may have
+        been cleared by another handler's cleanup, detaching the run. The endpoint
+        should still be able to serialize its response.
+
+        See https://github.com/vantage6/vantage6/issues/2656.
+        """
+        org = Organization(name=str(uuid.uuid1()))
+        col = Collaboration(name=str(uuid.uuid1()), organizations=[org])
+        col.save()
+        node = self.create_node(organization=org, collaboration=col)
+        task = Task(
+            image="localhost/algorithms/test:local", collaboration=col, init_org=org
+        )
+        task.save()
+        run = Run(task=task, organization=org, status=RunStatus.PENDING.value)
+        run.save()
+
+        try:
+            headers = self.login_node(node)
+
+            # Clear the session on emit, which is what a concurrent handler's
+            # __cleanup() does while the emit has yielded the greenlet.
+            with patch.object(
+                SocketIO,
+                "emit",
+                side_effect=lambda *a, **kw: DatabaseSessionManager.clear_session(),
+            ):
+                response = self.app.patch(
+                    f"/api/run/{run.id}",
+                    headers=headers,
+                    json={"status": RunStatus.COMPLETED.value},
+                )
+
+            self.assertEqual(response.status_code, HTTPStatus.OK)
+            self.assertEqual(response.json["status"], RunStatus.COMPLETED.value)
+        finally:
+            run.delete()
+            task.delete()
+            node.delete()
+            col.delete()
+            org.delete()
--- a/vantage6-hq/tests_hq/resources/test_task.py
+++ b/vantage6-hq/tests_hq/resources/test_task.py
@@ -1,6 +1,8 @@
 import logging
 from http import HTTPStatus
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
+
+from flask_socketio import SocketIO
 
 from vantage6.common import logger_name
 from vantage6.common.enum import AlgorithmStepType, RunStatus, TaskStatus
@@ -15,6 +17,7 @@ from vantage6.hq.model import (
     Study,
     Task,
 )
+from vantage6.hq.model.base import DatabaseSessionManager
 from vantage6.hq.model.rule import Operation, Scope
 from vantage6.hq.resource.event import kill_task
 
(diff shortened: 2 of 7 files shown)


frankcorneliusmartin:
Closed this one as a better solution has been found: https://github.com/vantage6/vantage6/pull/2670

PR #2671, opened 2026-07-23
Guarantee WebSocket DB session cleanup via teardown_request (#2654)

### LINKED ISSUE #2654: [Bug] WebSocket DB Connection Leaks
**Describe the bug**
WebSocket event handlers run in a Flask app context (so g is available), but Flask's after_request hook does not fire for WebSocket events. Every handler called __cleanup() at the end, but without try/finally — any exception occurring mid-handler would skip cleanup, permanently leaking a DB connection from the pool.

I wrapped the DB-touching body of every handler in try/finally to guarantee __cleanup() always runs:

- on_connect
- on_disconnect
- on_ping
- on_algorithm_status_change
- on_node_info_update
- on_algorithm_log
- on_node_metrics_update
- on_dataframe_deleted

frankcorneliusmartin:
## Summary

WebSocket event handlers run inside a Flask request context, but Flask's `after_request` hook never fires for socket events (they don't go through the full HTTP dispatch) and is also skipped whenever a handler raises. Each handler therefore cleared its own DB session with a trailing `self.__cleanup()` call — unguarded, so any exception mid-handler skipped it and permanently leaked a pooled connection. This replaces the per-handler cleanup with a single `teardown_request` hook that runs on every request-context pop, including on exceptions, so a handler can no longer leak a connection. Chosen over the alternative of wrapping all eight handlers in `try/finally`, which is the same pattern repeated eight times and easy to forget on the next handler.

Closes #2654

## What this does

- Registers one `@app.teardown_request` handler in the shared backend app factory (`backend/common/__init__.py`) that calls `clear_session()`. `teardown_request` fires on every request-context pop — for HTTP requests, for socket events, and on exceptions — which is exactly the guarantee `after_request` cannot give for sockets.
- Guards `clear_session()` (`backend/common/base.py`) with `if "session" in g:`, since teardown now fires for request contexts that never opened a session (socket events that touch no DB, pre-session early returns). Without the guard, accessing `g.session` would raise `AttributeError`.
- Removes the now-redundant manual cleanup from `hq/websockets.py`: the eight `self.__cleanup()` calls, the `__cleanup()` method, and its now-unused `DatabaseSessionManager` import. No `try/finally` added anywhere.

## Approach

`teardown_request` was verified empirically (Flask 3.1.3 + Flask-SocketIO 5.6.1, the pinned versions) to fire once per socket event and to fire even when the handler raises, while `after_request` never fires for sockets. `teardown_request` is used rather than `teardown_appcontext` because Flask reuses a single app context across nested same-app request contexts, so `teardown_appcontext` would not fire for a socket handler dispatched synchronously inside another request (the nested-handler scenario from #2656); `teardown_request` fires per request context and is correct in both the flat and nested cases. The HTTP `before_request`/`after_request`/errorhandler flow is left unchanged; teardown also firing for HTTP requests is a harmless idempotent double `clear_session()`.

## Deviations from the plan

- Also removed the now-unused `DatabaseSessionManager` import from `websockets.py` (only the deleted `__cleanup` used it) — needed for a clean lint.
- The "no session" test is a direct `clear_session()`-without-session unit test rather than a socket dispatch: teardown-exception propagation through the socket test client is unreliable to assert, so a direct call is deterministic (and confirmed to fail without the guard).
- The tests register the teardown hook in a mini-app mirroring the factory, so they guard the mechanism and the guard, not the factory wiring itself; that wiring is smoke-covered by the full suite booting the app through the factory.

## Testing

- `tests_hq/resources/test_websockets_cleanup.py` (new): `test_session_released_when_handler_raises` (a handler that opens a session then raises still has `clear_session` run via teardown) and `test_clear_session_noop_without_session` (the guard). Confirmed the latter fails with `AttributeError` without the guard.
- HQ suite: 141 passed (139 baseline + 2 new).
- Algorithm-store suite: 49 passed (shared `backend-common` change does not regress the store).
- `ruff check` clean; `ruff format` applied to `websockets.py`.

### DIFF
--- a/vantage6-hq/tests_hq/resources/test_websockets_cleanup.py
+++ b/vantage6-hq/tests_hq/resources/test_websockets_cleanup.py
@@ -1,0 +1,79 @@
+"""
+Regression tests for issue #2654 (WebSocket DB connection leaks).
+
+WebSocket event handlers run inside a Flask request context, but Flask's
+``after_request`` hook does not fire for socket events, and is skipped whenever a
+handler raises. Cleanup of the database session is therefore driven by a
+``teardown_request`` hook, which runs on every request-context pop - including on
+exceptions - so a handler that raises cannot leak a pooled connection.
+"""
+
+from unittest import TestCase
+from unittest.mock import patch
+
+from flask import Flask
+from flask_socketio import SocketIO
+
+from vantage6.hq.model import Organization
+from vantage6.hq.model.base import Database, DatabaseSessionManager
+
+
+class TestWebsocketSessionCleanup(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        Database().connect("sqlite://", allow_drop_all=True)
+
+    @classmethod
+    def tearDownClass(cls):
+        Database().clear_data()
+        Database().close()
+
+    @staticmethod
+    def _build_app():
+        """Minimal Flask + SocketIO app with the same teardown hook the backend
+        factory registers (backend/common/__init__.py)."""
+        app = Flask(__name__)
+        app.config["SECRET_KEY"] = "test"
+
+        @app.teardown_request
+        def _remove_db_session_on_teardown(exc):
+            DatabaseSessionManager.clear_session()
+
+        socketio = SocketIO(app, async_mode="threading")
+        return app, socketio
+
+    def test_session_released_when_handler_raises(self):
+        """A handler that opens a session and then raises must still have its
+        session cleared, via teardown."""
+        app, socketio = self._build_app()
+
+        @socketio.on("boom")
+        def boom(data):
+            # opens a per-request session (lazily, like any model access)
+            Organization(name="ws-teardown-org").save()
+            raise RuntimeError("handler blew up mid-event")
+
+        client = socketio.test_client(app)
+
+        with patch.object(
+            DatabaseSessionManager,
+            "clear_session",
+            wraps=DatabaseSessionManager.clear_session,
+        ) as clear_spy:
+            try:
+                client.emit("boom", {})
+            except RuntimeError:
+                pass  # the test client re-raises the handler's exception
+
+        self.assertTrue(
+            clear_spy.called,
+            "teardown_request must clear the session even when the handler raises",
+        )
+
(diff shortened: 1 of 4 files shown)


vantage6-hq/tests_hq/resources/test_websockets_cleanup.py
```diff
@@ -0,0 +1,79 @@
+"""
+Regression tests for issue #2654 (WebSocket DB connection leaks).
```
bartvanb: I'm not a fan of mentioning the issue number in a test file. At the very least, the emphasis should be on what it tests and not on what issue it arised from

vantage6-hq/tests_hq/resources/test_websockets_cleanup.py
```diff
@@ -0,0 +1,79 @@
...
+    @staticmethod
+    def _build_app():
+        """Minimal Flask + SocketIO app with the same teardown hook the backend
+        factory registers (backend/common/__init__.py)."""
+        app = Flask(__name__)
+        app.config["SECRET_KEY"] = "test"
+
+        @app.teardown_request
+        def _remove_db_session_on_teardown(exc):
+            DatabaseSessionManager.clear_session()
```
bartvanb, line 43: I want to note that this unit test tests that teardown_request works in a certain way - it does however not check that the decorator is used in vantage6. If we were to remove the decorator from the vantage6 code, this test would pass happily while the code would be broken. Can we get around this easily? (Or is my analysis wrong?)

vantage6-hq/tests_hq/resources/test_websockets_cleanup.py
frankcorneliusmartin: 50f93b2494a3b25341166dd19035a5627d669b8b

vantage6-hq/tests_hq/resources/test_websockets_cleanup.py
frankcorneliusmartin: Good catch. I updated it to fail when you remove the decorator from the class. 84f78e09ddee5ee9ff3eef38712651f5d318e193

bartvanb (approved):

PR #2677, opened 2026-07-27
fix(node-chart): quote policy list entries in ConfigMap

### LINKED ISSUE #2644: [Bug] Node chart's ConfigMap seems to render `allowed_users`/`allowed_organizations` without quotes — possible string/int mismatch?
### Context
We're migrating a local vantage6 deployment from 4.x to 5.0.2, running on Kubernetes via microk8s. While debugging a crash that appeared partway through the migration, we (with some help from Claude digging through the node/common source) traced it back to how the Node chart renders `allowed_users`/`allowed_organizations` into the ConfigMap. We're not 100% sure this is a chart bug rather than something off in our own setup/config, so happy to be corrected if we're missing something.

### Environment
- Chart: `node` (Helm), version `5.0.2` (`oci://ghcr.io/vantage6/helm/node --version 5.0.2`)
- Same code appears to be present on `main` as of the commit currently tagged `version/5.0.2`
- `vantage6-common` / `vantage6-node`: 5.0.2
- Kubernetes engine: microk8s

### What we observed
`charts/node/templates/node-configmap.yml` renders each entry of `policies.allowed_users` and `policies.allowed_organizations` as a bare `- {{ . }}`, without `| quote`:

https://github.com/vantage6/vantage6/blob/version/5.0.2/charts/node/templates/node-configmap.yml#L49-L60

```gotemplate
{{- if .Values.node.policies.allowed_users }}
allowed_users:
  {{- range .Values.node.policies.allowed_users }}
  - {{ . }}
  {{- end }}
{{- end }}
{{- if .Values.node.policies.allowed_organizations }}
allowed_organizations:
  {{- range .Values.node.policies.allowed_organizations }}
  - {{ . }}
  {{- end }}
{{- end }}
```

We noticed other scalar fields in the same file that need to preserve their string type *do* use `| quote` (e.g. `logging.datefmt`, `logging.format`, `private_docker_registries[].password`), so it looked to us like this might just be a gap rather than intentional — but flagging in case there's a reason it's this way.

### Steps to reproduce
With IDs written as strings in `values.yaml` (as we understood the node code expects them):

```yaml
node:
  policies:
    allowed_users:
      - "1"
      - "2"
    allowed_organizations:
      - "1"
      - "root"
```

```bash
helm template test oci://ghcr.io/vantage6/helm/node --version 5.0.2 -f values.yaml --show-only templates/node-configmap.yml
```

### What we expected
The rendered `config.yml` inside the ConfigMap to keep the values quoted:

```yaml
allowed_users:
  - "1"
  - "2"
allowed_organizations:
  - "1"
  - "root"
```

### What we actually got
The numeric entries lose their quotes and get parsed back as YAML integers:

```yaml
allowed_users:
  - 1
  - 2
allowed_organizations:
  - 1
  - root
```

### Why we think this matters
As far as we could tell from reading `vantage6-common`, the node-side check (`check_user_allowed_to_send_task`) compares `str(init_user_id) == value` / `str(init_org_id) == value` against these entries. If `value` ends up as a Python `int` (from the unquoted YAML) instead of a `str`, that comparison would never be `True`, which — in our case — pushed execution into a fallback lookup path that eventually raised:

```
AttributeError: 'str' object has no attribute 'get'
```

We could be wrong about this being the actual cause (it's possible our own config or setup triggered something adjacent), but wanted to raise it in case it's a real gap, since even quoting the IDs correctly in `values.yaml` didn't seem to prevent the issue for us.

### Possible fix (untested upstream, but seems to resolve it for us locally)
```diff
       {{- if .Values.node.policies.allowed_users }}
       allowed_users:
         {{- range .Values.node.policies.allowed_users }}
-        - {{ . }}
+        - {{ . | quote }}
         {{- end }}
       {{- end }}
       {{- if .Values.node.policies.allowed_organizations }}
       allowed_organizations:
         {{- range .Values.node.policies.allowed_organizations }}
-        - {{ . }}
+        - {{ . | quote }}
         {{- end }}
       {{- end }}
```

Side note: `allowed_algorithms` and `allowed_algorithm_stores` a few lines above have the same `- {{ . }}` pattern (lines 39-40 and 45-46). Those currently hold regex/URL strings that happen to survive unquoted YAML, so we haven't seen a concrete problem there — just mentioning it in case it's relevant for consistency.

frankcorneliusmartin:
## Summary
The Node Helm chart's ConfigMap rendered `policies.allowed_users`,
`allowed_organizations`, `allowed_algorithms`, and `allowed_algorithm_stores` list
entries as bare `- {{ . }}`, so numeric IDs (e.g. a user/org ID of `1`) were re-parsed
as YAML **integers** in the node's `config.yml`. The node-side allow-list check compares
them as strings (`str(init_user_id) == entry`), so numeric entries never matched and
execution fell through to a fallback path that raised
`AttributeError: 'str' object has no attribute 'get'`. This adds `| quote` to the four
policy list fields so entries keep their string type. Note: quoting the IDs in
`values.yaml` alone does not help — Go templates emit `{{ . }}` unquoted regardless, so
the fix has to live in the template.

Closes #2644

## What this does
Changes the four `range` bodies in `charts/node/templates/node-configmap.yml` that emit
policy list entries from `- {{ . }}` to `- {{ . | quote }}`. Rendered output now reads
`- "1"`, `- "root"`, etc. `allowed_algorithms`/`allowed_algorithm_stores` are included
for consistency (same latent gap; regex/URL strings are unambiguously strings and
unaffected in meaning).

## Approach
`| quote` is already the convention in this same file for every other scalar whose
string type matters (`logging.datefmt`, `logging.format`,
`private_docker_registries[].password`). The unquoted policy lists were an inconsistency,
not a deliberate choice, so this brings them in line. The fix is confined to the chart —
the node-side Python comparison is left unchanged, as fixing the rendered config is the
correct minimal fix and Python-side hardening is a separable concern. `algorithm_env`
(which holds `KEY=value` strings, not policy IDs) is intentionally left as-is.

## Deviations from the plan
None — implemented as planned.

## Testing
No unit-test harness exists for this chart. Verified with Helm directly:
- `helm template` with numeric and string IDs — before the change entries rendered as
  bare `- 1` / `- root`; after the change they render quoted as `- "1"` / `- "root"`
  across all four fields.
- `helm lint` with values supplied — passes clean (only the cosmetic "icon is
  recommended" INFO). The unrelated `node-secrets.yml` b64enc lint error only appears
  when linting with empty default values and reproduces identically on `main`.

### DIFF
--- a/charts/node/templates/node-configmap.yml
+++ b/charts/node/templates/node-configmap.yml
@@ -37,25 +37,25 @@ data:
       {{- if .Values.node.policies.allowed_algorithms }}
       allowed_algorithms:
         {{- range .Values.node.policies.allowed_algorithms }}
-        - {{ . }}
+        - {{ . | quote }}
         {{- end }}
       {{- end }}
       {{- if .Values.node.policies.allowed_algorithm_stores }}
       allowed_algorithm_stores:
         {{- range .Values.node.policies.allowed_algorithm_stores }}
-        - {{ . }}
+        - {{ . | quote }}
         {{- end }}
       {{- end }}
       {{- if .Values.node.policies.allowed_users }}
       allowed_users:
         {{- range .Values.node.policies.allowed_users }}
-        - {{ . }}
+        - {{ . | quote }}
         {{- end }}
       {{- end }}
       {{- if .Values.node.policies.allowed_organizations }}
       allowed_organizations:
         {{- range .Values.node.policies.allowed_organizations }}
-        - {{ . }}
+        - {{ . | quote }}
         {{- end }}
       {{- end }}
       require_algorithm_pull: {{ .Values.node.policies.require_algorithm_pull }}


bartvanb (approved):

bartvanb:
Need to fix ruff before merging this

frankcorneliusmartin:
> Need to fix ruff before merging this

Since we also merged lots of other PRs with this issue I decided to merge this one as well. We will fix all ruff errors in the PR that addresses the ruff issues as well: #2682

PR #2678, opened 2026-07-27
fix(hq): release DB sessions in background workers and make pool configurable (#2651)

### LINKED ISSUE #2651: [Bug] Database sessions where not released on exception
**Describe the bug**
If an exception happend where a database connection is openend at the server it resulted in a dangling connection. If this happend several times, the postgresql instance might run out of connections and put the server in a dead state. 

I added `finally` blocks:

```python
finally:
    DatabaseSessionManager.clear_session()
```
to prevent this from happening.

I also made the connection pool settable, I used this to decrease the number of connections to the database to try to remedy before I applied the `finally` fix everywhere. But these settings can be very useful when scaling the application. By default postgres accepts 100 connections, so you need to divide these between the instances you have. The settings:

```python
pool_size=5,
max_overflow=10,
```

help you preventing from requesting too many connections.

> [!TIP]
> These are SQLAlchemy connection pool settings.
>`pool_size=5` — the number of persistent connections kept open to Postgres at all times, even when idle. Think of it as your baseline pool.
> 
> `max_overflow=10` — how many extra connections can be opened on top of pool_size when demand spikes. These are temporary; once released they're closed rather than returned to the pool.
>
> So your maximum possible concurrent connections = 5 + 10 = 15.

frankcorneliusmartin:
Merge #2693 and #2692 first, then this review should become easier.

---

## Summary

Database work that runs *outside* the Flask request lifecycle — the background
worker threads (`__node_status_worker`, `__runs_data_cleanup_worker`) — was not
covered by the request hooks that normally release the DB session. When such a
thread raised, its session (and the pooled connection behind it) was stranded;
repeated over time this exhausts the Postgres connection pool and wedges HQ. This
adds a `session_scope()` context manager that guarantees the session is released
on every exit path, and wraps the background workers with it. It also makes the
SQLAlchemy connection pool (`pool_size` / `max_overflow`) configurable so
connections can be budgeted across horizontally-scaled instances.

The socket-event half of the reported problem is intentionally **not** touched
here — it is fixed more robustly by #2671 (teardown_request), which this branch
builds on. Pool settings are opt-in and pass-through: when unset, SQLAlchemy's
existing defaults (5 / 10) are preserved, and they are never sent to SQLite
engines (which would raise), so behaviour is unchanged unless an operator opts in.

Closes #2651

> [!NOTE]
> This PR is based on `fix/2654-websocket-db-connection-leaks` (#2671), not
> `main`, to keep the diff focused on the non-socket changes. Once #2671 merges,
> GitHub will retarget this PR to `main` automatically.

## What this does

- **`session_scope()` context manager** (`backend/common/base.py`, exposed as
  `DatabaseSessionManager.session_scope()` in `hq/model/base.py`): yields a
  session and calls `clear_session()` in a `finally`, so a raised exception can no
  longer strand a session outside the Flask request lifecycle.
- **Background workers wrapped** (`hq/__init__.py`): the node-status loop's DB
  block and the runs-data cleanup call now run inside `session_scope()`. This also
  covers the invalid-`runs_data_cleanup_days` early-return path in
  `cleanup_runs_data`, which acquired a session and returned without releasing it.
- **Configurable connection pool** (`backend/common/base.py`): `_connect` accepts
  `pool_size` / `max_overflow`, and a new `_build_engine_kwargs()` adds them to
  `create_engine` only when set *and* the backend is not SQLite. Threaded through
  both `Database.connect()` wrappers; `run_hq` / `run_store` read
  `database_pool_size` / `database_max_overflow` from config.
- **Deployment wiring**: both Helm ConfigMaps (HQ + store) emit the two keys when
  set; both config docs (`docs/hub/components/yaml/*.yaml`) and both CLI templates
  (`*_config.j2`) document them as optional, commented settings.

## Approach

Session lifecycle is centralised rather than sprinkled: the Flask side already
owns cleanup in one place (before/after/teardown hooks), so the non-Flask side
gets one context manager instead of per-call-site `try/finally`. The pool change
is deliberately pass-through — SQLAlchemy's `QueuePool` already defaults to 5/10,
so not passing anything preserves current behaviour exactly, and the SQLite guard
keeps the sqlite-backed test/dev engines working (they use a pool that rejects
`max_overflow`).

## Deviations from the plan

- The runs-data cleanup session scope was placed around the **call** in
  `__runs_data_cleanup_worker` rather than **inside** `cleanup_runs_data`.
  Wrapping inside the function cleared the session mid-call and broke six existing
  `test_cleanup` tests that hold the session across the call. Wrapping at the
  worker entry point (mirroring `__node_status_worker`) reaches the same
  guarantee, keeps `cleanup_runs_data` reusable and unchanged, and adds no test
  churn.

## Testing

- HQ suite: 146 passed (141 baseline + 5 new).
- Algorithm-store suite: 49 passed. CLI suite: 29 passed.
- New `tests_hq/model/test_session_scope.py`: session released on normal exit and
  when the body raises (the actual bug); `_build_engine_kwargs` omits pool keys
  for SQLite even when provided, includes them for Postgres when set, omits when
  unset.
- `ruff check .`: no net new findings versus the base branch; `ruff format
  --check` clean on changed files.
- `helm template` (HQ + store): pool keys render when set and are absent by
  default; both `*_config.j2` templates parse.


daniel-chou-rainho (approved):
I'm still not too familiar with the codebase but I believe I understand the problem and your general approach for solving it makes sense.

From what I understand, a background worker in HQ calling `db.Node.get_online_nodes()` behaves differently than the same call made from a REST endpoint, because the latter comes with a hook that releases the DB connection afterward, and the background worker doesn't. To replicate that for background workers, you added a wrapper (`session_scope()`) that opens a session and closes it in the end.

A lot of files show up as changed, but I've skimmed most of them (Keycloak changes, docs, config templates) and only reviewed what I believe are the core files:

- `vantage6-backend-common/vantage6/backend/common/base.py`
- `vantage6-hq/vantage6/hq/__init__.py`
- `vantage6-hq/vantage6/hq/model/base.py`
- `vantage6-algorithm-store/vantage6/algorithm/store/__init__.py`
- `vantage6-algorithm-store/vantage6/algorithm/store/model/base.py`

The only critique I have, a nit, is that almost every change here is accompanied by nearly the same explanatory comment. For example, in `hq/__init__.py` the two background-worker wrapping calls repeat almost the same comments. Maybe just keep the full explanation at the deepest reference, i.e. `backend/common/base.py`, and/or keep the usage-site comments shorter? But again, just a nit, also fine with leaving things as they are.

PR #2680, opened 2026-07-29
Fix server URL sent in POST task by UI

bartvanb:
We also need to check if this needs a v5 fix


frankcorneliusmartin (approved):
LGTM. Do you make sure to log that we need to check this also in version 5?

bartvanb:
I checked and this does not need a v5 fix

Issue #2676, opened 2026-07-26
[Change Request] Default to none for unspecifed list of allowed algos


lsago:
**Description**

An incomplete configuration file, or a new user overlooking the algorithm policy settings, should not result in any algorithm being allowed to run. When neither `allowed_algorithms` nor `allowed_algorithm_stores` is configured, the node should reject all algorithm images by default.

Administrators who want to allow every algorithm can still do so explicitly

bartvanb:
I actually made a low-severity security issue for exactly this last week - see [here](https://github.com/vantage6/vantage6/security/advisories/GHSA-hx2w-rjmx-934x) but note that link is for project admins only (for now).

Our solution there is to introduce a `production: false/true` setting in the node - if production is true and no policies are defined, we simply refuse to start the node as it couldn't run any algorithms anyway. But with the production toggle, we can still make it easy for people to test algorithms in development environments

bartvanb:
Closing this in favor of the draft advisory

PR #2685, opened 2026-08-02
Reject algorithms when no allow-list is configured

lsago:
See: https://github.com/vantage6/vantage6/issues/2676

This PR might no longer apply – as Bart points out, they are working towards a `production: false/true` toggle. I'd favor defaulting to deny all, perhaps production = true by default? Are there other production/dev settings?

This PR is a starting point. Needs discussion, careful review, and testing. Tim or someone else at MDW will continue work on this PR / proposal.


bartvanb:
Indeed let's do as follows:

- production: true -> default deny all
- production: false -> default allow all

And default value for `production` should be true

PR #2692, opened 2026-08-03
Fix KeycloakRealmImport race on fresh install

frankcorneliusmartin:
## Problem

On a fresh `helm install`/`upgrade` the realm import intermittently never ran. The `KeycloakRealmImport` CR got stuck with:

```
Done=False, Started=False
HasErrors=True  "No existing Deployment found, waiting for it to be created"
```

No import Job was ever created, so the `vantage6` realm was never imported.

**Root cause:** Helm applied the `Keycloak` CR and the `KeycloakRealmImport` CR at the same timestamp. The Keycloak StatefulSet takes ~1m45s to become Ready. The realm-import controller reconciled during that gap, found no Deployment, recorded `HasErrors`, and did not reliably re-reconcile once Keycloak became Ready. Deleting/recreating the CR only helped if the recreate happened to land after Keycloak was Ready — hence the intermittency.

## Fix

Order the two CRs explicitly rather than depending on winning the timing race:

- **`keycloak-wait-for-ready-job.yml`** (new) — a `post-install` hook Job at weight `0` that runs `kubectl wait --for=condition=Ready keycloak/<release>-kc`. Helm blocks on hook Job completion before moving to the next weight, which is what makes the ordering work.
- **`keycloak-realm-import-rbac.yml`** (new) — ServiceAccount + Role + RoleBinding for that Job. Namespace-scoped, single rule: `get/list/watch` on `keycloaks`.
- **`keycloak-realm-import.yml`** — the CR is now a `post-install` hook at weight `1`, so it is only reconciled once Keycloak reports Ready.

`kubectl wait` is used rather than an HTTP probe because the condition the realm-import controller itself gates on is `status.conditions[Ready]` on the Keycloak CR — an API-server field reflecting the operator's own readiness logic (DB connectivity, full rollout), not something an HTTP check can observe.

### Why `post-install` only

Deliberately **not** `post-upgrade`. Re-creating the CR on every upgrade would make the operator run a fresh import Job against an already-populated realm, risking overwriting state created at runtime (e.g. users the backend adds via `backend-admin-client`). The realm is imported once, on install.

`before-hook-creation` is kept explicit (it is Helm's default anyway) so a stale CR from a previous release is still replaced on a fresh install.

Trade-off: nothing clears a wedged CR automatically any more. Recovery is `devspace purge` (or deleting the CR and reinstalling) rather than a re-upgrade.

### Cleanup

Because the CR is now a hook resource it is not tracked in the release manifest, so `helm uninstall` will not remove it. `dev/purge_keycloak_storage.sh` deletes it alongside the existing PVC/PV cleanup — the same category of untracked state that script already handles.

### DevSpace selector

The operator labels the transient realm-import Job pod with the same `app.kubernetes.io/instance=vantage6-auth-kc` used to select the Keycloak server pod. DevSpace matched the `Completed` import pod and stalled (`DevSpace is waiting, because Pod vantage6-auth-realm-import-xxxxx has status: Completed`), then exited. Both selectors now also require `app: keycloak`, the label the existing `keycloak-service.yml` Service already uses to target the real server pods.

## Configuration

New `keycloak.realmImport.waitForKeycloak` block (`enabled` / `image` / `timeout`). Setting `enabled: false` fully restores the previous behaviour — the CR renders as a plain resource with no hooks and no RBAC.

## Testing

- `helm lint` passes.
- `helm template` verified with the feature enabled, disabled, and with `keycloak.realmImport: null` (the null case initially crashed with a nil-pointer error and is now guarded).
- Not yet verified against a live cluster.


charts/auth/values.yaml
```diff
@@ -27,6 +27,16 @@ keycloak:
 
   realmImport:
     enabled: true
+    # Ensures the KeycloakRealmImport CR is only applied once the Keycloak CR reports
+    # Ready, via a post-install hook Job. Prevents a race on fresh installs where the
+    # realm-import controller reconciles before the Keycloak StatefulSet exists and
+    # never gets requeued once it comes up.
+    waitForKeycloak:
+      enabled: true
+      image: rancher/kubectl:v1.31.0
```
bartvanb, line 36: I checked and apparently this image is a lightweight `kubectl` runner. So that's fine.

I think we did something similar in other places, but I think we need a strategy on updating the versions of these images. Let's discuss this

bartvanb (approved):

charts/auth/values.yaml
frankcorneliusmartin: @bartvanb we discussed this and we agreed that we should create a separate image for this purpose as this image is not well maintained and we do not want to run images from external users.

frankcorneliusmartin: When I started on working on our own image I found a DHI image that has `kubectl` installed. I think I prefer a maintained DHI image over maintaining one ourselves, so I update this PR to use that image instead.

I also upgraded some all other support images, with the exception of RabbitMQ which seems to be EOL on dhi.io: https://hub.docker.com/hardened-images/catalog/dhi/rabbitmq/images (which I find odd, as DHI is only launched so recently).

| old image | new image | 
| -- | -- | 
| curlimages/curl:8.7.1 | dhi.io/curl:8-alpine |
| busybox:latest | dhi.io/busybox:1 | 
| rancher/kubectl:v1.31.0 | dhi.io/kubectl:1 | 

> [!NOTE]
> that as the hardened `curl` image no longer has a shell, the UX goed a bit down as we do not have a shell to print the nice error messages (like `Attemt X`, and `Done`). For me this is a reasonable sacrifice as a shell comes with vulnerabilities. 

> [!NOTE]
> that I did not use hashes of the DHI images as I would like that the semi-latest image is used from the DHI registry in order to patch found vulnerabilities automatically.

frankcorneliusmartin:
Hi @bartvanb,

Since we started the discussion here: https://github.com/vantage6/vantage6/pull/2692#discussion_r3720965068 I made quite a few changes including more secure images. I would appreciate if you could review them again. A few additional notes:

1
Because we are moving to DHI image the cluster running one or more components needs to be logged in to `dhi.io`. I documented this (for all three envs: dev, sandbox and prod), and provided alternative images in case you really don't want to create an account. However I feel that using DHI as a default is reasonable. Let me know what you think.

2
Alternatively, we could consider forking the DHI image to our GitHub CR, but then we are also responsible for the update cycle. 

3
I had to enable Prometheus in order to test these changes. I fixed a bug that Prometheus was scraping the wrong address, so that is included in this PR as well. 

---

See all changes here:

https://github.com/vantage6/vantage6/pull/2692/files/ea147a0346a936a7522bf3ad698295c96b48fdb5..49a184b8ba97024edc2e55a28cb34aa00c240ee9

Proof that Prometheus works as included:
image

charts/auth/values.yaml
```diff
@@ -27,6 +27,16 @@ keycloak:
...
+    # Ensures the KeycloakRealmImport CR is only applied once the Keycloak CR reports
+    # Ready, via a post-install hook Job. Prevents a race on fresh installs where the
+    # realm-import controller reconciles before the Keycloak StatefulSet exists and
+    # never gets requeued once it comes up.
+    waitForKeycloak:
+      enabled: true
+      image: dhi.io/kubectl:1
+      # Keep below Helm's --timeout (default 5m): the release aborts on that budget
+      # first, so raising this without also raising --timeout has no effect.
+      timeout: 300s
```
bartvanb, line 32: You forgot to update docs/hub/components/auth_config.yaml

Also, that is where we include the comments on what the options mean. The values.yaml should be kept clean from comments to prevent that we have these comments in multiple places which will lead to conflicting/outdated explanation in one of those places

charts/common/templates/_wait-scripts.tpl
```diff
@@ -1,32 +1,54 @@
...
+- {{ $interval | quote }}
+{{- /* The target Service may not resolve at all yet on a fresh install, so
+       retry DNS and HTTP errors too, not just the transient ones. */}}
+- --retry-connrefused
+- --retry-all-errors
+- --output
+- /dev/null
+- --write-out
+- {{ $writeOut | squote }}
+- {{ $url | quote }}
```
bartvanb, line 70: Hard to read this code (not that the original was that pretty though) - I guess not possible to have it another way if we use the hardened image. I would really appreciate it if there would be logs on attempts. Maybe this can be achieved by not having `--silent`? Or specify something like `-v`?

charts/hq/templates/prometheus/prometheus-deployment.yml
```diff
@@ -22,7 +22,7 @@ spec:
       # initialize by making the volume writable inside the container
       initContainers:
       - name: volume-permissions
-        image: busybox:latest
+        image: dhi.io/busybox:1-alpine
```
bartvanb: This is not configurable. Therefore users would *have to* create an account with DHI, which was supposedly optional.

docs/introduction/quickstart.rst
```diff
@@ -50,6 +50,31 @@ Then, install the vantage6 command line interface (CLI) by running:
...
+
+    kubectl create secret docker-registry dhi-pull-secret \
+      --docker-server=dhi.io \
+      --docker-username=<your-docker-username> \
+      --docker-password=<your-access-token-or-password> \
+      --docker-email=<your-email>
+
+You can use your Docker Hub username and password, or create a
+`personal access token <https://docs.docker.com/security/access-tokens/>`_
+for better security.
```
bartvanb: I really don't like that running the sandbox becomes more complicated. It is not user friendly. Do we really need to use DHI for dev / sandbox envs? If so, can we find another solution - creating the secret on behalf of the user or something (is not really nice I guess but neither is this)?

devspace.yaml
```diff
@@ -356,7 +367,11 @@ deployments:
           volumePath: ${HQ_DATABASE_MOUNT_PATH}
           k8sNodeName: ${K8S_NODE_NAME}
         prometheus:
-          enabled: false
+          enabled: true
```
bartvanb, line 358: I see you needed prometheus for testing but that's not really used often in the dev env right? Can we disable this again? Also requires devs to answer less questions before setting up their env

bartvanb (changes requested):
I don't like placing the burden for DHI on the user. I think creating DHI user accounts can be a significant burden for someone just starting out with vantage6. 

I don't think node admins have to use DHI right? 

If so, could we have an easy-to-use option to not use DHI images for dev/sandbox? That would alleviate my concerns.

devspace.yaml
frankcorneliusmartin: I also feel this tension, but I decided I was pro enabling tis by default for as: we would otherwise forget to test (it has never worked before in v5), in the future I foresee that prom stats are displayed directly in the UI making it less of an optional component and its also only a single extra container and question. 

But lets discuss.

docs/introduction/quickstart.rst
frankcorneliusmartin: I agree that his downgrades the usability, I also see that we have alternative solutions that sacrifice maintainability and security to make it user-friendly.  I chose security in the PR, let also discuss this.

devspace.yaml
frankcorneliusmartin: We discussed:

* Prometheus enabled by default (inc) nodes
* Simplify the PV questions

docs/introduction/quickstart.rst
frankcorneliusmartin: We discussed:

* We create a copy of the dhi.io image on release time
* References in the code are changed to this copy
* Release candidates should spot any issues with using the newest dhi images

devspace.yaml
frankcorneliusmartin: Turned on Prometheus for the nodes in the Devspace environment: 558662dd934200ff714f4c2c87c5f636c9be184b

Removed Devspace Prometheus profile as its enabled by default (removed also some code :)), consolidated mount path questions: e4c070124701bc93ef4a5ca3aaa85ea096eddc6f

docs/introduction/quickstart.rst
frankcorneliusmartin: f307d0548ee65c05d8f4b44d9818892f17b0f10d

charts/auth/values.yaml
frankcorneliusmartin: Updated the helm values examples in the docs: fa14e715401baa36faa6595aeb44f0ba65386f88
Removed comments from `values.yaml`: 99ec39ae7b26e8313147069177fe704d53a23aa7

charts/common/templates/_wait-scripts.tpl
frankcorneliusmartin: Tried to improve the formatting, removed the `--silent` now you get a Request failed error for every attempt. e54c579288c5b24654821f3be5bc0b5e0bba3a84

charts/hq/templates/prometheus/prometheus-deployment.yml
frankcorneliusmartin: A bit hidden but in: f307d0548ee65c05d8f4b44d9818892f17b0f10d

frankcorneliusmartin:
I published the support images today manually under `5.0.2`. We need to merge the GH workflow to main in order to be able to publish them along with the images. I also added the options to trigger the mirror workflow manually so you can create mirrors for development purposes.

.github/workflows/release.yml
```diff
@@ -231,6 +231,15 @@ jobs:
...
         run: make support-image TAG=$version PUSH_REG=true TAG_LATEST=$tag_latest
 
+  # Re-publish the third-party images defined in docker/mirror-images.txt.
+  mirror-images:
+    name: Re-publish third-party images
+    needs: version
+    uses: ./.github/workflows/mirror-images.yml
+    secrets: inherit
+    with:
+      version: ${{ needs.version.outputs.version }}
```
bartvanb, line 241: Should we test this workflow before or after merging this PR? 
I'm fine with either - only if we do it after, we should work on it directly so we don't forget

bartvanb: Nice to refactor this in separate files btw

.github/workflows/mirror-images.yml
```diff
@@ -0,0 +1,69 @@
...
+      version:
+        description: Tag to publish the images under, i.e. the vantage6 version
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Tag to publish the images under, i.e. the vantage6 version
+        required: true
+        type: string
```
bartvanb, line 35: So this is for manual call. So let's say we're on version 5.5.5, what version would be provide manually? Would we simply overwrite the current images with new 5.5.5 tag? Or something like 5.5.5-post1? 
We should also document this

tools/mirror-images.sh
```diff
@@ -0,0 +1,64 @@
...
+  target="${REGISTRY}/${name}:${TAG}"
+  echo "${source} (${digest})"
+
+  if [ "${PUSH_REG}" = "true" ]; then
+    echo "  -> ${target}"
+    regctl image copy --referrers --digest-tags "${source}@${digest}" "${target}"
+  else
+    echo "  -> ${target} (skipped, PUSH_REG != true)"
+  fi
+done < <(sed -e 's/#.*//' -e '/^[[:space:]]*$/d' "${image_list}")
```
bartvanb, line 104: This doesn't ever push the `latest` tag right? Should we consider pushing `latest` or `5` as well? I guess it's not really necessary as the infra will pick up the current version, but could be useful option if users e.g. want to set in their config that they always want to use the latest that we published

docs/hub/components/yaml/auth_config.yaml
```diff
@@ -65,6 +65,17 @@ keycloak:
...
+    # Wait for Keycloak to report Ready before applying the realm import. This
+    # prevents a race on fresh installs where the realm-import controller
+    # reconciles before the Keycloak StatefulSet exists.
+    waitForKeycloak:
+      enabled: true
+      # Keep this below Helm's --timeout (default 5m): the release aborts on that
+      # budget first, so raising this without also raising --timeout has no effect.
+      timeout: 300s
+      # image used to check whether Keycloak is ready.
+      image: ghcr.io/vantage6/infrastructure/kubectl:<vantage6-version>
```
bartvanb: Related to my other comment about `5` and `latest` tags: if we have those, we could put them here in this config. But it's not too important

docs/hub/deploy.rst
```diff
@@ -32,6 +32,35 @@ their own (advanced usage).
...
+         image: my-registry.example.org/curl:5.0.0
+   auth:
+     keycloak:
+       realmImport:
+         waitForKeycloak:
+           image: my-registry.example.org/kubectl:5.0.0
+   hq:
+     prometheus:
+       image: my-registry.example.org/prometheus:5.0.0
+       initImage: my-registry.example.org/busybox:5.0.0
```
bartvanb: This seems incomplete. It only lists the new helper images, but not the infrastructure images that also need to be pulled. 

Also, I wonder if this documentation is important enough to be here? I know AI likes generating more docs, but I'd leave this out since it's a pretty advanced option that few will use on a pretty important docs page. Those that want to can read the full configuration file example, where they are explained.

vantage6/vantage6/cli/dev/start.py
```diff
@@ -11,9 +11,6 @@
 
 
 @click.command()
-@click.option(
-    "--with-prometheus", is_flag=True, default=False, help="Enable Prometheus"
```
bartvanb, line 15: If we are removing this option here, I think it should also be removed from dev/populate.py

docs/hub/components/yaml/hub_config.yaml
```diff
@@ -71,12 +76,12 @@ global:
...
       maxAttempts: 60
       intervalSeconds: 5
-      image: curlimages/curl:8.7.1
+      image: ghcr.io/vantage6/infrastructure/curl:<vantage6-version>
     store:
       enabled: true
       maxAttempts: 60
       intervalSeconds: 5
-      image: curlimages/curl:8.7.1
+      image: ghcr.io/vantage6/infrastructure/curl:<vantage6-version>
```
bartvanb: The changes in this file are currently not included in hub_config.j2. I guess that's fine right?

.pre-commit-config.yaml
```diff
@@ -1,6 +1,10 @@
...
-        language_version: python3.13
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.16.2
+  hooks:
+    # Run the linter.
+    - id: ruff-check
+      args: [ --fix ]
+    # Run the formatter.
+    - id: ruff-format
```
bartvanb, line 10: awesome

devspace.yaml
```diff
@@ -88,7 +82,8 @@ pipelines:
         --task-directory $TASK_DIRECTORY \
         --task-namespace $TASK_NAMESPACE \
         --starting-port-number $NODE_PROXY_STARTING_PORT_NUMBER \
-        --populate-marker $POPULATE_MARKER
+        --populate-marker $POPULATE_MARKER \
+        --prometheus-enabled
```
bartvanb, line 86: I mentioned this elsewhere but presumably this option can be removed? Only I wonder if that affects the sandbox. Have you tested the sandbox on this branch yet?

bartvanb (commented):
I like where this is going but still have some comments and questions

.github/workflows/release.yml
frankcorneliusmartin: I suggest we test this when we are building a beta version, then we are already using the release branch to build it so why not test at least the manual workflow at that point.

frankcorneliusmartin: Ok silly thing, we cannot trigger it manually as the workflow (older version is fine) needs to exist on the main branch. So we can only run this workflow once we merged it into main. So we need to be prepared on the release day to apply some hotfixes on the workflow.

I logged this into our pre-release obsidian list.

.github/workflows/mirror-images.yml
frankcorneliusmartin: This is indeed for a manual call, only to be used when working on a release and you need some update in one of these images. The version is something that you would need to supply manually, see this section in GH docs: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#providing-inputs

I updated the docs to include documentation on the `mirror-images` workflow: e68c71776078dbf3ed129bb5ef1a16195fe2fc5f

tools/mirror-images.sh
frankcorneliusmartin: Yes this crossed my mind too. It's simple enough and it echo's how we release the other images. I made it a flag which is settable by the manual dispatch and derived when using the triggered (mirrored the alpine behaviour): 42e8ba194a70c78cdc8eb3ca34f091b97564eb68

docs/hub/components/yaml/auth_config.yaml
frankcorneliusmartin: I used `5.0`, as I find `latest` ugly and we dont have a `5`: 06fb6540e7cb5ade28a5fe85cdb2a5af1849639e

docs/hub/deploy.rst
frankcorneliusmartin: I did not show all images there as I also found it to be quite verbose for an advanced option, so I removed a bit. Now that I looked at it again, I feel that almost no one is ever going to use this.. So I completely removed it from the docs:  73251d7e5016d951cb5eeb8c609d2e2fc30462d9

vantage6/vantage6/cli/dev/start.py
frankcorneliusmartin: No we can't, `sandbox` is indeed using this. I dont want to scope creep any further.

devspace.yaml
frankcorneliusmartin: https://github.com/vantage6/vantage6/pull/2692#discussion_r3895112140. Sandbox to be tested.

docs/hub/components/yaml/hub_config.yaml
frankcorneliusmartin: Yes, definitely should not be there; way to advanced to have it in there. With the same reasoning as: https://github.com/vantage6/vantage6/pull/2692#discussion_r3895072494 I also removed it here from the docs: cc51f8bd4bcf3a9c4c8184087cdfd656ea285698

devspace.yaml
frankcorneliusmartin: I've ran some tests:
✅ Devspace environment
✅ Sandbox, without Prometheus, operator already installed
✅ Sandbox, without Prometheus, operator not installed
🐞  Sandbox, including Prometheus, operator already installed

The Prometheus in sandbox failed due to an error unrelated to this issue. The wiring from the hub helm values to the HQ values lost the volume mount for Prometheus. I created a separate PR for this: https://github.com/vantage6/vantage6/pull/2729

PR #2704, opened 2026-08-06
Update Keycloak Operator version to 26.6.2 and enhance installation d…

### LINKED ISSUE #2661: [Bug] Installing keycloak operator fails (in dev/sandbox environment)
### Summary

`v6 auth install-keycloak` installs an incomplete set of CRDs for Keycloak Operator versions >= 26.7.0. The operator registers reconcilers for `KeycloakOIDCClient` and `KeycloakSAMLClient` unconditionally at startup, but the corresponding CRDs (`keycloakoidcclients.k8s.keycloak.org` and `keycloaksamlclients.k8s.keycloak.org`) are never applied. Because Java Operator SDK aborts startup when any registered informer fails to sync, the operator crash-loops indefinitely.

### Environment

- vantage6 v5 (`auth` Helm chart / `v6 auth install-keycloak` CLI)
- Keycloak Operator image: `quay.io/keycloak/keycloak-operator:26.7.0`
- Docker Desktop Kubernetes (single-node)

### Steps to reproduce

1. Run `v6 auth install-keycloak` (or otherwise let it resolve `--operator-version` to a version >= 26.7.0, e.g. via the "latest GitHub release" lookup).
2. Observe `kubectl get crd | grep keycloak` only shows:
   ```
   keycloakrealmimports.k8s.keycloak.org
   keycloaks.k8s.keycloak.org
   ```
3. The `keycloak-operator` deployment crash-loops. Logs show repeated `Reflector` failures on startup, one per missing CRD:
   ```
   ERROR [io.fabric8.kubernetes.client.informers.impl.cache.Reflector] (vert.x-eventloop-thread-0)
   listSyncAndWatch failed for k8s.keycloak.org/v2alpha1/namespaces/<ns>/keycloaksamlclients, will stop:
   java.util.concurrent.CompletionException: io.fabric8.kubernetes.client.KubernetesClientException:
   Failure executing: GET at: https://10.96.0.1:443/apis/k8s.keycloak.org/v2alpha1/namespaces/<ns>/keycloaksamlclients?resourceVersion=0.
   Message: Not Found.
   ```
   (Same failure for `keycloakoidcclients`.) The 404 causes JOSDK to abort operator startup, and the container restarts, repeating the same failure forever.

### Root cause

`vantage6/vantage6/cli/auth/install.py::cli_auth_install_operator` hardcodes the list of CRD files to install:

```python
crd_files_to_install = [
    "keycloaks.k8s.keycloak.org-v1.yml",
    "keycloakrealmimports.k8s.keycloak.org-v1.yml",
    "kubernetes.yml",
]
```

This matched the full CRD bundle for Keycloak Operator <= 26.6.2. Starting with Keycloak 26.7.0, the operator ships an experimental declarative client-management feature (`KeycloakOIDCClient` / `KeycloakSAMLClient`, see [[Keycloak docs](https://www.keycloak.org/operator/managing-clients)](https://www.keycloak.org/operator/managing-clients)), and the `keycloak-k8s-resources` repo added two new CRD manifests for it:

- `keycloakoidcclients.k8s.keycloak.org-v1.yml`
- `keycloaksamlclients.k8s.keycloak.org-v1.yml`

The operator binary registers reconcilers/informers for these resource kinds unconditionally regardless of whether vantage6 actually uses them - so if they're not installed, the operator fails at boot.

### Expected behavior

`v6 auth install-keycloak` should install whatever CRDs the resolved operator version actually requires, so the operator can start successfully, without vantage6 needing to use the new client CRDs itself.

### Proposed fix

Apply the two new CRDs alongside the existing ones, but tolerate failure for operator versions where they don't exist yet (< 26.7.0), so pinning an older `--operator-version` keeps working. A patch along these lines has already been applied locally in `install.py`:

```python
required_crd_files = [
    "keycloaks.k8s.keycloak.org-v1.yml",
    "keycloakrealmimports.k8s.keycloak.org-v1.yml",
]
optional_crd_files = [
    "keycloakoidcclients.k8s.keycloak.org-v1.yml",
    "keycloaksamlclients.k8s.keycloak.org-v1.yml",
]
# apply required_crd_files as before (check=True)
# apply optional_crd_files with check=False, warn (don't fail) if not found
# apply kubernetes.yml last
```

### Workaround (manual, for anyone hitting this now)

```bash
kubectl apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/26.7.0/kubernetes/keycloakoidcclients.k8s.keycloak.org-v1.yml
kubectl apply -f https://raw.githubusercontent.com/keycloak/keycloak-k8s-resources/26.7.0/kubernetes/keycloaksamlclients.k8s.keycloak.org-v1.yml
kubectl rollout restart deployment/keycloak-operator -n <namespace>
```

### Additional notes

- Since these client-management CRDs are explicitly marked "experimental" by Keycloak (may introduce breaking changes), it's worth deciding whether vantage6 should pin `--operator-version` to a known-good release by default rather than always resolving to "latest", to avoid future breakage from upstream experimental features shipping ahead of vantage6's compatibility testing.
</content>

jvsoest:
~Related to~ Fixes issue #2661 so fixing the keycloak version for now


vantage6/vantage6/cli/auth/install.py
```diff
@@ -1,78 +1,39 @@
...
 from vantage6.cli.common.k8s_utils import run_kubectl_command
 from vantage6.cli.k8s_config import KubernetesConfig, select_k8s_config
 
 
-def _get_latest_keycloak_version() -> str:
-    """
-    Get the latest Keycloak Operator version from GitHub releases. Exits when unable
-    to find what version is the latest version.
+SUPPORTED_KEYCLOAK_OPERATOR_VERSION = "26.6.2"
+KEYCLOAK_CLIENT_CRDS_MIN_VERSION = (26, 7, 0)
```
frankcorneliusmartin, line 15: These need to be moved to: `vantage6/vantage6/cli/globals.py` and add a line of documentation to it.

Also the versions seem different, is this intentional?

vantage6/vantage6/cli/auth/install.py
```diff
@@ -1,78 +1,39 @@
...
-        headers["Authorization"] = f"token {github_token}"
-
-    try:
-        response = requests.get(
-            github_api_url,
-            timeout=5,
-            headers=headers,
+def _parse_operator_version(version: str) -> tuple[int, int, int]:
+    """Validate and convert a Keycloak operator version to a comparable tuple."""
+    if not re.fullmatch(r"\d+\.\d+\.\d+", version):
```
frankcorneliusmartin: I always find these difficult to read these. I would prefer to use the build in:
```python
from packaging.version import Version
try:
  v = Version(version)
except:
  raise click....
```

vantage6/vantage6/cli/auth/install.py
```diff
@@ -1,78 +1,39 @@
...
-            error(
-                f"Failed to fetch latest Keycloak Operator version from GitHub: {e}. "
-            )
-        info("Please specify a version manually using the --operator-version flag.")
-        sys.exit(1)
-    except (requests.RequestException, KeyError, ValueError) as e:
-        error(f"Failed to fetch latest Keycloak Operator version from GitHub: {e}. ")
-        info("Please specify a version manually using the --operator-version flag.")
-        sys.exit(1)
+    return tuple(int(part) for part in version.split("."))
```
frankcorneliusmartin: The this also becomes more readable:
```python
return (v.major, v.minor, v.micro)
```

frankcorneliusmartin (changes requested):
Some minor comments from me.

@bartvanb I did not raise the issue that not the standard flow was done for the exception handling (`info(.)` + `sys.exit(.)`) as the click errors seems also to be a nice alternatice. For consistency sake I think we should change them in this PR as well. What do you think.

@jvsoest as this has been mostly worked on by @bartvanb I want to await his review aswell.

vantage6/vantage6/cli/auth/install.py
bartvanb: Disagree that these constants need to be moved: we used to do that, but at some point we decided to no longer do that for constants that are just used in a single file - that's also better practice as it's clear but does not create a huge globals.py.

Same question about the version difference though. Maybe it's because if a custom version is to be used, it should be newer than the default? Not sure why we'd want to enforce that though?

vantage6/vantage6/cli/auth/install.py
```diff
@@ -215,6 +186,9 @@ def _wait_for_operator_ready(
         "You can check the operator status manually with: "
         f"kubectl get deployment keycloak-operator -n {namespace}"
     )
+    raise click.ClickException(
+        f"Keycloak Operator did not become ready within {timeout} seconds."
+    )
```
bartvanb, line 194: This will print ugly exception to the terminal if CLI command is unsuccessful. The error+warning above it are already sufficient error, this raise will not add new information

frankcorneliusmartin: You are right, but let me note that the `click.ClickException` is a good way to inform the users as it is catched by the click wrapper and invokes the `.show` method on this Exception type which does not print the entire exception but a message, depending on the exception it terminates the process and gives a non zero exit code.

But again, I agree lets make it conform the rest of the code base.

frankcorneliusmartin (changes requested):

vantage6/vantage6/cli/auth/install.py
bartvanb: Hm didn't know that. Maybe that would actually be better then? Should we create issue to (let AI) do this everywhere?

vantage6/vantage6/cli/auth/install.py
frankcorneliusmartin: ~~a2ad9a98c819819f68e257557ab347325ba5ef60~~ Reverted this change: bcfbcf848e4d86f567c77c477ae199b2676797d9.

This is because the `Version` class accepts incomplete version specification (and pre/post releases).

vantage6/vantage6/cli/auth/install.py
frankcorneliusmartin: a2ad9a98c819819f68e257557ab347325ba5ef60

vantage6/vantage6/cli/auth/install.py
frankcorneliusmartin: Its a bit cleaner from the code perspective, but I don't think its worth it. So I am going to close it.

vantage6/vantage6/cli/auth/install.py
frankcorneliusmartin: Since `--operator-version` can be set higher than the pinned 26.6.2, the CRD list has to match whatever tag the user picks. The client CRDs only exist upstream from 26.7.0, so applying them against an older tag would fail. `KEYCLOAK_CLIENT_CRDS_MIN_VERSION` handles that: nothing happens on the default, and the extra CRDs get installed once someone picks 26.7.0 or later.

So this is actually valid.

PR #2722, opened 2026-08-25
Prevent data leak from non-pandas errors

### LINKED ISSUE #2499: [Bug] Handling pandas errors is not preventing all sensitive data errors from being printed
**Describe the bug**
The idea behind the `@handle_pandas_errors` decorator is that it catches errors in pandas dataframe actions so that no sensitive data is printed. However, non-pandas specific errors can arise from pandas actions, e.g. a ValueError when one tries to convert a string to an integer in a pd.DataFrame which then fails.

**Expected behavior**
Improve decorator to prevent printing sensitive data

daniel-chou-rainho:
## Related issue

Fixes #2499

## Description

The `@handle_pandas_errors` decorator only caught exceptions from `pandas.errors` but pandas operations often raise plain Python errors that contain data values.

Those were passed on untouched and ended up in the algorithm logs, which are sent to HQ and can be read by the researcher.

## Changes

- The decorator now replaces every exception instead of only the pandas-specific ones.
- The vantage6 `AlgorithmError` exceptions are re-raised unchanged, so that algorithm developers can still report their own errors to the user.
- The generic message keeps the exception type and the function name, for example `An error of type ValueError occurred in 'select_rows'`.
- The new error is raised with `from None`. Without it, the original error is still printed as part of the traceback.
- Renamed the decorator to `handle_data_errors`, as it is no longer specific to pandas. `handle_pandas_errors` is kept as an alias, so existing algorithms keep working.
- Removed `PANDAS_ERROR_TYPES` and `_collect_pandas_error_types()`, which are no longer used.
- Updated the algorithm development documentation and added 2 unit tests.

### DIFF
--- a/docs/algorithms/develop.rst
+++ b/docs/algorithms/develop.rst
@@ -257,24 +257,26 @@ vantage6 algorithm tools:
    for instance that a certain data value is not a valid date.
 
    To help you keep such sensitive information private, vantage6
-   provides a decorator that can be used to handle pandas errors. This decorator will
-   catch all pandas errors and return a generic error message. You can use this
-   decorator by adding it to your algorithm function:
+   provides a decorator that can be used to handle errors from data manipulation. This
+   decorator will catch such errors and return a generic error message. You can use
+   this decorator by adding it to your algorithm function:
 
    .. code:: python
 
-      from vantage6.algorithm.tools.error_handling import handle_pandas_errors
+      from vantage6.algorithm.tools.error_handling import handle_data_errors
 
-      @handle_pandas_errors
+      @handle_data_errors
       def my_function(data: pd.DataFrame):
          return data
 
    All data extraction and preprocessing functions provided by the vantage6 algorithm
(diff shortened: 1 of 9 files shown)


docs/algorithms/develop.rst
```diff
@@ -257,24 +257,26 @@ vantage6 algorithm tools:
...
-   and a generic error message will be returned instead of the traceback. Note that
-   this decorator does not catch all errors, so you should still be careful with the
-   data you handle in your algorithm functions.
+   tools package are decorated with the ``handle_data_errors`` decorator, so that any
+   error occurring during the execution of the function will be caught and a generic
+   error message will be returned instead of the traceback.
+
+   The errors defined in ``vantage6.algorithm.tools.exceptions`` are an exception to
+   this rule: they are not replaced, as their messages are written by you and are
+   meant to be shown to the user. Make sure not to include any sensitive data in them.
```
bartvanb, line 279: Neat that these are excluded

bartvanb (approved):

PR #2723, opened 2026-08-27
Add 'pulling image' run status

### LINKED ISSUE #2230: [Feature Request] Algorithm status 'pulling image'
**Problem description**
Sometimes it can take a while to download an algorithm image, especially on nodes with a slow connection.

**Desired solution**
It would be nice if the run gets status 'downloading image' while the image is being downloaded. This is especially valuable to users working with the UI, who may be waiting on a short task while it is being computed, and may be impatient why it's taking so long (like me, often, in developing stuff, wondering if it may be a bug instead)

daniel-chou-rainho:
## Related issue

Closes #2230

## Description

While a node pulls an algorithm image, the run stays at `pending`. That status means no node has started the run yet, usually because it is offline, so a user waiting in the UI cannot tell a slow download from a node that never picked the task up.

The node already knows what is happening. `compute_run_pod_status` maps kubelet's `ContainerCreating` reason during the pull, but `__wait_until_pod_running` used `INITIALIZING` as a sentinel meaning "keep waiting" and returned only once the pod reached a final state. Every status before that was discarded, so nothing was ever reported to HQ.

Simply returning a new status from the mapper would break the node. The wait loop treated anything other than `INITIALIZING` as final, so it would start the log stream against a container that has not been created yet. The loop condition had to change along with the new status.

## Before / after

Left is `release/5.0`, right is this branch. Same task, same image, same machine.

https://github.com/user-attachments/assets/e87d6ace-649e-4b10-af92-88b4ce66b733

The old build sits at `Pending` for the whole pull, which is indistinguishable from a node that never picked the task up. The new one shows `Pulling image` until the container starts.

## Changes

- Added `RunStatus.PULLING_IMAGE` ("pulling image") and included it in `alive_statuses()`, so the derived task status stays correct while a run is pulling.
- Split the pending-reason map so `ContainerCreating` maps to `PULLING_IMAGE`, while `PodInitializing` keeps `INITIALIZING`.
- Replaced the `!= INITIALIZING` sentinel in `__wait_until_pod_running` with a positive check for `ACTIVE` or a finished status, so any non-final status keeps the loop waiting instead of leaking out as a result.
- Added an `on_status_change` callback to `run()` and `__wait_until_pod_running`, called only when the status actually changes, so the run is not patched on every pod event.
- Extracted `Node.__report_run_status()` from the existing patch-and-emit pair, and used it for both the intermediate and the final reports.
- UI: added the status to `TaskStatus` and to the sort pipe, mapped it to the active chip type and status group, and added the "Pulling image" translation.
- Added 2 unit tests for the new mapping.

## Testing

Tested with DevSpace on a 529 MB algorithm image:

- Runs now report `pending` → `initializing` → `pulling image` → `active`, and spend 14 seconds showing "Pulling image" on a cold pull.
- kubelet measured the pull at 12.2s, against under a second to create and start the container, which is why `ContainerCreating` is reported as an image pull even though it also covers container creation.

Because DevSpace runs the node on the same machine, `pending` passes almost instantly and the run sits in `pulling image` for nearly the whole startup.

### DIFF
--- a/vantage6-node/vantage6/node/k8s/container_manager.py
+++ b/vantage6-node/vantage6/node/k8s/container_manager.py
@@ -6,6 +6,7 @@ import re
 import threading
 import time
 import uuid
+from collections.abc import Callable
 from itertools import groupby
 from pathlib import Path
 
@@ -349,6 +350,7 @@ class ContainerManager:
         token: str,
         databases_to_use: list[dict],
         action: AlgorithmStepType,
+        on_status_change: Callable[[RunStatus], None] | None = None,
     ) -> RunStatus:
         """
         Run a vantage6 algorithm on the Kubernetes cluster.
@@ -371,6 +373,9 @@ class ContainerManager:
             Metadata of the databases to use.
         action: AlgorithmStepType
             The action to perform
+        on_status_change: Callable[[RunStatus], None] | None
+            Called whenever the run changes status while the algorithm is being
+            started. Not called with the final status that is returned here.
 
         Returns
         -------
@@ -622,6 +627,7 @@ class ContainerManager:
         status = self.__wait_until_pod_running(
             run_io=run_io,
             label=f"app={run_io.container_name}",
+            on_status_change=on_status_change,
         )
 
         # start streaming logs to HQ
@@ -728,7 +734,12 @@ class ContainerManager:
         finally:
             w.stop()
 
-    def __wait_until_pod_running(self, run_io: RunIO, label: str) -> RunStatus:
+    def __wait_until_pod_running(
+        self,
+        run_io: RunIO,
+        label: str,
+        on_status_change: Callable[[RunStatus], None] | None = None,
+    ) -> RunStatus:
         """"
         This method monitors the status of a Kubernetes POD created by a task job and
         waits until it transitions to a 'Running' state or another terminal state
@@ -749,6 +760,9 @@ class ContainerManager:
             RunIO object that contains information about the run
         label : str
             Label selector to identify the POD associated with the task job.
+        on_status_change: Callable[[RunStatus], None] | None
+            Called whenever the pod changes to a status that is not final yet, such as
+            RunStatus.PULLING_IMAGE.
 
         Returns
         -------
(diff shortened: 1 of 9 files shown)


vantage6-ui/src/assets/localizations/en.json
```diff
@@ -907,6 +907,7 @@
     "Killed": "Killed by user",
     "NoDockerImage": "Non-existing Docker image",
     "Pending": "Pending",
+    "PullingImage": "Pulling image",
```
frankcorneliusmartin: Pulling image is quite a technical term. I would replace this with `Downloading algorithm` so its more clear to the user what is happing without the knowledge about images and containers.

This should only be used for display. Internally we should remain using the technical term.

vantage6-node/vantage6/node/k8s/container_manager.py
frankcorneliusmartin, line 1: So you've made the `on_status_change` an optional argument, which is reasonable as the container run will work perfectly fine without it. Also the server is OK, as it just does not receive the intermediate states. However there is - I think - also no one calling this without the `on_status_change`, so I would opt to remove it as I dont see any use case for not supplying it. 

I agree with the choice of doing this through a callback rather than a injection of the client in the container manager.

vantage6-node/vantage6/node/k8s/jobpod_state_to_run_status_mapper.py
```diff
@@ -46,12 +46,22 @@
...
-# The reasons reported for a 'waiting' state on a POD's container (when the POD is
-# Pending) that are related to an ongoing image pull or container initialization
-POD_INITIALIZATION_RELATED_REASONS_FOR_STILL_PENDING = {
+# The reason reported for a 'waiting' state on a POD's container (when the POD is
+# Pending) that indicates the algorithm image is still being fetched onto the node.
+#
+# Note: kubelet reports 'ContainerCreating' for both pulling the image and creating the
+# container. Creating takes well under a second, while pulling a large image dominates
+# the wait, so this is reported as PULLING_IMAGE. Separating the two would require
+# reading the pod's events.
```
frankcorneliusmartin: I am not super fond of these super verbose comments (that are likely spit out by Claude? Sorry if thats not the case).

frankcorneliusmartin (changes requested):

vantage6-ui/src/assets/localizations/en.json
daniel-chou-rainho: Changed to "Downloading algorithm" in d7b982626.
Internal value and `RunStatus.PULLING_IMAGE` unchanged.

vantage6-node/vantage6/node/k8s/container_manager.py
daniel-chou-rainho: Made required in 3c08a8372.

vantage6-node/vantage6/node/k8s/jobpod_state_to_run_status_mapper.py
daniel-chou-rainho: Trimmed in 32d5f3246.

PR #2724, opened 2026-08-28
Fix #2650: make socketio max_decode_packets configurable

### LINKED ISSUE #2650: [Bug] Bundled WS messages resulting in lost messages
**Describe the bug**
Internally the `python-socketio` library bundles messages to limit the number of requests (or it happens at the server before ingestion, I am not sure). This results in the server in a message it was unable to unpack the payload, which results in that these messages get lost.

This happens when you have many nodes (10 in my case) that send lots of status messages, I fixed it by setting the  `max_decode_packets` to something bigger.

```python
# Increase max_decode_packets to handle rapid message bursts
# Default is 16, which causes "Too many packets in payload" errors
# when many messages are batched together
Payload.max_decode_packets = 128
```

We need to investigate what this setting exactly does, what a sensible value would be or that we make this settable by the user in the configuration file.

> [!TIP]  
> ### Only for HTTP protocol
> This setting is only relevant for the polling http protocol. If we decide not to support it anymore we might consider just closing this issue, see #2646. Note that the upgrade mechanism (which we might keep) is affected by this setting.

agu2347:
## Problem

python-engineio (used internally by Flask-SocketIO) bundles multiple websocket packets into a single payload when using the HTTP long-polling transport. By default it only allows `Payload.max_decode_packets = 16` packets per payload. As described in #2650, when many nodes send status messages in a short time window (e.g. many nodes connecting/reporting around the same time), this limit is exceeded and the server fails to decode the remaining packets in that payload, silently dropping the corresponding websocket messages.

The issue reporter confirmed that manually raising `Payload.max_decode_packets` to a higher value fixed message loss for their 10-node deployment, and asked that this be investigated and made configurable rather than left at the library default.

## Fix

`HQApp.setup_socket_connection()` now sets `Payload.max_decode_packets` once, before either the primary (`async_mode="gevent_uwsgi"`) or the fallback `SocketIO(...)` constructor is attempted, so it applies regardless of which branch is taken. The value is read from a new `socketio_max_decode_packets` config key, defaulting to `128` (the value that resolved message loss for the issue reporter) when unset.

```python
Payload.max_decode_packets = self.ctx.config.get(
    "socketio_max_decode_packets", 128
)
```

This follows the same `self.ctx.config.get("key", default)` convention already used elsewhere in this method (`rabbitmq`, `cors_allowed_origins`) and elsewhere in the file.

## Testing

Added `vantage6-hq/tests_hq/test_socket_connection.py`, which extracts the real `setup_socket_connection` method body via `ast`/`inspect` (so it always tests the actual production code, not a re-implementation) and runs it against lightweight `SocketIO`/`DefaultSocketNamespace` fakes, since a full Flask/DB/Keycloak app context isn't needed to exercise this logic. It covers:
- default config leaves `max_decode_packets` at 128 (raised from the library default of 16)
- a configured `socketio_max_decode_packets` value is honored
- the setting is applied even when `rabbitmq` is configured
- the setting is applied even when the `gevent_uwsgi` `SocketIO()` constructor raises and the method falls back to the plain constructor

I also verified locally (outside the committed test) that with the original code, `Payload.max_decode_packets` stays at the library default of 16 regardless of config, reproducing the reported bug, and that the fixed code correctly raises/configures it in all four scenarios above.

Fixes #2650

### DIFF
--- a/vantage6-hq/tests_hq/test_socket_connection.py
+++ b/vantage6-hq/tests_hq/test_socket_connection.py
@@ -1,0 +1,152 @@
+"""
+Regression tests for HQApp.setup_socket_connection().
+
+python-engineio (used by Flask-SocketIO) bundles multiple websocket packets
+into a single HTTP long-polling payload, and by default only allows 16
+packets per payload (``engineio.payload.Payload.max_decode_packets``). When
+many nodes send status messages in a short time window this limit is
+exceeded, causing the server to silently fail to decode the remaining
+packets in the payload and lose messages (see GH issue #2650).
+
+These tests exercise the real ``setup_socket_connection`` method body (parsed
+directly out of ``vantage6.hq.HQApp`` via ``ast``, so the tests run against
+the actual production source rather than a re-implementation of it), with
+Flask/SocketIO/rabbitmq collaborators replaced by lightweight fakes so the
+test does not require a running Flask app, database, or message broker.
+"""
+
+import ast
+import inspect
+import logging
+
+import pytest
+from engineio.payload import Payload
+
+from vantage6.hq import HQApp
+
+
+def _extract_method_source(cls, method_name):
+    """Pull the literal source of a method off a class via ast.
+
+    This ensures the test always exercises whatever the method currently
+    contains (including future edits), rather than a hand copied version of
+    it that could silently drift out of sync with the real implementation.
+    """
+    src = inspect.getsource(cls)
+    tree = ast.parse(src)
+    class_node = tree.body[0]
+    for item in class_node.body:
+        if isinstance(item, ast.FunctionDef) and item.name == method_name:
+            return ast.get_source_segment(src, item)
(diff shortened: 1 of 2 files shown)


frankcorneliusmartin:
Hi @agu2347,

Thanks for addressing this. I have a couple of questions. 

Looking at your profile it seems to me that you are not a user of vantage6 but are looking to contribute to open source projects? Would you want to introduce yourself in the community: https://github.com/orgs/vantage6/discussions/283?

On topic for the PR:
Is there any good use case that you need the HTTP polling mechanism to work? We internally discussed last week to drop that polling support by default (to be re-enabled though config). This is because the polling mechanism seems to have many issues ranging from frequent disconnects to putting an extra load on the HQ.

The only reason we identified to enable long polling is that http 1.1 is blocked or when the upgrade sequence is not allowed.

frankcorneliusmartin:
Closing for now - re open if my earlier comment is addressed.

PR #2729, opened 2026-09-02
Bugfix/prometheus in sandbox failing volume mount

frankcorneliusmartin:
Prometheus failed to start in the Sandbox environment as the volume mount location was missing.


daniel-chou-rainho (commented):
LGTM

Issue #2736, opened 2026-09-09
[Bug] Chunk size not set for large run-data


tim-hendriks:
**Describe the bug**
  `UwsgiChunkedStream` in `vantage6-server/vantage6/server/resource/blobstream.py`  passes a byte count to `uwsgi.chunked_read()`, but that argument is a per-call **timeout in seconds** (the uwsgi Python API documents a default of 4), not a size:

  ```python
  def __init__(self, chunk_size=4096):
      """
      Initialize the UwsgiChunkedStream.
      """
      self.chunk_size = chunk_size
  ...
      chunk = uwsgi.chunked_read(self.chunk_size)
  ```

  So the default `chunk_size=4096` is really a **4096-second (≈68 minute) read
  timeout**. Two consequences:

  1. A client that begins a chunked upload and then stalls mid-body holds the request open for over an hour instead of failing fast.
  2. The name misleads. A reader of `read(size)` reasonably assumes reads are bounded at 4 KiB.

**To Reproduce**
  1. Start a server with the large run-data store enabled.
  2. `POST /api/blobstream` with `Transfer-Encoding: chunked`.
  3. Send one chunk, then hold the connection open without sending more and without closing it.
  4. The request remains open for ~4096 seconds rather than timing out.

**Expected behavior**
 Reads should use a bounded, sensible timeout.

**Additional context**
  Affects both maintained lines, with the same code in both:

  - v4: `vantage6-server/vantage6/server/resource/blobstream.py` (L281–306 on `release/4.16`)
  - v5: `vantage6-hq/vantage6/hq/resource/blobstream.py` (L284–309 on `main`)

  Suggested fix:

  ```python
  _DEFAULT_CHUNKED_READ_TIMEOUT_S = 30

  def __init__(self, read_timeout_seconds: int = _DEFAULT_CHUNKED_READ_TIMEOUT_S):
      self.read_timeout_seconds = read_timeout_seconds
  ...
      chunk = uwsgi.chunked_read(self.read_timeout_seconds)
  ```

PR #2744, opened 2026-09-15
Size chunked run-data uploads below the uwsgi per-part limit

tim-hendriks:
Chunked uploads used DEFAULT_CHUNK_SIZE, 1 MiB, as the size of each HTTP chunk on the wire. uwsgi refuses a chunked-input part strictly larger than --chunked-input-limit, whose built-in default is also 1 MiB (core/init.c sets chunked_input_limit = 1024*1024; the check in core/buffer.c is `len > limit`). That left the upload paths in two states:

- Direct client uploads and the node proxy's encrypted path (AES-CTR is length-preserving) sent parts of exactly 1 MiB — at 100% of the limit with zero headroom, passing only because the comparison is strict.
- The node proxy's path for unencrypted collaborations base64-encodes the stream, inflating a 1 MiB read to a ~1.37 MiB part, which uwsgi rejects with "unable to receive chunked part". The endpoint turned that into a generic 500, so it presented as a flaky upload rather than a sizing bug. Any result over ~768 KiB on that path failed deterministically.

Also fix UwsgiChunkedStream passing a byte count where uwsgi.chunked_read() expects a per-call timeout in seconds: the default chunk_size=4096 was really a ~68-minute read timeout, letting a stalled client hold a gevent worker for over an hour. Reads now use an explicit 30-second timeout, and a rejected part returns a 400 recommending the client part size instead of a bare 500.

Fixes vantage6/vantage6#2736


MarlouvS (approved):
Ziet er prima uit, weet alleen zelf inhoudelijk niet veel van UwsgiChunkedStream

vantage6-common/vantage6/common/globals.py
```diff
@@ -59,9 +59,24 @@
...
+# built-in default is 1 MiB): uwsgi counts each part plus its 2-byte CRLF
+# framing against that limit, so the largest usable part is limit - 2 and
+# anything bigger fails with ``IOError: unable to receive chunked part``.
+HTTP_UPLOAD_CHUNK_SIZE = 256 * 1024  # 256 KiB
+
+# Per-part cap (in bytes) that `v6 server start` passes to uwsgi as
+# ``--chunked-input-limit`` (server.sh sets the same value inline).
+# Deliberately much larger than ``HTTP_UPLOAD_CHUNK_SIZE`` so the friendly
+# client always has headroom, and small enough to reject obviously hostile
+# bodies before they consume server memory.
```
bartvanb, line 77: These are typical LLM-verbose explanations - we'd prefer to keep them shorter so that it's more human-readable

vantage6-node/vantage6/node/proxy_server.py
```diff
@@ -509,7 +509,9 @@ def stream_handler_post() -> FlaskResponse:
     url = f"{server_url}/blobstream"
     log.debug("Making proxied POST request to %s", url)
 
-    encrypted_stream = client.cryptor.encrypt_stream(request.stream, pubkey_base64)
+    encrypted_stream = client.cryptor.encrypt_stream(
+        request.stream, pubkey_base64, chunk_size=HTTP_UPLOAD_CHUNK_SIZE
```
bartvanb, line 513: I think these changes are not really necessary as you're providing a value that is already the default

vantage6-common/vantage6/common/globals.py
bartvanb: (btw on v5 branches we now have AGENTS.md where one of our instructions to LLM is not to make superlong comments that humans don't read)

vantage6-server/vantage6/server/resource/blobstream.py
```diff
@@ -264,6 +267,27 @@ def post(self):
...
         except Exception as e:
+            # "unable to receive chunked part" is the error uwsgi raises
+            # *most commonly* when a single chunked-input part exceeds
+            # ``--chunked-input-limit``, but the same error string can come
+            # from other transport-level issues (client disconnects mid-part,
+            # malformed chunk framing, …). Because we can't know which one
+            # we hit, the response is a generic 400 — a 413 would assert
+            # too much. The enforced limit lives in uwsgi's own
+            # configuration, so the message recommends the client-side part
+            # size rather than quoting a number this process cannot verify.
```
bartvanb, line 278: Could you shorten / remove this as well? The error message below should already clarify what is going on

vantage6-server/vantage6/server/resource/blobstream.py
```diff
@@ -264,6 +267,27 @@ def post(self):
...
+            if "unable to receive chunked part" in str(e).lower():
```
bartvanb, line 279: I think it's dangerous to catch an error like this, because we don't know if the error message wil change. Usually it'd be safer to catch error type - can you comment on why you did it like this?

vantage6-server/vantage6/server/resource/blobstream.py
```diff
@@ -264,6 +267,27 @@ def post(self):
...
+                log.error("Chunked upload rejected for run data %s: %s", result_uuid, e)
+                return {
+                    "msg": (
+                        "Upload rejected while receiving a chunked-input "
+                        "part — a part likely exceeded the server's "
+                        "per-part limit. The vantage6 client sends parts "
+                        f"of {HTTP_UPLOAD_CHUNK_SIZE} bytes; if you are "
+                        "calling the API directly, use a similar part "
+                        "size. Check the server logs to confirm the cause."
```
bartvanb, line 288: this can be shorter as well :-)

bartvanb (changes requested):
Looks good, pretty minor comments, mostly on style. At least for the time being, we want to ensure the vantage6 code stays as readable as possible which means being rather strict on verbosity of LLM comments / error messages. They tend to get so long and often rather technical that they are hard to read.

We've recently tried to address this in release/5.0 branch by adding an AGENTS.md that says LLMs shouldn't be too lengthy in their comments - maybe also place that on v4 branch

PR #2746, opened 2026-09-15
Fix repeated password reset requests silently failing after the first

### LINKED ISSUE #2267: [Bug] password reset broken on second attempt
**Describe the bug**
Once a password reset has been requested, it cannot be requested again.

**To Reproduce**
1. Request password reset
2. Wait
3. Request again

**Screenshots**
Example logs:
```
2025-09-17 09:50:21 - recover        - INFO     - Password reset requested for 'lsago'
Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 900, in gevent._gevent_cgreenlet.Greenlet.run
  File "/vantage6/vantage6-server/vantage6/server/resource/recover.py", line 170, in _handle_password_recovery
    dt.datetime.now(dt.timezone.utc)
TypeError: can't compare offset-naive and offset-aware datetimes
2025-09-17T09:50:21Z <Greenlet at 0x734b1f300f40: _handle_password_recovery(<Flask 'vantage6-server'>, None, lsago@example.com', {'description': 'Vantage6 Server for ....<vantage6.backend.common.mail_service.MailService )> failed with TypeError
```

**Details**
 - Vantage6: 4.12.2
 - Postgres 16 as database

**Additional context**
 https://github.com/vantage6/vantage6/pull/1707
https://github.com/vantage6/vantage6/pull/1807

tim-hendriks:
Closes #2267

`last_email_recover_password_sent` is a naive `DateTime` column. The throttle in  `_handle_password_recovery` compared the naive value read back from the database  against an aware `datetime.now(timezone.utc)`, which raises
`TypeError: can't compare offset-naive and offset-aware datetimes`. The check runs  in a gevent greenlet after the 200 response has already been sent, so the error was  swallowed. Since the timestamp is never cleared, every later reset request for  that user died the same way. After the first reset email a user ever received,  self-service password reset was permanently broken for them.

Fixed by utilizing `.replace(tzinfo=timezone.utc)` to attach the UTC label without changing the actual value (as opposed to astimezone() which would assume the servers local timezone and make an incorrect conversion).


vantage6-server/tests_server/test_models.py
```diff
@@ -103,6 +103,29 @@ def test_duplicate_user(self):
 
         session.session.remove()
 
+    def test_last_email_recover_password_sent_survives_db_round_trip(self):
```
tim-hendriks, line 106: As is probably fairly obvious by phasing, tests written with the help of Claude. At the same time I couldn't really find an argument to make changes here. Let us know what your stance on this is, of course we can alter them or remove them before review if you feel like it's a time waste.

MarlouvS (approved):
In mijn optiek is meer testen altijd goed

vantage6-server/vantage6/server/resource/recover.py
```diff
@@ -168,7 +168,7 @@ def _handle_password_recovery(
     # check that email has not already been sent recently
     email_sent_recently = user.last_email_recover_password_sent and (
         dt.datetime.now(dt.timezone.utc)
-        < user.last_email_recover_password_sent
+        < user.last_email_recover_password_sent.replace(tzinfo=dt.timezone.utc)
```
bartvanb, line 171: Now we're converting database-stored timestamps to UTC. I wonder if it is not better to automatically store all timestamps in the database as UTC - not sure how that can be done but pretty sure it's possible. 

Doing it that way is I think better to prevent this type of bug in the future. If you agree, we should maybe create an issue to pick this up (only v5?)

vantage6-server/tests_server/test_models.py
```diff
@@ -103,6 +103,29 @@ def test_duplicate_user(self):
...
+        user = User(username="reset-pw-user", email="reset-pw-user@org.org")
+        user.last_email_recover_password_sent = sent_at
+        user.save()
+
+        # force a reload from the database, as happens the second time
+        # `_handle_password_recovery` looks up the user
+        reloaded_user = User.get_by_username("reset-pw-user")
+        stored = reloaded_user.last_email_recover_password_sent
+        self.assertIsNone(stored.tzinfo)
+        self.assertEqual(stored.replace(tzinfo=datetime.timezone.utc), sent_at)
```
bartvanb, line 127: I'm a bit confused by this test, doesn't it just pass always unless the database manipulates the timestamp? Would it not be a more informative test if the `sent_at` timestamp is not a UTC timestamp?

bartvanb (approved):
Looks good, please have a look at the comments before merging

PR #2747, opened 2026-09-15
Only move the latest Docker tags for the current major

tim-hendriks:
The split step decided whether to move the ':latest' tags from the release stage alone, so any final release set them, whatever its major version. 'latest' is a single tag shared by every major, so a v4 patch release published after v5.0.0 pulled it backwards for the six images both lines build.


MarlouvS (approved):

bartvanb (approved):
Thanks a lot, both for noticing and fixing this!

bartvanb:
I'm including this in the patch in 4.15 (to be released early next week as discussed) as I think it's important to fix this asap

PR #2749, opened 2026-09-16
V4/bugfix/2737 after delete listener leak

### LINKED ISSUE #2737: [Bug] After delete listener leak
**Describe the bug**
`AzureStorageService.__init__` registers a SQLAlchemy `after_delete` listener on the `Run` **class**:

  ```python
  # vantage6-server/vantage6/server/service/azure_storage_service.py:66
  event.listen(Run, "after_delete", self.delete_blob_after_run_delete)
  ```
  The target is the mapped class, which is global and lives as long as the process, and the listener is a bound method of that particular instance. Nothing ever removes it. So every `AzureStorageService` that gets constructed adds one more listener that fires on every subsequent `Run` deletion, for the lifetime of the server.

 Instances are not constructed once. `cleanup_runs_data` builds a fresh one on each call:

  ```python
  # vantage6-server/vantage6/server/controller/cleanup.py:27
  storage_adapter = AzureStorageService(azure_config)
  ```

  and that function is called in an unbounded loop by the background cleanup worker:
 
  ```python
  # vantage6-server/vantage6/server/__init__.py:879-895
  def __runs_data_cleanup_worker(self):
      while True:
          try:
              cleanup.cleanup_runs_data(...)
          ... 
          time.sleep(3600)
  ```

The listener re-raises on failure:

  ```python
  # azure_storage_service.py:147-161
  def delete_blob_after_run_delete(self, mapper, connection, target):
      if target.blob_storage_used:
          try:
              if target.result:
                  self.delete_blob(target.result)
              ... 
          except Exception as e:
              error_msg = f"Failed to delete blob for run {target.id}: {e}"
              log.error(error_msg)
              raise RuntimeError(error_msg)
  ```

**To Reproduce**
  1. Configure a server with `large_result_store` (Azure) **and** `runs_data_cleanup_days`.
  2. Start it. Two adapters now exist: one from `setup_large_result_store`, one from the
     first `cleanup_runs_data` call. 
  3. Create a task whose run data goes to blob storage, so `Run.blob_storage_used` is true.
  4. Delete that `Run`.
  5. The deletion fails with `RuntimeError: Failed to delete blob for run <id>:
     ResourceNotFoundError`, and the row remains in the database.

**Expected behavior**
  The listener should be registered once per process regardless of how many adapter instances are created, and a single `Run` deletion should attempt each stored entry  exactly once.

**Additional context**

  Present on both maintained lines, with the same structure:

  - v4 `release/4.16`: `azure_storage_service.py:66`, `controller/cleanup.py:27`,
    `server/__init__.py:879-895`
  - v5 `main`: `vantage6-hq/vantage6/hq/service/azure_storage_service.py:67`,
    `vantage6-hq/vantage6/hq/controller/cleanup.py:29`,
    `vantage6-hq/vantage6/hq/__init__.py:411-427`

tim-hendriks:
AzureStorageService init registered its SQLAlchemy after_delete listener on the mapped Run class, a process-lifetime target that never releases what is registered against it. Every construction therefore added one more listener, and all of them fired on every Run deletion.

Fixed by registering the Run delete cascade once per process. A single module-level listener is attached the first time a usable service is built. The server's storage service in the cleanup worker now takes the service as a parameter and the app passes the one it built at startup.

Closes #2737


MarlouvS (commented):
Dit heeft alleen invloed op handmatige deletion van runs toch? Of wordt dit proces nog op andere momenten aangeroepen?

vantage6-server/vantage6/server/controller/cleanup.py
```diff
@@ -11,20 +11,26 @@
...
     ----------
-    days : int
-        The number of days after which results should be cleared.
+    config : dict
+        Server configuration.
+    storage_adapter : AzureStorageService | None
+        Service for the large result store, or None when it is not
+        configured.
+    include_input : bool
+        Whether to clear the `input` field as well as the `result` field.
```
bartvanb, line 31: Thanks for fixing this.
I would prefer to not make the storage_adapter a separate argument in this function though, this code change appears to be not really necessary and fewer args to a function is preferable to me. Only when I check now, the issue with the old code might be that storage_adapter was in some cases not defined right? So then it should be something like 
```
storage_adapter = AzureStorageService(azure_config) if azure_config else None
```

vantage6-server/vantage6/server/service/azure_storage_service.py
```diff
@@ -12,11 +12,23 @@
 log = logging.getLogger(module_name)
 
 
+def _delete_blob_after_run_delete(mapper, connection, target) -> None:
+    """
+    Forward a ``Run`` deletion to the active storage service.
+    """
+    service = AzureStorageService._active
+    if service is not None:
+        service.delete_blob_after_run_delete(mapper, connection, target)
```
bartvanb, line 21: Why did you choose to put this function outside the class?

bartvanb (changes requested):
very minor comments. Nice find and decent fix

PR #2752, opened 2026-09-18
Read `require_algorithm_pull` from `policies` in the k8s container manager

### LINKED ISSUE #2732: Node ignores `policies.require_algorithm_pull` (reads `node.require_algorithm_pull`); locally built algorithm images can never run
**Component:** node (`vantage6/node/k8s/container_manager.py`) and node Helm chart.

**Describe the bug**
The node chart renders the setting into the node config as
```yaml
policies:
  require_algorithm_pull: false
```
(`templates/node-configmap.yml`, from `.Values.node.policies.require_algorithm_pull`), but the node reads it from a different place:
```python
require_algorithm_pull = self.ctx.config.get("node", {}).get("require_algorithm_pull", True)
```
so the value is always the default `True` and every algorithm job is created with `imagePullPolicy: Always`. An image that exists only in the cluster's Docker/containerd (the documented way to test an algorithm locally) fails with `ErrImagePull` / status "non-existing Docker image", even with `require_algorithm_pull: false` set as documented.

**To reproduce**
1. Build an algorithm image locally (e.g. `docker build -t myalg:0.1 .`) on a cluster that shares the Docker daemon (colima/k3s or Docker Desktop).
2. Set `node.policies.require_algorithm_pull: false` in the node chart values / node config as documented.
3. Create a task with image `myalg:0.1`.
4. Job is created with `imagePullPolicy: Always`; pod shows `ErrImagePull`, task status "non-existing Docker image".

**Expected behavior**
`container_manager.py` should read `policies.require_algorithm_pull` (where the chart, the config template and the docs put it), or the chart should also write `node.require_algorithm_pull`.

**Workaround**
Patch the ConfigMap by hand and restart the node (`helm upgrade` reverts it):
```
kubectl -n <ns> get configmap <release>-configmap -o jsonpath='{.data.config\.yml}' > cfg.yml
printf 'node:\n  require_algorithm_pull: false\n' >> cfg.yml
kubectl -n <ns> create configmap <release>-configmap --from-file=config.yml=cfg.yml --dry-run=client -o yaml | kubectl apply -f -
kubectl -n <ns> rollout restart deployment/<release>
```

**Environment**
vantage6 5.0.2 (CLI, node, algorithm tools); colima 0.10.3 with k3s v1.35.0 on macOS 26.6 (Apple silicon), Docker runtime.

Found while building a federated ALD registry (Amsterdam UMC / UK Leipzig) on vantage6.

developer-rpai:
## Bug

The node Helm chart (`charts/node/templates/node-configmap.yml`) renders the setting into the node config as:

```yaml
policies:
  require_algorithm_pull: false
```

and the documented node configuration (`docs/node/node_config.yaml`) places it under `policies` as well. But `ContainerManager.run()` in `vantage6-node/vantage6/node/k8s/container_manager.py` read it from a different place:

```python
require_algorithm_pull = self.ctx.config.get("node", {}).get("require_algorithm_pull", True)
```

So the documented setting had no effect: the value was always the default `True`, and every algorithm job was created with `imagePullPolicy: Always`. An image that exists only in the cluster's local container runtime (the documented way to test an algorithm locally) fails with `ErrImagePull` / status "non-existing Docker image", even with `policies.require_algorithm_pull: false` set.

## Fix

Read the setting from `policies` via the existing `NodePolicy.REQUIRE_ALGORITHM_PULL` constant, consistent with the chart, the docs, and the rest of the node code (e.g. `_setup_policies` in the same file reads `config.get("policies", {})`):

```python
# The node Helm chart and the documented node configuration place this
# setting under ``policies`` (not under ``node``).
require_algorithm_pull = self.ctx.config.get("policies", {}).get(
    NodePolicy.REQUIRE_ALGORITHM_PULL.value, True
)
```

`NodePolicy` was already imported in this module; no new dependencies. The default stays `True`, so behavior is unchanged for configs that don't set the option.

## Test evidence

Added `vantage6-node/tests_node/test_require_algorithm_pull.py` (unittest, same style as the existing `tests_node` tests). It exercises `ContainerManager.run()` end-to-end with a mocked cluster: it captures the `V1Job` passed to `create_namespaced_job` and asserts on the container's `imagePullPolicy`.

- `policies.require_algorithm_pull: false` -> `IfNotPresent` (this case **fails before the fix** with `Always`, reproducing the reported bug; passes after)
- `policies.require_algorithm_pull: true` -> `Always`
- no setting at all -> `Always` (default preserved)

Run: `python -m unittest discover -s vantage6-node/tests_node` -> all 9 tests OK (6 existing + 3 new).

Also verified the pre-existing ruff findings on `container_manager.py` are unchanged by this patch, and `ruff format --check` / `ruff check` are clean on both changed files.

Fixes #2732


bartvanb:
Already fixed in https://github.com/vantage6/vantage6/pull/2748

